PR #54, opened 2014-01-29
Addition fixes for #3842: add support for enabling a list of repos after registration

skottler:


windkisser:
is there anyone encountered a problem like mine. Current hostname is shortname  when i use subscription-manager like below in my ks snippets the subscription status is "not available". Any solutions? Thanks

" subscription-manager register --org="<%= @host.rhsm_organization_label %>" --name="<%= @host.name %>" --activationkey="<%= @host.params['kt_activation_keys'] %>"

PR #59, opened 2014-02-25
Basic Layout No Swap

abyss1:
(Now in the right place)


ares:
@abyss1 could you please reopen against develop branch? we renamed master to develop since we started branching this repo per foreman releases

PR #112, opened 2014-09-10
rhbz#1140057 don't run auto-attach if pool is specified

mburns72h:
currently we always run auto-attach with subscription-manager.
we should only run it if we don't have a pool.

Signed-off-by: Mike Burns <mburns@redhat.com>


mmoll:
what's the status here?

ekohl:
I don't know subscription-manager, but it looks like it makes sense and I trust @mburns72h so :+1:

domcleal:
Thanks @mburns72h!

PR #133, opened 2014-12-22
change puppet package for SLES

mmoll:
The package name in the systemsmanagemnt:puppet repository got changed (similar to https://bugzilla.opensuse.org/show_bug.cgi?id=908212).

I'd like to continue installing the Augeas rubygem together with Puppet in our templates to be able to use Augeas statements in Puppet manifests out of the box also on SLES. Other Puppet packages (from Puppetlabs, Debian/Ubuntu, EPEL) depend on that gem, so this functionality is just there on other distros.

### DIFF
--- a/autoyast/provision_sles.erb
+++ b/autoyast/provision_sles.erb
@@ -74,8 +74,7 @@ oses:
       <package>openssh</package>
       <package>vim</package>
 <% if puppet_enabled -%>
-      <package>puppet</package>
-      <package>rubygem-ruby-augeas</package>
+      <package>rubygem-puppet</package>
 <% end -%>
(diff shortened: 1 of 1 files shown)


mmoll:
As it turned out, the puppet package was renamed to rubygem-puppet and is depending correctly on augeas now. The other problems (augeas not working and dmidecode[ ->facter -> puppet ] not installable on SLES 11) will be solved by the systemsmanagement:puppet people.

domcleal:
Thanks @mmoll, confirmed via https://build.opensuse.org/project/show/systemsmanagement:puppet and the bug report you linked.

PR #134, opened 2015-01-09
Kickstart: Install lsb package by default, it is required for core facts

sathieu:


domcleal:
Fine by me, :+1:

ekohl:
I always hated redhat-lsb because of the many dependencies, but this doesn't appear to have any. Is redhat-core-lsb available on all platforms and why was I not aware of this in the past?

domcleal:
It was added circa RHEL 6.3 precisely for this reason.  It has *some* dependencies, but certainly none of the LSB printing group, etc.

ekohl:
I don't have EL5 at hand, but does it also work there?

domcleal:
No, it won't - so it should be wrapped in `if os_major >= 6` (it should be fine on Fedora).

sathieu:
Thanks for your remarks. I've updated the PR.

mmoll:
I have a modification like this in my local templates for a long time, so :+1:.

On Debian/Ubuntu ```lsb-release``` could be added and It's also needed for some grains when using SaltStack.

sathieu:
@mmoll lsb-base is already in Debian preseed:
```
d-i pkgsel/include string <%= puppet_package %> <%= salt_package %> lsb-release
```
Ref: https://github.com/theforeman/community-templates/blob/master/preseed/provision.erb#L130

mmoll:
Indeed, sorry.
Merging, thanks @sathieu

PR #135, opened 2015-01-15
Workaround for Debian/jessie GRUB boot device selection

fraenki:
Debian Jessie is currently in BETA, but an annoying bug makes automatic installs impossible:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=712907

This workaround fixes this. I'm in doubt that this bug will be fixed in time for the Debian 8.0 release.


mmoll:
I also ran into that, but for the moment decided to wait until jessie is released as the bug is already fixed in unstable and "only" has to find it's way into testing. I'm open for different opinions, though. :mega:

fraenki:
@mmoll I think it's a valid option to wait until jessie is released.

mmoll:
Thanks, leaving this open as reference until the fix gets into testing or jessie gets released.

mmoll:
@fraenki, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=759737 is in testing, could you try again with a recent installer build?

mmoll:
Confirmed that something like
```d-i grub-installer/bootdev string default```
will be necessary, thanks @lazyfrosch for digging this up on IRC. I'm already having an idea how to hook this up with ```@host.params['install-disk']```, PR incoming in some days.

PR #139, opened 2015-01-20
Added base-url support for CoreOS to enable local mirror server

johscheuer:
Added the base url support for the coreos-install script. Also if no SSH keys are supplied the core user gets the root password to enable login even if you forget to add your SSH key.


endyman:
Would it make sense to implement @mediumpath for this purpose ? Otherwise we would need to replicate information already available in the install medium

eLobato:
ack cc @mmoll

mmoll:
thanks @johscheuer!

PR #142, opened 2015-01-28
Fixes #9139 - restore compatibility with latest networking

ares:


snippets/kickstart_networking_setup.erb
```diff
@@ -113,8 +116,14 @@ DEVICE="$real"
 HWADDR="<%= interface.mac -%>"
 <% end -%>
 ONBOOT=yes
+<% if interface.respond_to?(:primary) && interface.primary -%>
+DEFROUTE=yes
```
domcleal: this is a PPP-specific option, I don't think it's relevant as the route'd be determined by the gateway?

ares: I think that default route is usually determined by interface gateway which comes from DHCP. We can have more interfaces, each with different gateway. My understanding was that in such case DEFROUTE sets which gateway is used for default route (making it default gateway). I'm not 100% sure about all DEFROUTE impacts and there's not much documentation.

domcleal:
Looks about right to me :+1:

mmoll:
I didn't test this, but looking OK, so merged. Thanks @ares!

PR #143, opened 2015-01-28
Added Discovery PXELinux template snippet

lzap:
This is the first stage of some Discovery integration. Currently users mus
create whole configuration from scratch, which is error prone. My goal for
Foreman 1.8 is to provide this snippet which is not yet set as the default.
Users will be asked to edit the line (`YOURPROXY` must be set) and making it
the default option to get Discovery working.

Next stage would be to automatically show the option when discovery is
installed a global setting flag is set. But we can only do this once we have
multi-proxy default PXELinux file support (as multiple subnets may require
different proxy URLs).


lzap:
Since the recommended way of configuring discovery for 2.0 is direct communication, I changed the configuration accordingly.

pxe/PXELinux_default.erb
```diff
@@ -17,6 +17,13 @@ LABEL local
      MENU DEFAULT
      LOCALBOOT 0
 
+LABEL discovery
+     MENU LABEL (discovery)
+     MENU DEFAULT
+     KERNEL boot/fdi-image/vmlinuz0
+     APPEND initrd=boot/fdi-image/initrd0.img rootflags=loop root=live:/fdi.iso rootfstype=auto ro rd.live.image acpi=force rd.luks=0 rd.md=0 rd.dm=0 rd.lvm=0 rd.bootif=0 rd.neednet=0 nomodeset proxy.url=https://FOREMAN_INSTANCE proxy.type=foreman
```
mmoll: can ```unattended_url``` or some other value be used to set the URL automatically?

pxe/PXELinux_default.erb
```diff
@@ -17,6 +17,13 @@ LABEL local
      MENU DEFAULT
      LOCALBOOT 0
 
+LABEL discovery
+     MENU LABEL (discovery)
+     MENU DEFAULT
```
mmoll: The grey bar is at ```(discovery)``` in my setup then, although ```(local)``` gets booted. That may depend on the PXELinux version, but to be sure, I would delete this line.

pxe/PXELinux_default.erb
domcleal: It would be pretty easy I'd have thought to pass the TFTP proxy server info into the template renderer in core.

pxe/PXELinux_default.erb
lzap: Yeah, removed.

pxe/PXELinux_default.erb
lzap: I tired and actually it is complicated, I'd rather have a first cut as a manual edit.

https://github.com/theforeman/foreman/pull/1827

mmoll:
thanks @lzap!

PR #144, opened 2015-01-30
Add spacewalk_gpg_keys host parameter to install rpm gpg keys from url

Klaas-:
I ran into a problem with spacewalk and redhat_register, when using channels through spacewalk that are signed with keys that are not installed on system this leads to installation not completing (yum update and install don't work). usually gpg keys get added through repository file but spacewalk doesn't provide those.
I've created a host parameter that allows to use urls to add those gpg keys through urls - I usually put all gpg keys into http://spacewalkserver/pub/ so they are easly accessable. If you want to use https for your spacewalk server it needs a valid cert or you need to add it (rpm uses curl) before calling redhat_register snippet.
For convenience its easiest to add the parameter on operating system level :)
If you don't explicitly set an url the default behaviour is unchanged.


mmoll:
Please open one PR for one change.

Regarding the ssh root login with password, that's also an issue with Debian/jessie, but I don't want to have this in upstream templates, unless there's as real demand for that. Personally, I think it's much more feasible to leave it as is and use some config management system to distribute SSH keys if you really want to login as root remotely. If it really should be open for logins using passwords, there's probably some debconf variable that can be set in the preseed file.

mmoll:
@Klaas- how do we proceed further on this one?

mmoll:
Closing this, @Klaas-, please consult the GitHub help on how to use branches in git in regard to pull requests.

PR #146, opened 2015-02-06
set grub-installer/bootdev in preseed

mmoll:
see Debian bugs 712907 and 759737 for reference, required for Debian/jessie
closes GH-135

As already found out in earlier PRs, older installers will ignore unknown lines, so this will do no harm.
Maybe @lazyfrosch and @sathieu could review?

### DIFF
--- a/preseed/iPXE.erb
+++ b/preseed/iPXE.erb
@@ -5,7 +5,9 @@ name: Preseed default iPXE
 oses:
 - Debian 6.0
 - Debian 7.
+- Debian 8.
 - Ubuntu 12.04
+- Ubuntu 14.04
 %>
 <% if @host.operatingsystem.name == 'Debian' -%>
 <% keyboard_params = "auto=true domain=#{@host.domain}" -%>
--- a/preseed/userdata.erb
+++ b/preseed/userdata.erb
@@ -4,7 +4,9 @@ name: Preseed default user data
 oses:
 - Debian 6.0
 - Debian 7.
+- Debian 8.
 - Ubuntu 12.04
+- Ubuntu 14.04
 %>
 #!/bin/bash
 
--- a/preseed/PXELinux.erb
+++ b/preseed/PXELinux.erb
@@ -4,9 +4,11 @@ name: Preseed default PXELinux
 oses:
 - Debian 6.0
 - Debian 7.
+- Debian 8.
 - Ubuntu 10.04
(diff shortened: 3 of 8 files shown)


lazyfrosch:
lgtm

ekohl:
:+1: Looks sane to me.

preseed/iPXE.erb
```diff
@@ -6,6 +6,7 @@ oses:
 - Debian 6.0
 - Debian 7.
```
sathieu: debian 8 is missing here

preseed/userdata.erb
```diff
@@ -5,6 +5,7 @@ oses:
 - Debian 6.0
 - Debian 7.
```
sathieu: debian 8 is missing here

sathieu:
otherwise, looks good! (but I consider this as a workaround, waiting proper fix for https://bugs.debian.org/712907)

lazyfrosch:
Yes in terms of we should mention Debian 8 there. As the installer will not have many changed until the release.

@sathieu `bootdev default` will actually be the proper setting to tell the installer to use the same disk for the base system and bootloader. The Jessie installer is designed to ask the user for the bootdev. In terms of how Foreman handles the installation, this PR is the proper way to supply that information. I guess https://bugs.debian.org/712907 will be closed soon.

lazyfrosch:
@sathieu but still I like to know why you consider this a workaround, a reply to the Debian bug might help :)

sathieu:
@lazyfrosch See the comment from Petter Reinholdtsen (#190) which clearly says that this but is workarounded by a choose_bootdev.

lazyfrosch:
@sathieu In terms of that funny shell function, or any other early_command, that is true.

But we are using:
```
d-i grub-installer/bootdev string default
```

Or a disk configured via a host param.

Which has been added with http://bugs.debian.org/759737 (http://anonscm.debian.org/cgit/d-i/grub-installer.git/commit/?id=67d7246accead493823e34f933aea7887d191ea7)

sathieu:
@lazyfrosch OK.

mmoll:
Added Debian 8 to the headers of the remaining two files, otherwise I'm feeling consensus from all.

sathieu:
@mmol +1

mmoll:
thanks all!

PR #147, opened 2015-02-09
Fixes #9287 - Configure networking during image provisioning

ares:


mmoll:
visually OK, I presume you tested it, I remember vaguely a discussion about network restart, but can't remember if that was for RH based OSes...

ares:
yes, the test was also presented in cfgmgmt demo :-) we removed restarting of networking from kickstart template but here we must restart it, since the host is not rebooted after finish template is done

ekohl:
I don't have a strong opinion about this. I don't like restarting the network during kickstart, but I see that it solves a problem so :+1:

mmoll:
thanks @ares!

PR #149, opened 2015-02-16
refs #8498 - set root password in finish script

domcleal:
Taken from https://github.com/theforeman/foreman/commit/c47be8a#diff-7ccbfbfcac27e2ec369bf0f121c15635 which should've come here first.


ekohl:
:+1:

mmoll:
thanks @TJM and @domcleal!

PR #151, opened 2015-02-22
Add built-in support for chef

ares:
It's not ideal to add code required for plugins to work, but there's probably no better way of doing this right now. It would be really awesome if we could get this into 1.8, Chef integration could be finally installed easily and could compete to other cfgmgmt tools.

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -21,6 +21,7 @@ oses:
   proxy_string = @host.params['http-proxy'] ? " --proxy=http://#{@host.params['http-proxy']}:#{@host.params['http-proxy-port']}" : ''
   puppet_enabled = pm_set || @host.params['force-puppet'] && @host.params['force-puppet'] == 'true'
   salt_enabled = @host.params['salt_master'] ? true : false
+  chef_enabled = @host.respond_to?(:chef_proxy) && @host.chef_proxy
   section_end = (rhel_compatible && os_major <= 5) ? '' : '%end'
 %>
 install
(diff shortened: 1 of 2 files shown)


kickstart/provision.erb
```diff
@@ -143,6 +144,11 @@ echo "updating system time"
 # update all the base packages from the updates repository
 yum -t -y -e 0 update
 
+<% if chef_enabled %>
+echo "Bootstraping chef"
+<%= respond_to?(:chef_bootstrap) ? chef_bootstrap(@host) : snippet_if_exists("chef-client omnibus bootstrap") %>
```
mmoll: Where does this snippet come from? Also, preseed support would be nice.

ares: foreman_chef will provide it (will push the new version today/tomorrow), but the template should work even without it
I'll also add debian version soon

ares: if you're interested, here's the foreman_chef PR that adds a snippet(s) https://github.com/theforeman/foreman_chef/pull/11

mmoll: I see... As we're having the salt snippets also in community-templates, could you add the chef ones also here, please?

ares:
I added a debian template as well.

kickstart/provision.erb
ares: I'd prefer not to because
* foreman_chef is not released on regular basis and may need to change template from release to release I need better control over these templates, this may change later
* I believe this belongs to plugin itself, these templates would be useless without plugin (which is difference to salt IIUC, salt uses global parameters, these templates rely on bootstrap helper provided by the plugin)

mmoll: I understand, however, AFAIK changes to the templates can not be done by editing the seed files, you would have to add a migration in that case and you can't be sure what people are changing (but you could lock them probably). Havin the templates here in git and using foreman_templates is better, IMHO.

I leave it up to you :)

mmoll:
OK :+1:

kickstart/provision.erb
ares: They are seeded locked and seeds.rb updates their content if needed, same approach they took in katello. Later when they are more stabilized I may move them here, but now I need to change them as I'm adding new features (client precreation instead of using validation.pem etc).

domcleal:
Thanks @ares!

PR #156, opened 2015-03-03
Use .empty? instead of .present? when rendering root_pass

eLobato:
`.present?` is not available for String::Jail, 'not nil and not empty' is the closest equivalent I think.

Currently it will throw the following error when trying to render:
`There was an error rendering the Kickstart default finish template: undefined method 'present?' for String::Jail (String)`


kickstart/finish.erb
```diff
@@ -24,7 +24,7 @@ service network restart
   puppet_enabled = pm_set || @host.params['force-puppet'] && @host.params['force-puppet'] == 'true'
 %>
 
-<% if @host.provision_method == 'image' && root_pass.present? -%>
+<% if @host.provision_method == 'image' && !root_pass.nil? && !root_pass.empty? -%>
```
domcleal: shouldn't need to check nil, `nil.empty?` works in Foreman

domcleal:
:+1:

eLobato:
@domcleal Fixed, but untested

mmoll:
could any image RH users test?

eLobato:
I can record a gif tomorrow if no one volunteers

eLobato:
Just one last comment, root passwords like '          ' are not `.empty?`, would that still be :ok: ?

mmoll:
:+1:

domcleal:
@eLobato I'm not sure how you've achieve that, as Foreman will always try to crypt root passwords, so a space would get saved as `$1$Qhg7.66F$2Z1yJ32bjjPNuZh0mxQ0i1` or similar.

eLobato:
@domcleal I didn't try to do it, but right.. it shouldn't be the case normally.

domcleal:
Thanks!

PR #157, opened 2015-03-03
Refs #9615 - Use parent interface MAC for virtual interfaces

ares:
This relies on https://github.com/theforeman/foreman/pull/2222, I use respond_to? as I don't know whether this will get into 1.8. If it will, we can simplify it to use just `interface.inheriting_mac` without any condition.


ekohl:
Since we prefer to keep compatibility for a bit, I'd prefer the current version because that'll work on 1.7.

ares:
right

ekohl:
in case it wasn't clear, :+1: from me as it is.

mmoll:
thanks @ares!

PR #158, opened 2015-03-03
WIP: Add netcfg/hostname to ensure a correct hostname

lazyfrosch:
By default the Debian installer will prefer any hostname/domain received from
DHCP.

Unless you set this setting, the domain might get changed on the fly.

The real problem is even very irritating, because the resulting hostname will
make it through Puppet.

    Foreman host:    host123.mydomain.com
    DHCP domain:     foobar.com
    System hostname: host123.foobar.com
    Puppet certname: host123.mydomain.com (set via the puppet.conf template)
                     (which will be signed by the master)
    Resulting host:  host123.mydomain.com (freshly created in Foreman - based
                     on the fact fqdn)

The Debian preseed docs state the following:

    # Any hostname and domain names assigned from dhcp take precedence over
    # values set here. However, setting the values still prevents the questions
    # from being shown, even if values come from dhcp.
    d-i netcfg/get_hostname string unassigned-hostname
    d-i netcfg/get_domain string unassigned-domain

    # If you want to force a hostname, regardless of what either the DHCP
    # server returns or what the reverse DNS entry for the IP is, uncomment
    # and adjust the following line.
    #d-i netcfg/hostname string somehost


lazyfrosch:
This should work in >= wheezy

Thoughts?

GregSutcliffe:
:+1: pending testing ;)

lazyfrosch:
Bummer, looks like Ubuntu 14.04 does not like that flag, will investigate

lazyfrosch:
Sorry guys, this won't work, it is actually only possible to force the hostname to be that name.

Even if being a fqdn the domain from DHCP will be preferred.

I read into the code, unless the change the options in the installer.

Things like `hostname=` and `domain=` on cmdline will only set default values, getting overwritten by DHCP it supplies that information.

PR #159, opened 2015-03-03
Use proxy on FreeBSD if specified

sathieu:


ekohl:
:+1:

mmoll:
Please add --direct to the fetches that go the foreman instance, as that's in line with the other OSes (using proxy only for the repositories). Also, it should be tested if the behaviour is like desired in the chroot (finish script), as the pkg commands also would need the proxy to be set then.

sathieu:
I've not finished testing FreeBSD yet. Please wait a bit before merging.

mmoll:
It would also be nice if @mmatuska could provide a 10.1 image over at http://mfsbsd.vx.sk/files/foreman/

sathieu:
I've now finished the tests and the FreeBSD is running well

mmoll:
Works very well and finally forced me to tweak my apt-cacher-ng config for FreeBSD pkgng, merci @sathieu!

PR #162, opened 2015-03-09
Fixes #5458 - use subnet dhcp boot mode if possible

ares:


ares:
warning: not tested most of it

lzap:
I don't understand this patch, why you change only Debian part of this? Is Red Hat template already modified this way? Can you explain please? Is 61cfd65bc9c96ee483c38fbc578bf05c4e849faa related?

domcleal:
@lzap yes, the commit you linked changed the kickstart to obey the subnet boot mode and not just the (deprecated?) `@static` variable.

preseed/provision.erb
```diff
@@ -146,4 +152,4 @@ d-i grub-installer/bootdev string default
 <% end -%>
 d-i finish-install/reboot_in_progress note
 
-d-i preseed/late_command string wget -Y off <%= @static ? "'#{foreman_url('finish')}&static=true'" : foreman_url('finish') %> -O /target/tmp/finish.sh && in-target chmod +x /tmp/finish.sh && in-target /tmp/finish.sh
+d-i preseed/late_command string wget -Y off <%= dhcp ? foreman_url('finish') : "'#{foreman_url('finish')}&static=true'" %> -O /target/tmp/finish.sh && in-target chmod +x /tmp/finish.sh && in-target /tmp/finish.sh
```
domcleal: Perhaps keep the conditional based on `@static` here, so we don't begin passing static=yes to finish scripts when a subnet is set to static.  Finish scripts can still check for the subnet boot mode.

domcleal:
I haven't had chance to test this, but would like to see it in 1.8 for parity with kickstarts - if anybody gets the chance to install via preseed to see that the subnet boot mode on static configures the host correctly (or at least attempts to), it'd be appreciated.

mmoll:
tested, working fine :+1:

preseed/provision.erb
ares: Makes sense, reverted.

ares:
The only change I did after @mmoll testing was reverting the condition in preseed provision template as @domcleal suggested.

domcleal:
Thanks @ares and @mmoll, merged as 5339d8c.

PR #163, opened 2015-03-12
use correct provisioning URL for FreeBSD

mmoll:
found that after upgrading to 1.8, I wonder why it worked before :)


ekohl:
I think it ```foreman_url``` changed in 1.8. Explicit over implicit anyway so :+1:

domcleal:
How did it change, OOI?

ekohl:
Maybe I only remember us talking about it, but I do recall us talking about the default argument for ```foreman_url``` because it's ambiguous what it exactly does.

domcleal:
Yeah, true.  Often it refers to the "built" URL, so this change is good.  Thanks @mmoll!

PR #164, opened 2015-03-16
Added directory structure support to the url to the epel rpm file, as ce...

krizzo:
Added directory structure support to the url to the epel rpm file, as centos 7 changed to be under e/ and this is a bit more flexible with each repo change.


domcleal:
:+1:

ekohl:
Thanks!

PR #166, opened 2015-03-20
Add keyboard option for kickstart files

claer:
- Add a new parameter 'keyboard' for customizing keyboard at install time
 - Summarize the list of parameters for the kickstart files. (It would be nice
   to do it for all templates =))


mmoll:
Is it possible to implement some logic like in #119? If not, it's not a problem, just drop a line and I'll test with CentOS when I have time.

claer:
Internal functions are doing it for graphical installer but I found nothing in anaconda documentation to do it from the kickstart file.

mmoll:
merged, thanks @claer! :dancers:

PR #167, opened 2015-03-24
Simplify EPEL snippet with new permalinks

ekohl:


domcleal:
Too good to be true!

ekohl:
Net negative patches are the best patches :)

PR #168, opened 2015-03-30
kickstart: Set proxy permanently

sathieu:


sathieu:
This is required for all "yum install" calls.

mmoll:
I do have something similar in my local templates for usage with apt-cacher-ng and would not be against it, but am unsure. Let the kickstart experts decide. :boot:

kickstart/provision.erb
```diff
@@ -30,7 +30,8 @@ This template accepts the following parameters:
   realm_compatible = (@host.operatingsystem.name == 'Fedora' && os_major >= 20) || (rhel_compatible && os_major >= 7)
   # safemode renderer does not support unary negation
   pm_set = @host.puppetmaster.empty? ? false : true
-  proxy_string = @host.params['http-proxy'] ? " --proxy=http://#{@host.params['http-proxy']}:#{@host.params['http-proxy-port']}" : ''
+  proxy_uri = @host.params['http-proxy'] ? "http://#{@host.params['http-proxy']}:#{@host.params['http-proxy-port']}" : ''
+  proxy_string = proxy_uri ? " --proxy=#{proxy_uri}" : ''
```
domcleal: proxy_uri is always trueish, because the empty string assigned on line 33 is trueish, so it's generating `--proxy=` in the kickstart

domcleal: Probably best to set proxy_uri to `nil` if it's not set, then it'll also work with the if statement added below.

sathieu:
I've updated the PR replacing '' by nil.

mmoll:
tested with CentOS 6 and working fine :+1:

mmoll:
@sathieu sorry, this got lost somehow, merged as f0def55 now, thanks!

sathieu:
thanks!

PR #177, opened 2015-05-07
Update provision.erb

NuxRo:
added conditionals for EL5 network as it does not support "network --device=$macaddr"


kickstart/provision.erb
```diff
@@ -51,7 +51,11 @@ skipx
 <% dhcp = !@static -%>
 <% end -%>
 
+<% if rhel_compatible && os_major = 5 -%>
```
ekohl: Shouldn't that be ```==```?

mmoll:
Could you please update the RHEL provisioning template also in this PR?

PR #178, opened 2015-05-07
Update provision_rhel.erb

NuxRo:
Don't use "network --device=$macaddr" for EL5


kickstart/provision_rhel.erb
```diff
@@ -42,7 +42,12 @@ skipx
 <% else -%>
   <% dhcp = !@static -%>
 <% end -%>
+
+<% if rhel_compatible && os_major == 5 -%>
```
domcleal: This file's for RHEL only, so it doesn't define rhel_compatible.  Just having the os_major check should suffice.

PR #179, opened 2015-05-08
EL5 remove device=macaddr

NuxRo:
Remove "--device=$macaddr" from the "network" kickstart section for EL5 as it does not support it.

PR #180, opened 2015-05-09
"--device=<mac-address>" breaks RHEL <6

### LINKED ISSUE #153: EL5 installation fails with "The provided network interface $MAC does not exist"
It looks like EL5 (5.11 anyway) doesn't support MAC addresses as identifiers in a `network` configuration stanza:

    network --bootproto dhcp --device=52:54:00:86:94:db --hostname el5.example.com

screenshot_el5 example com_2015-02-23_16 26 57

mmoll:
The idea is taken from #179, but implemented much shorter (@NuxRo, could you test this change here?) and closes #153.

@ares, could you review?

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -51,7 +51,7 @@ skipx
 <% dhcp = !@static -%>
 <% end -%>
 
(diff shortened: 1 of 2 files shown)


ares:
Looks good to me, not tested

NuxRo:
Tested, works fine, much more elegant solution.

mmoll:
I also tested with CentOS 5.11 and 6.6.

domcleal:
Thanks @mmoll and @NuxRo!

PR #181, opened 2015-05-20
refs #9782 - SUSE: usage of mediapath in PXELinux

mmoll:
Also see theforeman/foreman#2393.

If anybody wants to pour in their energy to workaround another SUSE specific oddity cleanly in Foreman core it would be very welcome, for now this hacky workaround should get most things going again.

@mattiasgiese, please test.

### DIFF
--- a/autoyast/PXELinux.erb
+++ b/autoyast/PXELinux.erb
@@ -6,4 +6,4 @@ DEFAULT linux
 
 LABEL linux
     KERNEL <%= @kernel %>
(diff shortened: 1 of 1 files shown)


domcleal:
:+1: untested.

mattiasgiese:
Trying to test it this evening.

mattiasgiese:
I have '@host.os.medium_uri(@host)' set in all my autoyast pxe templates.
It's a hacky workaround but we have to work with it until a better fix is implemented.

mmoll:
thanks, merged.

PR #182, opened 2015-05-21
Atomic kickstart and PXE

eLobato:


kickstart/PXELinux_rhel_atomic.erb
```diff
@@ -0,0 +1,17 @@
+<%#
+kind: PXELinux
+name: RHEL Atomic PXE
+description: Start RHEL Atomic from the bare metal (installer) version. Requires to previously
+             download the image and mount it at /pub in the Foreman server.  You can run
```
domcleal: Shouldn't this be pointing to an installation medium instead?

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,28 @@
+<%#
+kind: provision
+name: RHEL Atomic Kickstart
+%>
+lang <%= @host.params['lang'] || 'en_US.UTF-8' %>
+keyboard <%= @host.params['keyboard'] || 'us' %>
+timezone --utc <%= @host.params['time-zone'] || 'UTC' %>
+clearpart --all --initlabel
```
domcleal: part of the ptable?

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,28 @@
...
+part /boot --size=300 --fstype="ext4"
+part pv.01 --grow
+volgroup atomicos pv.01
+logvol / --size=3000 --fstype="xfs" --name=root --vgname=atomicos
+bootloader --timeout=3
+ostreesetup --nogpg --osname=rhel-atomic-host --remote=rhel-atomic-host --url=file:///install/ostree --ref=rhel-atomic-host/7/x86_64/standard
```
domcleal: version number hardcoded here, should use the OS

kickstart/PXELinux_rhel_atomic.erb
domcleal: If so, I think this could use the existing PXELinux file.

kickstart/provision_rhel_atomic.erb
ekohl: Same for arch? I can imagine at some point aarch64 could be supported.

eLobato:
Thanks for the comments @ekohl @domcleal , updated.

kickstart/PXELinux_rhel_atomic.erb
```diff
@@ -0,0 +1,14 @@
+<%#
+kind: PXELinux
+name: RHEL Atomic PXE
```
domcleal: This all looks like the standard one, can't it use that?

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,27 @@
+<%#
+kind: provision
+name: RHEL Atomic Kickstart
```
domcleal: Is this RHEL-specific or is it, like the oses list above, applicable to CentOS and Fedora?

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,27 @@
...
+<%= @host.diskLayout %>
+
+bootloader --timeout=3
+ostreesetup --nogpg --osname=rhel-atomic-host --remote=rhel-atomic-host --url=file:///install/ostree --ref=rhel-atomic-host/<%= @host.os.major %>/<%= @host.architecture %>/standard
+services --disabled cloud-init,cloud-config,cloud-final,cloud-init-local
+rootpw --iscrypted <%= root_pass %>
+
+%post
+(
+# report success back to satellite
```
domcleal: Foreman

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,27 @@
+<%#
+kind: provision
+name: RHEL Atomic Kickstart
+%>
+lang en_US.UTF-8
+keyboard us
+timezone America/New_York
+
+# Partition table should create /boot and a volume atomicos
+<%= @host.diskLayout %>
```
domcleal: Perhaps support the %pre/scripted stuff that's in the standard provision script too.

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,27 @@
+<%#
+kind: provision
+name: RHEL Atomic Kickstart
+%>
+lang en_US.UTF-8
+keyboard us
```
domcleal: parameterised usually, see the existing script

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,27 @@
...
+bootloader --timeout=3
+ostreesetup --nogpg --osname=rhel-atomic-host --remote=rhel-atomic-host --url=file:///install/ostree --ref=rhel-atomic-host/<%= @host.os.major %>/<%= @host.architecture %>/standard
+services --disabled cloud-init,cloud-config,cloud-final,cloud-init-local
+rootpw --iscrypted <%= root_pass %>
+
+%post
+(
+# report success back to satellite
+curl -s -o /dev/null --insecure <%= foreman_url %>
+) 2>&1 | tee /mnt/sysimage/root/install.post.log
```
domcleal: Is this path correct, isn't that usually for a nochroot %post?

kickstart/provision_rhel_atomic.erb
```diff
@@ -0,0 +1,27 @@
...
+
+exit 0
+
+%end
+
+reboot
```
domcleal: Normally goes into the header section

kickstart/disklayout_atomic.erb
```diff
@@ -0,0 +1,14 @@
+<%#
+kind: ptable
+name: Kickstart Atomic
+oses:
+- CentOS 7 Atomic
+- Fedora 22 Atomic
+- RedHat 7 Atomic
```
domcleal: Are these OS names going to work with foreman_templates?  Seems odd to have "Atomic" after the version number, not sure if it'll understand that.

kickstart/PXELinux_rhel_atomic.erb
dLobatog: Right, but then I'd have to define a new OS ("Atomic") which would encompass CentOS/Fedora/RHEL Atomic on core, would that be preferred?

kickstart/disklayout_atomic.erb
dLobatog: It seems to understand CentOS-Atomic, I'll change it to that

eLobato:
@ekohl @domcleal Updated to be more generic about any Atomic system not just RHEL

kickstart/PXELinux_rhel_atomic.erb
domcleal: Sorry, I'm not sure I follow.  I mean, this template looks like the default Kickstart PXELinux file, can any atomic OS definition just use that existing template?

kickstart/provision_atomic.erb
```diff
@@ -0,0 +1,36 @@
+<%#
+kind: provision
+name: Atomic Kickstart
```
ekohl, line 3: oses?

kickstart/PXELinux_atomic.erb
```diff
@@ -0,0 +1,14 @@
+<%#
+kind: PXELinux
+name: Atomic PXE
+description: Start Atomic from the bare metal (installer) image.
```
ekohl, line 4: oses?

dLobatog:
I've added `oses` everywhere, but I'm not sure how useful it'll be though.

rothgar:
This didn't work for me on Satellite 6.1 beta

`initrd=<%= @host.medium.path %>isolinux/initrd.img`

needed to be changed to

`initrd=<%= @initrd %>`

I noticed it from watching the logs and the host would always fail to download the initrd.img file even though the URL was correct. Then I noticed foreman would wget the file in /var/lib/tftpboot/boot for me and was trying to load from the local disk.

I'm not sure if there's another way to force loading the initrd from an http source but that's what I found.

ares:
@eLobato could you please reopen against develop branch? we renamed master to develop since we started branching this repo per foreman releases

PR #185, opened 2015-06-08
Fixes #10072: an 'admin' user with sudo rights is now being configured for cloud-init setups.

dmitri-d:
This allows for spcification of non-clear text password for root-like administrative account. Part of support for atomic vm provisioning.


cloudinit/userdata_cloudinit.erb
```diff
@@ -21,8 +24,21 @@ hostname: <%= @host.shortname %>
...
+lock-passwd: false
+passwd: <%= @host.root_pass %>
+
 <%# Contact Foreman to confirm instance is built -%>
 phone_home:
- url: <%= foreman_url('built') %>
- post: []
- tries: 10
+url: <%= foreman_url('built') %>
+post: []
```
domcleal: seems many lines have lost their YAML hash indentation?

dmitri-d:
@domcleal: fixed!

ekohl:
I wonder if this should be hidden behind an if.

dmitri-d:
Not sure what could be used for a flag. I think I'd rather have a dedicated template rather then a flag/conditional check if this isn't needed in general.

domcleal:
We'd normally use conditionals in templates so we don't have to seed so many.  If it's specifically needed for Atomic somehow, then put it behind a check for the OS name?

dmitri-d:
With cloud-init the issue is that if one needs to change the password on an existing account (including root or root-like ones), it will have to be specified in clear-text. This is not distribution-specific, but probably more influenced by usage scenarios, hence my question. 

Having said that, the old root (or `centos`, `fedora`, or what have you) account is still present and available, this only creates an additional account.

ares:
@witlessbird could you please reopen against develop branch? we renamed master to develop since we started branching this repo per foreman releases

PR #189, opened 2015-06-11
Fixes #10072: an 'admin' user with sudo rights is now being configured for cloud-init setups.

dmitri-d:
This allows for spcification of non-clear text password for root-like administrative account. Part of support for atomic vm provisioning.

### DIFF
--- a/cloudinit/userdata_cloudinit.erb
+++ b/cloudinit/userdata_cloudinit.erb
@@ -10,6 +10,9 @@ oses:
 - Fedora 17
 - Fedora 18
 - Fedora 19
+- Fedora 20
+- Fedora 21
+- Fedora 22
 - Debian 6.0
 - Debian 7.
 - Ubuntu 10.04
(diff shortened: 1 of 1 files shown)


ekohl:
Trivial nitpick: typo in spcification in the commit message

dmitri-d:
@ekohl: fixed.

ekohl:
For reference, this replaces https://github.com/theforeman/community-templates/pull/185

I'm still not quite sure if everyone wants this by default. Maybe make passwordless sudo configurable and default to password sudo?

dmitri-d:
Not sure re: configurable passwordless sudo -- it's pretty easy to change the template if someone decides they don't want authentication. Probably makes sense to ship a slightly more secure version of the template with authenticated sudo, unless there are objections?

ekohl:
I think secure-by-default and have users modify the templates is fine.

dmitri-d:
@ekohl: removed `NOPASSWD` from `admin` user sudo config.

ekohl:
:+1:

domcleal:
Thanks @witlessbird.

PR #190, opened 2015-06-11
Fixes #10897 - Atomic kickstart and PXE

dLobatog:


kickstart/provision_atomic.erb
```diff
@@ -0,0 +1,40 @@
...
+<% else -%>
+<%= @host.diskLayout %>
+<% end -%>
+
+
+bootloader --timeout=3
+<% if @host.operatingsystem.name =~ /.*Fedora.*/ -%>
+ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=<%= @host.medium.path %>/content/repo/ --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
+<% else -%>
+ostreesetup --nogpg --osname=rhel-atomic-host --remote=rhel-atomic-host --url=file:///install/ostree --ref=rhel-atomic-host/<%= @host.os.major %>/<%= @host.architecture %>/standard
```
ekohl, line 25: Not that familiar with Atomic, but why does Fedora have a variable medium path where others have it hardcoded?

ekohl: Not that familiar with Atomic, but why does Fedora have a variable medium path where others have it hardcoded?

kickstart/PXELinux_atomic.erb
```diff
@@ -0,0 +1,18 @@
...
+kind: PXELinux
+name: Atomic PXE
+description: Start Atomic from the bare metal (installer) image.
+oses:
+- CentOS-Atomic 7
+- Fedora-Atomic 22
+- RedHat-Atomic 7
+%>
+
+DEFAULT pxeboot
```
domcleal: Is there any necessary difference between this file and the standard?

kickstart/provision_atomic.erb
dLobatog, line 25: RHEL Atomic lives on the Red Hat CDN and for the moment the way to deploy it with Foreman seems to be to download it and mount the image, so `file:///` works. Fedora can use local storage or a remote URL to fetch the tree as it's publicly accessible.

kickstart/PXELinux_atomic.erb
dLobatog: I could add another case to the if statement that catches 'Atomic' on the name if that's what you're referring to? Otherwise I can't set the kernel or initrd instance variables without creating an specific Atomic OS family. It's almost identical to the default template for Fedora 16+ except for the `repo=` part and the paths for initrd/kernel.

domcleal: repo= seems like a good thing to put behind a conditional based on the name, yep.  I don't understand why the vmlinuz/initrd would be different though, why's that?  Shouldn't Foreman manage those from the installation medium?

kickstart/PXELinux.erb
```diff
@@ -24,6 +24,8 @@ LABEL linux
     APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac
     <% elsif @host.operatingsystem.name != 'Fedora' and @host.operatingsystem.major.to_i >= 7 -%>
     APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> network ks.sendmac
+    <% elsif @host.operatingsystem.name =~ /.*Atomic.*/ -%>
+    APPEND initrd=<%= @host.medium.path %>isolinux/initrd.img repo=<%= @host.medium.path %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac
```
dLobatog: @domcleal
Following up on your [previous comment](https://github.com/theforeman/community-templates/pull/190#discussion_r32411890), I think I can possibly do without the KERNEL directive and use `vmlinuz` (not tested yet though). 
`initrd` is a bit of a different beast as Red hat in Foreman will try to fetch it from `boot/osfullname-arch/initrd.img`, which isn't Atomic's path.

domcleal: That's something to do in Foreman core then by way of an OS family or changes to the OS classes, so it can download the right initrd to the TFTP proxy.  If we start changing templates to ignore the TFTP proxy and go straight to the installation media then we're going to have logic about the layout of installation media in two places.

kickstart/PXELinux.erb
```diff
@@ -24,6 +24,8 @@ LABEL linux
     APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac
     <% elsif @host.operatingsystem.name != 'Fedora' and @host.operatingsystem.major.to_i >= 7 -%>
     APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> network ks.sendmac
+    <% elsif @host.operatingsystem.name =~ /.*Atomic.*/ -%>
```
dmitri-d: This line fails when safemode is turned on. `String#match` works however.

kickstart/provision_atomic.erb
```diff
@@ -0,0 +1,40 @@
...
+# Partition table should create /boot and a volume atomicos
+<% if @dynamic -%>
+%include /tmp/diskpart.cfg
+<% else -%>
+<%= @host.diskLayout %>
+<% end -%>
+
+
+bootloader --timeout=3
+<% if @host.operatingsystem.name =~ /.*Fedora.*/ -%>
```
dmitri-d: This also fails when safemode is on.

kickstart/PXELinux.erb
dmitri-d: Also, I think the order of checks should be different -- I have both 'Fedora' and 'Atomic' in the os name, which results in Fedora kernel parameters being used.

kickstart/PXELinux.erb
dmitri-d: This will also fail when safemode is on: `Host#medium` isn't accessible from the jail, and neither is `Medium#path`

kickstart/provision_atomic.erb
```diff
@@ -0,0 +1,40 @@
...
+ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=<%= @host.medium.path %>/content/repo/ --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
```
dmitri-d: This fails when safemode is on.

dLobatog:
Updated with changes that hopefully make it run in Safemode. To be followed up with a PR to core enabling `@initrd` and `@atomic_repo` for Atomic

dLobatog:
Actually I could do with just prepending `isolinux` to initrd & using `media_uri`. Up for review again, no need to touch core :smile:

kickstart/PXELinux.erb
```diff
@@ -37,7 +37,11 @@ DEFAULT linux
 LABEL linux
     KERNEL <%= @kernel %>
     <% if @host.operatingsystem.name == 'Fedora' and @host.operatingsystem.major.to_i > 16 -%>
-    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac <%= options %>
+      <% if @host.operatingsystem.name.match(/.*atomic.*/i) %>
+        APPEND initrd=isolinux/<%= @initrd %> ks=<%= foreman_url('provision') %> repo=<%= @host.operatingsystem.media_uri(@host) %>  ks.device=bootif network ks.sendmac <%= options %>
```
dmitri-d: should be `Operatingsystem#medium_uri` instead of `#media_uri`

kickstart/PXELinux.erb
```diff
@@ -37,7 +37,11 @@ DEFAULT linux
 LABEL linux
     KERNEL <%= @kernel %>
     <% if @host.operatingsystem.name == 'Fedora' and @host.operatingsystem.major.to_i > 16 -%>
-    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac <%= options %>
+      <% if @host.operatingsystem.name.match(/.*atomic.*/i) %>
```
domcleal: Is this meant to be within the Fedora block?  The other templates appear to apply to other OSes.

kickstart/disklayout_atomic.erb
```diff
@@ -0,0 +1,14 @@
+<%#
+kind: ptable
+name: Kickstart Atomic
+oses:
+- CentOS-Atomic 7
```
domcleal: Are these OS names correct, based on what Facter and Foreman imports?

kickstart/disklayout_atomic.erb
```diff
@@ -0,0 +1,14 @@
+<%#
+kind: ptable
+name: Kickstart Atomic
```
domcleal: usually we keep "default" in the name

kickstart/provision_atomic.erb
```diff
@@ -0,0 +1,40 @@
+<%#
+kind: provision
+name: Atomic Kickstart
```
domcleal: ditto, default

kickstart/provision_atomic.erb
```diff
@@ -0,0 +1,40 @@
...
+<% else -%>
+<%= @host.diskLayout %>
+<% end -%>
+
+
+bootloader --timeout=3
+<% if @host.os.name.match /.*fedora.*/i -%>
+ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=<%= @host.operatingsystem.medium_uri(@host) %>/content/repo/ --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
+<% else -%>
+ostreesetup --nogpg --osname=rhel-atomic-host --remote=rhel-atomic-host --url=file:///install/ostree --ref=rhel-atomic-host/<%= @host.os.major %>/<%= @host.architecture %>/standard
```
domcleal, line 25: Use of "rhel" here, is that applicable to CentOS?

dmitri-d, line 25: Not 100% sure, but I don't think the contents of `osname` and `remote` options has any effect on anything in this case.

kickstart/disklayout_atomic.erb
dLobatog: These OSs can't run Puppet for the moment, so they're kind of irrelevant for the moment. Since the names cannot contain spaces I chose to hyphenate Atomic.

dLobatog:
After some testing this is what I got to:
- I removed the partition table, autopart (default partition table) works fine.
- Fedora/CentOS/RHEL include a proper vmlinuz/initrd in their Atomic isos. Unfortunately they don't have it in ther mirrors, so fully provision from the official mirrors is impossible for the moment. I tried using regular vmlinuz/initrd from the non-atomic versions but that doesn't work when provisioning Atomic. When `@host.params['atomic-upstream']` is true, the installation media is overridden to use the upstream mirror for ostree. 

The best way to test this is with the [Fedora Atomic installer image](http://download.fedoraproject.org/pub/fedora/linux/releases/22/Cloud_Atomic/x86_64/iso/Fedora-Cloud_Atomic-x86_64-22.iso) (or RHEL). I uncompressed it in /var/www/html/pub/fedora-atomic/ of a host, then added an installation media to Foreman with URL http://fedora-installationmedia/pub/fedora-atomic/. Same process goes for RHEL and CentOS, although there's a bug (I think) on the CentOS image that prevents the installation from completing.

dmitri-d:
looks good to me.

kickstart/PXELinux.erb
```diff
@@ -36,7 +36,9 @@ DEFAULT linux
 
 LABEL linux
     KERNEL <%= @kernel %>
-    <% if @host.operatingsystem.name == 'Fedora' and @host.operatingsystem.major.to_i > 16 -%>
+    <% if @host.operatingsystem.name.match(/.*atomic.*/i) %>
```
ohadlevy, line 39: nitpick: prefer case over elsif

domcleal, line 39: That'd be rather messy I think.

domcleal:
Merged, thanks @eLobato.

PR #194, opened 2015-07-07
FreeBSD: set hostname in chroot & use UTC for CMOS

mmoll:
- somehow having the hostname explicitely set is required for newer releases (10.2-PRERELEASE in my tests), maybe also puppet behavour changed here, but it doesn't harm anyway
- removing /etc/wall_cmos_clock means, the time is stored in UTC in the CMOS clock, which is a good thing for UNIX systems


ekohl:
Thanks!

mmoll:
BTW, I'd like to see this in 1.9 also, but don't know what's our policy for that with the new branches :)

domcleal:
I'd suggest you're free to cherry pick it (-x) to the 1.9-stable branch as you see fit (ditto for any other committer) and then I'll make a note to sync the changes before doing each release.

mmoll:
allright :100:

mmoll:
cherry-picked into 1.9-stable as 8e26ebe

PR #195, opened 2015-07-10
bootloader append

mikenowak:
The real reason for doing this is that quiet results in startup messages being hidden, which is not very useful for debugging purposes.

But it would also be useful to be able to add extra params.


ekohl:
Looks good to me. Could you squash the commits?

mikenowak:
@ekohl My apologies, this should be done now

ekohl:
@mikenowak you now have 4 commits. I think http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html is a good read on how to get it back to 1. If you can't manage, we can do it for you when we merge it.

mmoll:
it looks to me like the rebase did work, but the result was not force pushed.

mikenowak:
@ekohl - thanks for the instructions - I think I've done this now.

ekohl:
:+1:

domcleal:
Merged as a5d76fb, thanks @mikenowak!

PR #196, opened 2015-07-14
Adding selinux parameter

mikenowak:
Giving a choice to these of us that think selinux isn't worth the hassle


domcleal:
Nice, thanks @mikenowak!

PR #198, opened 2015-07-27
Templates for deploying xenserver. Additional manual work required.

pujan14:
This patch provides required templates to provision xenserver with some additional work. Additional work is linked in readme file.

### DIFF
--- a/xenserver/xenserver-answerfile.erb
+++ b/xenserver/xenserver-answerfile.erb
@@ -1,0 +1,21 @@
+<?xml version="1.0"?>
+  <installation mode="fresh" srtype="lvm">
+    <primary-disk gueststorage="<%= @host.params['gueststorage'] || 'yes' %>"><%= @host.params['primary-disk'] || 'sda' %></primary-disk>
+    <keymap>de</keymap>
+    <hostname><%= @host.shortname %></hostname>
+    <root-password type="hash"><%= root_pass %></root-password>
+    <source type="<%= @host.params['source-type'] %>"><%= @host.params['source-path'] %></source>
+    <admin-interface name="<%=@ host.primary_interface || 'eth0' %>" proto="static">
+      <ip><%= @host.ip %></ip>
+      <subnet-mask><%= @host.subnet.mask %></subnet-mask>
+      <gateway><%= @host.subnet.gateway %></gateway>
+    </admin-interface>
+    <name-server><%= @host.subnet.dns_primary %></name-server>
+    <name-server><%= @host.subnet.dns_secondary  %></name-server>
+    <ntp-server>pool.ntp.org</ntp-server>
+    <timezone>Europe/Berlin</timezone>
+    <time-config-method>ntp</time-config-method>
+    <script stage="installation-complete" type="url">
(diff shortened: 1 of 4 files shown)


xenserver/xenserver-pxe.erb
```diff
@@ -0,0 +1,5 @@
+default xenserver
+label xenserver
+kernel mboot.c32
+append boot/XenServer-6.5-x86_64-xen.gz dom0_max_vcpus=16 dom0_mem=752M com1=115200,8n1 console=com1,vga --- boot/XenServer-6.5-x86_64-vmlinuz xencons=hvc console=hvc0 console=tty0 answerfile=<%= foreman_url('provision')%> install --- boot/XenServer-6.5-x86_64-initrd.img
```
ekohl, line 4: I wonder if variables could be used here rather than hardcoding the version and arch.

mmoll:
Thanks, I also did think about starting this, I hope to have a patch ready to add XenServer as an OS to Foreman core, so we can use the templates more flexible.

xenserver/xenserver-pxe.erb
pujan14, line 4: I am not sure if that's a good idea. General work flow for this setup to work is.
- Copy boot files from iso to tftp boot folder
- Create an http share and extract iso to that share (required for installation)
- Use these templates in foreman

I am not sure including this in foreman is a good idea or not. If it becomes part of standard foreman, there should be some sort of wizard or something that takes in iso and creates http share and copies required files into tight places.

pujan14:
@mmoll are you working on patch to add XenServer as an OS to foreman core? If so I would like to help (not sure how, testing maybe?)

xenserver/xenserver-pxe.erb
ekohl, line 4: In Foreman you have a version, so if the guide describes to download a version and then enter the correct version, I think it's much more flexible.

pujan14, line 4: Original files from ISO are named vmlinux.gz, vmlinuz and xen.gz. I renamed them on foreman host to right convention. User can copy them as it is and just use same name in template.

mmoll:
testing will be required for sure :)

xenserver/xenserver-answerfile.erb
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+  <installation mode="fresh" srtype="lvm">
+    <primary-disk gueststorage="<%= @host.params['gueststorage'] || 'yes' %>"><%= @host.params['primary-disk'] || 'sda' %></primary-disk>
+    <keymap>de</keymap>
+    <hostname><%= @host.shortname %></hostname>
+    <root-password type="hash"><%= root_pass %></root-password>
+    <source type="<%= @host.params['source-type'] %>"><%= @host.params['source-path'] %></source>
+    <admin-interface name="<%=@ host.primary_interface || 'eth0' %>" proto="static">
```
jonaz, line 8: synax error here? @ should be before host.

mmoll:
continuation in #250

PR #199, opened 2015-07-31
Blacklist option for Red Hat and Debian PXELinux

lzap:
This introduces new host param `blacklist` which renders for Red Hats and
Debian/Ubuntu. Also I have refactored Debian PXELinux template variables so
it's more consistent now.


mmoll:
díky @lzap!

PR #200, opened 2015-08-14
Converge preseed disklayouts

sathieu:
No default behavior change, unless the partitioning-\* parameters are
set.

New parameters:
- partitioning-expert-recipe
- partitioning-vg-name

Parameters ported to LVM ptable:
- partitioning-method (defaults to LVM)
- partitioning-recipe (defaults to atomic)
- partitioning-filesystem (defaults to ext4)

### DIFF
--- a/preseed/README.md
+++ b/preseed/README.md
@@ -14,8 +14,15 @@ Tested on:
 
 The templates use some Host Parameters to contol the flow of the template. These are:
 
-* install-disk: What device to install to (default: /dev/sda | /dev/vda)
-* enable-puppetlabs-repo: Add the Puppet Labs APT repo to the APT sources during install (default: no)
-* enable-saltstack-repo: Add the SaltStack APT repo to the APT sources during install (default: no)
-* salt_master: SaltStack Master (default: empty)
-* salt_grains: Salt client specific information, like facter (default: empty)
+* `install-disk`: What device to install to (default: `/dev/sda /dev/vda /dev/xvda`, the first available is picked)
+* `partitioning-method`: `regular` (default for `Preseed default`), `lvm` (default for `Preseed default lvm`) or `crypto`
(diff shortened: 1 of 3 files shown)


sathieu:
I'm tempted to drop the LVM disk layout now, but this would need a warning in the release notes.

sathieu:
The current LVM expert recipe is out of date from upstream (http://anonscm.debian.org/cgit/d-i/partman-auto.git/tree/recipes/multi):
- doesn't support EFI
- /usr should not be on a separate partition
- doesn't use $default_filesystem

domcleal:
> I'm tempted to drop the LVM disk layout now, but this would need a warning in the release notes.

I wouldn't worry, we now branch this repo and we don't delete any templates when Foreman is upgraded from say 1.9 to 1.10 - we only add templates.  So deleting templates is fine I'd say, if it's beyond useful.

sathieu:
I'm also tempted to change partitioning_method to lvm. @mmoll What do you think?

mmoll:
The layouts work fine, however using a variable to switch between LVM and non-LVM is something that's not fitting into Foreman's workflow, IMHO.

I would propose to delete the "Preseed custom LVM" and instead introduce a new, more upstream-inspired "Preseed default LVM" and continue using "Preseed default" as non-LVM layout. The vg parameters are only relevant for LVM template then and the expert-recipe can go then.

@sathieu your opinion please? :)

sathieu:
I think I've implemented something that fit with your description.

mmoll:
This is working great :smiley:
Could you add a short documentation of the parameters in `preseed/README.md`?

sathieu:
@mmol, another commit for updated README.

mmoll:
@sathieu perfect, plase squash all your commits together into one

sathieu:
@mmol done

mmoll:
merged, merci beaucoup @sathieu!

PR #201, opened 2015-08-21
Fixes #11440 - Network bonding is not working properly when provision…

ruzickap:
Removing "DEFROUTE=no" which prevents CentOS/RHEL server to properly set default gw when on the bonded interface.

Issue is described here: http://projects.theforeman.org/issues/11440


domcleal:
Thanks for re-opening.  @ares may remember the history better than me, but my suspicion is that it's here to stop a default route being added to non-"primary" interfaces.  If that's the case, then we'd probably wrap it in a `bond.primary?` conditional.

ares:
Sorry, my memory is not as good as yours but what you say makes sense to me. @ruzickap I suppose you use passive LACP so anaconda uses eth0 and after reboot you expect bond0 to be used instead. So the line should look like following

```
DEFROUTE=<%= bond.primary? ? 'yes' : 'no' -%>
```

ruzickap:
Hello.

I changed it like you suggested. 
Please check it.

Thank you

PetrR

domcleal:
Merged, thanks @ruzickap!

PR #204, opened 2015-09-16
Fixes #11440 - Network bonding is not working properly when provisioning CentOS/RHEL servers because the default gateway is not set

ruzickap:
Hello.

The previous commit related to this issue (http://projects.theforeman.org/issues/11440) was wrong: https://github.com/theforeman/community-templates/pull/201.

I modified the file and tested it with real hardware and this commit should fix it.

Regards

PetrR


mmoll:
díky @ruzickap!

PR #206, opened 2015-09-18
Don't specify Debian and Ubuntu version explicitely

sathieu:
This make future releases and non-LTS Ubuntu releases automatically associate, and also fixes divergence between files

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,12 @@ determine information abut the template. The metadata is read until the first no
 line, and contains, the template name, template kind, and any appropriate OS
 associations. For example:
 
-    #kind: provision
-    #name: My Preseed
+    #kind: user_data
+    #name: My Userdata
     #oses:
-    #- Debian 6.0
-    #- Debian 7.
+    # - CentOS 7
+    # - Debian
+    # - Ubuntu
 
 Because all the data is contained in the metadata, filenames are arbitrary, however,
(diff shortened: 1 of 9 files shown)


sathieu:
@mmoll : This one's for you ;-)

domcleal:
Fine by me.  I don't think there's much need to list the versions since we're doing well at keeping compatibility across versions within a single template.

It looks like this will work with foreman_templates, since it does a regex match on the start of the OS label, but I've not tested it.

sathieu:
Anything needed before the merge?
I can assure that it works.

mmoll:
Not really, something regarding association seems to be broken for my test-setup in general, and so this PR went off my radar, sorry for that!

The only question I was asking myself was if that should also be applied to other OSes also, and personally I think it should. @domcleal? @GregSutcliffe?

domcleal:
> The only question I was asking myself was if that should also be applied to other OSes also, and personally I think it should.

Yes, it probably should if @sathieu's willing, else we can sort it out after.

GregSutcliffe:
> Yes, it probably should if @sathieu's willing, else we can sort it out after.

+1.

sathieu:
I'm willing to do this, but it'll take time as I'm really busy.

So, I prefer that you merge it and somebody else do the remaining work.

thanks

mmoll:
Sure thing, merged as 4445d3f, merci @sathieu!

PR #207, opened 2015-10-07
Updated to etcd v2 and corrected link to cores_cloudconfig snippet

johscheuer:
@endyman and I updated the etcd version to v2 since etcd v0 is deprecated. Also we added support for static networking on the CoreOS side.


mmoll:
@eLobato could you have a look?

snippets/coreos_cloudconfig.erb
```diff
@@ -4,14 +4,44 @@ kind: snippet
...
 <% end -%>
-          addr: <%= @host.ip %>:4001
-          peer-addr: <%= @host.ip %>:7001
+          advertise-client-urls: http://<%= @host.ip %>:2379,http://<%= @host.ip %>:4001
+          initial-advertise-peer-urls: http://<%= @host.ip %>:2380,http://<%= @host.ip %>:7001
+          listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
+          listen-peer-urls: http://0.0.0.0:2380,http://0.0.0.0:7001
         units:
-          - name: etcd.service
+<% subnet = @host.subnet -%>
```
dLobatog: Mind to submit the networking changes separately? I think they don't exactly belong in a set of etcd2 changes and while the etcd2 changes are fine, this needs more thorough testing :smile:

dLobatog:
:+1: for the etcd2 changes. Are the networking changes essential? I have only tested it with dhcp. Do you think moving static networking to a different PR would make sense? I'd be glad to test that and merge the etcd2 changes right away.

@johscheuer do you want to maintain etcd/CoreOS changes in Foreman after this pull request is merged? You're probably the one around with the most experience on this.

I think only @ohadlevy can give commit access, does the above sound :ok_hand: to you?

snippets/coreos_cloudconfig.erb
johscheuer: Of course :) We just wanted to push these changes in one PR for simplicity but you are right :+1:

johscheuer:
If you want to set a static IP address yes otherwise no :) I think it would make sense because then we have separated these two things (etcd update + static networking). I will remove the static networking from this PR :+1: 

@eLobato and @ohadlevy would be a honor for me :)

dLobatog:
@mmoll :+1: to this PR (notice static networking is removed)

mmoll:
Danke @johscheuer, also cherry-picked as 299286a to 1.10-stable, please open a PR for the networking changes. :)

PR #209, opened 2015-10-09
Windows templates from wimaging-ng 

helge000:
Lets try get the discussion started to add templates I am using to provision Windows Hosts. Currently, the function I use to render `root_pass` does not play nicely with `safe_mode = true`. 

Suggestions very welcome!

### DIFF
--- a/waik/waik_finish.erb
+++ b/waik/waik_finish.erb
@@ -1,0 +1,95 @@
+<%#
+kind: finish
+name: WAIK default
+# Parameters are expected to be set in Foreman (globally or per group/host)
+params:
+- windowsLicenseKey: ABCDE-ABCDE-ABCDE-ABCDE-ABCDE # Valid Windows license key
+- windowsLicenseOwner: Company, INC # Legal owner of the Windows license key
+- localAdminiAccountDisabled: false
+- ntpSever: time.windows.com,other.time.server
+- domainAdminAccount: administrator@domain.com # use this account to join domain
+- domainAdminAccountPasswd: Password for the domain Admin account
+- computerOU: OU=Computers,CN=domain,CN=com # Place the computer account in specified Organizational Unit
+- computerDomain: domain.com # domain to join
+-%>
+@echo off
+<% unless @host.params['localAdminiAccountDisabled'] -%>
+  echo Activating administrator
+  net user administrator /active:yes
+<% end -%>
+
+<% if @host.pxe_build? %>
+  set ctr=0
+  set nettimeout=10
+  set nwhost=<%= @host.puppetmaster %>
+
+  (echo Updating time)
+  (sc config w32time start= auto)
+  sc start w32time
+  :ntp_testip
+    ::ipconfig /renew
+    ping -n 1 %nwhost% > NUL
+    if %errorlevel% == 0 goto ntp_testip_ok
+    timeout /t 10 >NUL
+    set /a ctr=ctr+1
+    echo Retry network. Run %ctr% time(s)
+    if %ctr% GTR %nettimeout% (
+        echo Network error! Exiting.
+        exit 1
+        )
+    goto ntp_testip
+  :ntp_testip_ok
+
+  echo Ping to %nwhost% OK!
+
+  <% if @host.params['ntpSever'] %>
+    echo setting time server
+    w32tm /config /manualpeerlist:<%= @host.params['ntpSever'] %> /syncfromflags:manual /update
+  <% end %>
+
+  echo sync time
+  w32tm /resync
+  w32tm /resync
+
+  :: You can join your machine to the domain right here >
+
+  <% if @host.params['domainAdminAccount'] and @host.params['domainAdminAccountPasswd'] -%>
+  echo joining domain
+  powershell.exe -OutputFormat text -ExecutionPolicy remotesigned -command c:\deploy\joinDomain.ps1
+  <% end %>
+
+  :: < You can join your machine to the domain right here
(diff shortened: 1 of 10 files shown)


waik/waik_provision_unattend.xml.erb
```diff
@@ -0,0 +1,158 @@
...
+              <UILanguageFallback>en-US</UILanguageFallback>
+              <UILanguage>en-US</UILanguage>
+              <UserLocale>en-US</UserLocale>
+            <% end -%>
+        </component>
+        <component name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS" xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+            <UserAccounts>
+                <AdministratorPassword>
+                    <PlainText>false</PlainText>
+                    <Value><%= Base64.encode64(Encoding::Converter.new("UTF-8", "UTF-16LE",:undef => nil).convert(Base64.decode64(root_pass)+"AdministratorPassword")).delete!("\n").chomp %></Value>
```
helge000: This renders `root_pass` and encoding it in a realy "nice" Microsoft Windows way. Does not like `safe_mode`:
`Safemode doesn't allow to access 'constant' on Base64`

waik/waik_finish.erb
```diff
@@ -0,0 +1,98 @@
...
+#- Windows 6.1
+#- Windows 6.3
+# Parameters are expected to be set in Foreman (globally or per group/host)
+params:
+- windowsLicenseKey: ABCDE-ABCDE-ABCDE-ABCDE-ABCDE # Valid Windows license key
+- windowsLicenseOwner: Company, INC # Legal owner of the Windows license key
+- localAdminiAccountDisabled: false
+- ntpSever: time.windows.com,other.time.server
+- domainAdminAccount: administrator@domain.com # use this account to join domain
+- domainAdminAccountPasswd: Password for the domain Admin account
```
sathieu: The user to join the domain doesn't need to be admin.
Also, other templates don't use camelCase for params.

sathieu:
Can you add a README with a quick howto too?

helge000:
@sathieu: for a hopefully comprehensive guide on how these templates are supposed to work is in my [original project](https://github.com/helge000/wimaging-ng/blob/master/README.md).
[This doc](https://github.com/helge000/wimaging-ng/blob/master/doc/foreman.md#iii-add-provision-templates) discusses the templates

Also, for the record: I opened [feature 12128](http://projects.theforeman.org/issues/12128) against foreman.org to get these docs merged to the official foreman docs. Hopefully I have time to open a PR this weekend.

waik/waik_finish.erb
helge000: @sathieu This dependents on how your domain is configured and witch [GPOs are in place](https://support.microsoft.com/en-us/kb/243327). A domain admin always work (for instance, my users may not add any computer to the domain...)
But I see your point, this would better be named `domainJoinAccount` suppose?

About these or any other parameters: To keep I will refactor them once we're at point near  merging this PR ;)

helge000:
Refernce: [PR for the docs](https://github.com/theforeman/theforeman.org/pull/450)

waik/waik_provision_unattend.xml.erb
helge000: Has anyone an idea how to solve this? Obviously, this does not work as well:

```
 <PlainText>true</PlainText>
   <Value><%= "#{Base64.decode64(root_pass)}AdministratorPassword".delete("\n").chomp %></Value>
```

If there is something like `safe_mode?`, I could check and render some host param as fallback. (Not very nice, though)

domcleal: I don't think there's an easy way to see if it's enabled.  Ideally we would add a helper into lib/foreman/renderer.rb or app/models/concerns/host_template_helpers.rb to help out with the decoding.

helge000: @domcleal Another idea:

We ship two templates for this; one working with safe mode, the other requiring it to be off. The one for safe mode would render `root_pass` directly. The user would need to append the string `AdministratorPassword` to the root password.

If `safe_mode?` is available at any point it could be merged again.

I still like the encoder better because it is more resilient (eg, the template will not render if password is encoded in a wrong way). 

Also, since Base64 password encoding is there, maybe there will be `Plain` in the future ;)

domcleal: I'd prefer not to ship a template that only works when safemode's off, as I strongly recommend people always leave it enabled - it's such a security risk if you have non-admin users otherwise.  So my preference is to do something/anything, in Foreman to make it possible to run in safemode.

helge000: Ok, I change this to work only in safe mode, see _idea_ above.

I leave this line commented out for reference in the template and change docs accordingly?

domcleal: Yep, okay.  Perhaps file a ticket in Foreman too for a helper so we know precisely what we need (I may be able to help, or find help to implement it) and in future we can change it to use that.

helge000:
Can this PR be closed? In the mean time, foreman docs reference to these templates

UXabre:
Perhaps this PR can be closed now that we have the other windows templates? This in an effort to clean-up stale PRs?

PR #210, opened 2015-10-17
Added the possibility to set a static network address on CoreOS

johscheuer:
Added the possibility to set a static network address on CoreOS. Before this patch network could only be set by DHCP.

Thanks to @endyman


mmoll:
seems to work fine :+1:

mmoll:
Also merged as 9b05a85 for 1.10-stable, Danke @johscheuer und @endyman!

PR #218, opened 2015-11-22
Fixes param_true? issue from PR #212 and ubuntu 12 static networking

dgagnon:
fixed remaining templates.


dgagnon:
forgot to commit

PR #222, opened 2015-11-24
Implement param_true/false? methods on host struct to fix tests

domcleal:
Fixes GH-215

Sorry about breaking tests when merging, I completely overlooked it.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -22,6 +22,15 @@ class FakeStruct
   def to_s
     as_string
   end
+
+  # Roughly equivalent to HostCommon#param_true?/false in Foreman core
+  def param_true?(name)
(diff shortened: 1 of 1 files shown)


mmoll:
could you add `sudo: false` to .travis.yml also? :)

domcleal:
Running, should fix that deprecation warning.

mmoll:
adding

``` yaml
addons:
  apt:
    packages:
    - python-urlgrabber
```

should fix the new error.

domcleal:
More occurred as it then installs pykickstart using sudo into the root directory.  I'll have a look at that separately in my own time :)

mmoll:
merged, thanks @domcleal!

PR #223, opened 2015-11-24
Switch Travis CI to container infra

domcleal:
~~~The runtime of the test is quite a bit slower (at about 17 to 34s) due to the additional network-based installations and setting up the virtualenv manually (since we can't use the Travis Python environment), but~~~ it should be quicker to start the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
+sudo: false
 language: ruby
 rvm: 1.9.3
 install:
-  - sudo apt-get -y install python-urlgrabber
-  - git clone --depth 1 -b r1.99.48-1 git://git.fedorahosted.org/git/pykickstart.git
-  - cd pykickstart && sudo make install DESTDIR=/
-env: KSVALIDATOR=/usr/local/bin/ksvalidator
+  - pip install --user six requests
+  - git clone --depth 1 -b r2.20-1 git://github.com/rhinstaller/pykickstart.git
+  - (cd pykickstart && pip install --user .)


.travis.yml
```diff
@@ -1,7 +1,14 @@
...
 install:
-  - sudo apt-get -y install python-urlgrabber
+  - virtualenv py
+  - source py/bin/activate
+  - pip install pycurl
+  - pip install urlgrabber
   - git clone --depth 1 -b r1.99.48-1 git://git.fedorahosted.org/git/pykickstart.git
-  - cd pykickstart && sudo make install DESTDIR=/
-env: KSVALIDATOR=/usr/local/bin/ksvalidator
+  - (cd pykickstart && easy_install .)
```
ekohl: You could try `pip install --user .` (and avoid the whole virtualenv).

.travis.yml
```diff
@@ -1,7 +1,14 @@
...
+  apt:
+    packages:
+      - python-virtualenv
 install:
-  - sudo apt-get -y install python-urlgrabber
+  - virtualenv py
+  - source py/bin/activate
+  - pip install pycurl
+  - pip install urlgrabber
   - git clone --depth 1 -b r1.99.48-1 git://git.fedorahosted.org/git/pykickstart.git
```
ekohl: Maybe switch to the current location as well: https://github.com/rhinstaller/pykickstart

.travis.yml
domcleal: That works well, thanks!

.travis.yml
domcleal: Changed, and also updated it to the latest version.

ekohl:
:+1:

.travis.yml
ekohl: Any reason not to follow master?

mmoll:
merged, thanks @domcleal!

PR #224, opened 2015-11-24
Removed Docker tcp socket

johscheuer:
I think there is no need to expose the Docker TCP socket. Most people will login into the Machine and run their containers during their SSH session or will use a cluster scheduler like [Kubernetes](https://github.com/kubernetes/kubernetes), [Mesos](https://mesos.apache.org), [Swarm](https://docs.docker.com/swarm), [fleet](https://github.com/coreos/fleet) or something else, so that there is no need to have a Docker socket exposed.

 From the [Docker docs](https://docs.docker.com/engine/userguide/basics):
"Warning: Changing the default docker daemon binding to a TCP port or Unix docker user group will increase your security risks by allowing non-root users to gain root access on the host. Make sure you control access to docker. If you are binding to a TCP port, anyone with access to that port has full Docker access; so it is not advisable on an open network."


mmoll:
:+1: visually, @dLobatog, what do you think?

dLobatog:
:-1: , I don't think it's desirable on production envs but without it now you would have to turn it on manually, which in a development environment is additional pain for no gain

domcleal:
Parameterise it, with the default as off for security?

ekohl:
What @domcleal said

johscheuer:
@domcleal sounds good :+1:

mmoll:
:+1:

mmoll:
merged, danke @johscheuer!

PR #225, opened 2015-11-24
Remove deprecated etcd ports

johscheuer:
I think we should remove the old deprecated etcd ports. People who spin up a new CoreOS cluster don't need these deprecated ports (and should use the new [standardized ports](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=etcd)). By removing the old ports we make it easier for new people to use the right ports in their setup.


mmoll:
according to the docs I could find, this is perfectly fine, merging...

mmoll:
Danke @johscheuer!

PR #227, opened 2015-12-12
add support for iPXE with suse

ghost:
Need review for other versions of SLES, or opensuse. That should work OOTB, if not please take a look at http://projects.theforeman.org/issues/12791


mmoll:
Can this work without the $arch bug getting fixed in Foreman core first?

ghost:
Yes, for example i've hardcoded the arch in https://github.com/theforeman/foreman/blob/develop/app/models/operatingsystems/suse.rb#L13

```
  def pxedir
    "boot/x86_64/loader"
  end
```

mmoll:
I can't get it to work with OpenSUSE 13.2 with a DHCP Subnet. After manually selecting the NIC (there's only one) it seems to continue correctly.

mmoll:
could you please change this to have no menu to allow an unattented installation? In a quick test with OpenSUSE 13.2 after selecting the "boot with DHCP" option the install worked. With the "boot with static IP" option, I got the NIC selection menu of yast again.

autoyast/iPXE.erb
```diff
@@ -0,0 +1,65 @@
...
+item reseau Configure network
+item hostinfo Infos system
+item shell Shell access
+item reboot Reboot
+item shutdown Shutdown
+choose target && goto ${target}
+
+
+:static
+kernel <%=@host.url_for_boot(:kernel) %> splash=silent hostip=${ip}/24 gateway=${gateway} install=<%=@host.os.medium_uri(@host)%>  autoyast=<%= foreman_url('provision') %> text-mode=0 domain=<%=@host.domain %> Nameserver=<%=@host.subnet.dns_primary%> netsetup=hostip UseDHCP=0
```
mmoll: `/24` is a static value here that has to come from the Foreman subnet information.

ghost: Sorry for the late response. I had hardcoded `/24` because the static way of booting with my menu was attended to be used with the "Configure Network" option, and after a user would change the mask with the configure option, the mask would no longer be the one from the Foreman. Plus, iPXE does not provide a cidr variable for the mask (e.g `/24`), only the traditional way `${netmask}`. So I had hard coded the value

ghost: ... forgot to say AFAIK, suse / opensuse only allow to configure the IP address and the mask with this form `xxx.xxx.xxx.xxx/xx`

mmoll:
Seems to work for me... but everything regarding `static` is not used in the template right now? Also, please change the name in the template to `AutoYaST default iPXE`.

mmoll:
merged, merci @SkySymbol!

PR #228, opened 2016-01-04
Fixes #12943 - interface vlan setting is considered too

ares:


mmoll:
Also merged as bf4b1e7 for 1.10, díky @ares!

PR #229, opened 2016-01-19
Add a conditional to set net.ifnames=0 on the kernel command line

hjensas:
If the compute resource provider is 'VMWare'.

Predictable network names are not predictable on VMWare, it is better
to fall back to the kernel default network device naming by adding <%= 'net.ifnames=0' if ['VMWare'].include?(@host.provider)%> to kernel command line in iPXE, PXELINUX and provisioning templates.

### DIFF
--- a/kickstart/PXELinux.erb
+++ b/kickstart/PXELinux.erb
@@ -37,11 +37,11 @@ DEFAULT linux
 LABEL linux
     KERNEL <%= @kernel %>
     <% if @host.operatingsystem.name.match(/.*atomic.*/i) %>
-    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision') %> repo=<%= @host.operatingsystem.medium_uri(@host) %> ks.device=bootif network ks.sendmac <%= options %>
+    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision') %> repo=<%= @host.operatingsystem.medium_uri(@host) %> ks.device=bootif network ks.sendmac <%= options %> <%= 'net.ifnames=0' if ['VMWare'].include?(@host.provider)%>
     <% elsif @host.operatingsystem.name == 'Fedora' and @host.operatingsystem.major.to_i > 16 -%>
-    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac <%= options %>
+    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> ks.device=bootif network ks.sendmac <%= options %> <%= 'net.ifnames=0' if ['VMWare'].include?(@host.provider)%>
     <% elsif @host.operatingsystem.name != 'Fedora' and @host.operatingsystem.major.to_i >= 7 -%>
-    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision')%> network ks.sendmac <%= options %>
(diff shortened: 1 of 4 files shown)


mmoll:
Setting this dependent on the compute resource doesn't feel right. Could you refactor this to use a global parameter like `disable-net-ifnames` or similar and set net.ifnames=0 when that parameter is set to `true`? See the `enable-puppetlabs-repo` param for an example.

mmoll:
@hjensas anything we can help to get this further?

hjensas:
Hi,

Actually net.ifnames are great for anything but VMWare. On oVirt, KVM etc the ethX are the default, on actual hardware the net.ifnames does tell a lot about which pysical NIC it is.

I really do not want to create a global parameter that will either disable net.ifnames for _all_ or _none_. Doing it for VMWare only does make sense, because consistent network names are broken on the VMWare side of things. This thread on VMWare forums discuss the issue with Consistent Network Naming on VMWare. https://communities.vmware.com/thread/478439?start=0&tstart=0

Regards
Harald

mmoll:
I do get the value of this change and also that it makes perfectly sense to disable ifnames on VMware.

However, I'm somewhat on the POLA (principle of least astonishment) side here, as RH/CentOS 7 is out for quite a while and our templates do net set this value since then. Also, there are situations where it can make sense to disable ifnames also in other environments (e.g. I disabled it on my Laptop), so having a possibility to disable it with one variable would IMHO bring more value in general.

Not having the possibility to set global paramters per compute resource (or CR type) in Foreman does make bringing in this change in a nice way hard, so I propose to split this PR:
1. bring in the possibility to disable ifnames per a variable (preferably also implement this for the preseed templates, idealy also for autoyast).
2. disable ifnames by default for VMware nodes (maybe not only for the VMware CR, but also for the machine type), but this needs some more discussion with other users, best on the -users list and also things like interface identifiers on foreman-discovery must be taken into account (yes, there are people using discovery for VMs).

@ekohl @domcleal your opinion?

ekohl:
I think splitting it makes sense. That allows us to backport the parameter to previous versions. In master I wouldn't mind making the default behave as the PR does now (unless explicitly overridden through the parameter).

mmoll:
@hjensas ping?

mmoll:
what's the status here?

sathieu:
FYI, I've just created a ticket on "My VMware" about this.

A more precise workaround would be to put `slot` before `onboard`. With `{/usr}/lib/systemd/network/90-vmware.link`:

```
[Match]
Virtualization=vmware

[Link]
NamePolicy=kernel database slot path onboard
MACAddressPolicy=persistent
```

But this should be done before network setup, which is a chicken-and-egg problem with response file from network.

sathieu:
@hjensas This also apply to Debian (and probably all dists). Can you update the preseed files too?

sathieu:
Latest udev (>= 229) workaround this by ignoring high onboard indices (>16383):

```
        /* Some BIOSes report rubbish indexes that are excessively high (2^24-1 is an index VMware likes to report for
         * example). Let's define a cut-off where we don't consider the index reliable anymore. We pick some arbitrary
         * cut-off, which is somewhere beyond the realistic number of physical network interface a system might
         * have. Ideally the kernel would already filter his crap for us, but it doesn't currently. */
        if (idx > ONBOARD_INDEX_MAX)
                return -ENOENT;


```

See https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L155 and https://github.com/systemd/systemd/commit/6c1e69f9456d022f14dd00737126cfa4d9cca10c.

I think this is the correct fix, and pushing it to the distros directly is the way forward.

mmoll:
closing as OP didn't react

PR #231, opened 2016-01-25
Fixes #13362 - Remove upstream parameter from Fedora Atomic kickstart

dLobatog:
Fedora Atomic can be provisioned now from upstream entirely,
removing the need for this parameter.

Before http://projects.theforeman.org/issues/13340 ,
we had to set up a local installation media for the PXE boot stuff. So
provisioning from upstream was optional. Now, we can default to
provision from upstream and anyone with a custom installation media can
change the kickstart to use it.

CentOS unfortunately still has to be provisioned through a local
installation media so I'm not changing that line.

### DIFF
--- a/kickstart/provision_atomic.erb
+++ b/kickstart/provision_atomic.erb
@@ -18,7 +18,10 @@ bootloader --timeout=3
 text
 
 <% if @host.os.name.match /.*fedora.*/i -%>
-ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=<%= @host.param_true?('atomic-upstream') ? "https://dl.fedoraproject.org/pub/fedora/linux/atomic/#{@host.os.major}/" : "#{@host.operatingsystem.medium_uri(@host)}/content/repo/" %> --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
+# Use @host.operatingsystem.medium_uri(@host)}/content/repo/ as the URL if you
+# have set up a local installation media for Fedora
+<% fedora_atomic_url = @host.params['atomic_refs_url'] || "https://dl.fedoraproject.org/pub/fedora/linux/atomic/#{@host.os.major}/" %>
+ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=<%= fedora_atomic_url %> --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
 <% elsif @host.os.name.match /.*centos.*/i -%>
 ostreesetup --nogpg --osname=centos-atomic-host --remote=centos-atomic-host --url=<%= @host.param_true?('atomic-upstream') ? "http://mirror.centos.org/centos/#{@host.os.major}/atomic/#{@host.architecture}/repo/" : @host.operatingsystem.medium_uri(@host) %> --ref=centos-atomic-host/<%= @host.os.major %>/<%= @host.architecture %>/standard
 <% else -%>


kickstart/provision_atomic.erb
```diff
@@ -18,7 +18,9 @@ bootloader --timeout=3
 text
 
 <% if @host.os.name.match /.*fedora.*/i -%>
-ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=<%= @host.param_true?('atomic-upstream') ? "https://dl.fedoraproject.org/pub/fedora/linux/atomic/#{@host.os.major}/" : "#{@host.operatingsystem.medium_uri(@host)}/content/repo/" %> --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
+# Use @host.operatingsystem.medium_uri(@host)}/content/repo/ as the URL if you
+# have set up a local installation media for Fedora
+ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=https://dl.fedoraproject.org/pub/fedora/linux/atomic/#{@host.os.major}/ --ref=fedora-atomic/f<%= @host.os.major %>/<%= @host.architecture %>/docker-host
```
ekohl: Why does it need to be hardcoded instead of the medium?

dLobatog: My bad - full context here https://github.com/theforeman/foreman/pull/3111 . In short, the medium URL where the PXE stuff is fetched from, is different from the URL where the kickstart trees for `ostreesetup` are fetched from.

ekohl: Won't it still be useful to be able to specify a mirror?

dLobatog: If you can deploy your own repository for kickstart trees as I explain here http://blog.daniellobato.me/unattended-deployments-of-fedora-and-rhel-atomic-with-foreman/ , then yeah. That used to be the only option to provision Atomic. I think it's fine to default to use the Fedora URL, and if you set up your own Atomic installation media then change the template. 

There's one comment in lines 21 and 22 with the command to use as the URL if there's a local Fedora Atomic installation media.

GregSutcliffe: Maybe it could be a parameter with a default, similar to what we do with the http-proxy{,-port} ones in preseed?

dLobatog: Not sure I follow, the default would be the Fedora Atomic URL or the local installation media?

GregSutcliffe: Default to what you have here, just gives @ekohl his option to override it via a parameter on the host. Something like

```
<% real-atomic-url = @atomic-url || "https://dl.fedora ..." -%>
```

should work? (note, psuedo-code, usual parameter existence checks apply ;P)

dLobatog:
@ekohl @GregSutcliffe Updated - now it defaults to upstream but can be overridden using the atomic_refs_url param.

dLobatog:
Updated - changed dashes for underscores in the variable name, apparently safemode doesn't like dashes in variable names.

mmoll:
We already have some dash containing variables like `time-zone` or `enable-puppetlabs-repo`, if this is a problem now, that would indicate a core problem.

domcleal:
Hyphens in parameter names are fine, but you can't put a hyphen in a local ERB variable name as that's invalid in Ruby.

mmoll:
When I want to test this with Fedora 23 I'm running into https://bugzilla.redhat.com/show_bug.cgi?id=1287870 - But I wouldn't be against merging anyway...

dLobatog:
@mmoll Interesting - I only tried with f22 and it provisioned all the way, maybe you want to try that one out?

mmoll:
Maybe I'm doing something wrong. I have `http://dl.fedoraproject.org/pub/alt/atomic/stable/Cloud_Atomic/$arch/os/` as install media and Fedora 22/23 with `Atomic Kickstart default` and `Kickstart default PXELinux` as templates with Foreman 1.10.1.

mmoll:
BTW, it would be nice to bring in that Atomic ptable here, also

dLobatog:
@mmoll I'm using the default Kickstart ptable that comes with foreman, that seems to work well. I think I always create Atomic vms with around 15GB HDD and 1GB memory. You can create a more specific layout for Atomic like this, but Anaconda should work with the default ptable. If it doesn't but this works, let me know: 

```
zerombr
clearpart --all --initlabel
part /boot --size=300 --fstype="ext4"
part pv.01 --grow
volgroup atomicos pv.01
logvol / --size=3000 --fstype="xfs" --name=root --vgname=atomicos
```

mmoll:
this is my rendered kickstart, is there any obvious error?

```
lang en_US.UTF-8
keyboard us
timezone --utc Europe/Berlin

# Partition table should create /boot and a volume atomicos
zerombr
clearpart --all --initlabel
part /boot --size=300 --fstype="ext4"
part pv.01 --grow
volgroup atomicos pv.01
logvol / --size=3000 --fstype="xfs" --name=root --vgname=atomicos

bootloader --timeout=3
text

# Use @host.operatingsystem.medium_uri(@host)}/content/repo/ as the URL if you
# have set up a local installation media for Fedora

ostreesetup --nogpg --osname=fedora-atomic --remote=fedora-atomic --url=https://dl.fedoraproject.org/pub/fedora/linux/atomic/22/ --ref=fedora-atomic/f22/x86_64/docker-host
services --disabled cloud-init,cloud-config,cloud-final,cloud-init-local
rootpw --iscrypted $5$/7VB7QVa$0nAVr8hysE0ZlTYIEuEHExmTHWglF98MzncyKf4tDEC

reboot

%post
(
# Report success back to Foreman
curl -s -o /dev/null --insecure http://foreman.local:80/unattended/built?token=0796657e-0a8f-4f32-bd62-628e2a234c49
) 2>&1 | tee /mnt/sysimage/root/install.post.log

exit 0

%end
```

dLobatog:
Looks like the one I use too :/

## 

Daniel Lobato Garcia

@dLobatog
blog.daniellobato.me
daniellobato.me

GPG: http://keys.gnupg.net/pks/lookup?op=get&search=0x7A92D6DD38D6DE30
Keybase: https://keybase.io/elobato

mmoll:
Well, no idea, as I never did provision Atomic before, I'm not against merging this anyway.

mmoll:
BTW, it looks like there's no install.img pulled after retrieving the kickstart, is that normal? Is there such an image on the ISO, by any chance?

mmoll:
What's the status here? I'm not against merging this if it's working for @dLobatog...

mmoll:
merged, gracias @dLobatog!

PR #234, opened 2016-02-05
add preseed_networking_setup snippet

mmoll:
@ares @sathieu could you please review/test?


mmoll:
I tested this and it seems to work fine for static and DHCP setups, VLAN and bonding are not supported, yet, but it's still an improvement compared to the situation before.

ares:
I'm fine with this as temporary solution but also note that a different approach was started few days ago (not sure when this will be finished) https://github.com/ares/foreman_net_templates
also this template won't work on some deb versions (or all of them?) since the installer overwrites interfaces file before reboot

last comment, it requires user to set interface identifier (which is not required in the form)

GregSutcliffe:
>  also this template won't work on some deb versions (or all of them?) since the installer overwrites interfaces file before reboot

I believe this bug is confined to Wheezy, although I haven't _actually_ tested it on Jessie yet.

ares:
I just verified this on both wheezy and jessie, works nicely on jessie so :+1: from me

snippets/preseed_networking_setup.erb
```diff
@@ -0,0 +1,49 @@
...
+<% unless dhcp -%>
+    address <%= @host.ip %>
+    gateway <%= @host.subnet.gateway  %>
+    netmask <%= @host.subnet.mask  %>
+    dns-nameservers <%= @host.subnet.dns_primary %> <%= @host.subnet.dns_secondary %>
+    dns-search <%= @host.domain %>
+<% end %>
+EOF
+
+<% @host.interfaces.each do |interface| %>
```
ares: this caused problems when safe_mode is enabled, also I think it should be `@host.managed_interfaces` to skip unmanaged interfaces

mmoll: changed to `@host.managed_interfaces`, what was/is the problem in safe_mode? I do have safe_mode on, but this works (with Foreman 1.10.1).

ares: I'm getting following error with safe mode on, changing it to managed_interfaces fixes it

There was an error rendering the Preseed default finish template: The snippet 'preseed_networking_setup' threw an error: undefined method 'each' for Safemode::Jail (ActiveRecord::Associations::CollectionProxy)

ares:
ACK

snippets/preseed_networking_setup.erb
domcleal: That might be worth filing as a bug against 1.11, since it's probably similar to http://projects.theforeman.org/issues/13168 and will probably affect other people.

ares: good catch, http://projects.theforeman.org/issues/13755

mmoll:
Thanks, develop and 1.11 would be ok for this, I presume.

ares:
yes, sounds good to me

ares:
Merged as 4e64d1ab0599aa16cc4fa098ed0670d046103bbd and CPed as 99e9e29ea51ba38b45c82edc8719fd71b7753751. Thanks all that participated.

PR #237, opened 2016-02-12
Fix issue #13684 RHN registration on RHEL 5

orpiske:
This is a proposed patch to fix issue #13684 RHN registration on RHEL 5. The patch modifies the code to use subscription-manager if the host is RHEL 5 or older.


mmoll:
@orpiske I don't have a RHEL 5 to test, but this is looking sane. However, please delete the added empty lines (except the last linefeed, which can stay).

mmoll:
Merged, obrigado @orpiske!

PR #239, opened 2016-02-17
refs #12791 - use boot_files_uri for SUSE iPXE

mmoll:
Also for 1.11, please (and as the original bug is targeted for 1.10.2 now, also 1.10, it's broken now with 1.10.1 anyway). See the original issue and PR for details.

@SkySymbol The referenced Foreman core patch should also apply to 1.10, so you could test these patches together, if interested.


domcleal:
I checked the Debian iPXE template renders while merging the core PR, so the rest should be fine too.

dLobatog:
@mmoll This is what you showed me on cfgmgmtcamp, no? :+1: nice!

ghost:
Thank you Michael, i'll test by the end of the week

domcleal:
Thanks @mmoll, merged to develop and 3677ded on 1.11-stable.

PR #242, opened 2016-02-29
refs #10926 - eject bootdisk when host is bootdisk provisioned

timogoebel:
This PR adds the ability to eject the bootdisk after installation when the host is provisioned via bootdisk.

This unfortunately does not work for rhel 7. There is already a bug opened. https://bugzilla.redhat.com/show_bug.cgi?id=1191386

See:
https://github.com/theforeman/foreman_bootdisk/pull/19

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -97,7 +97,11 @@ bootloader --location=mbr --append="<%= @host.params['bootloader-append'] || 'no
 <% end -%>
 
 text
+<% if @host.respond_to?(:bootdisk_build?) && @host.bootdisk_build? %>
+reboot --eject
+<% else -%>
 reboot
+<% end -%>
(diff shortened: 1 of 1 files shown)


mmoll:
would hurt to always reboot with `--eject` regardless of the hardware configuration?

domcleal:
I think always ejecting is reasonable.

timogoebel:
@domcleal : Do you know what happens on hardware? I'm picturing all servers in a datacenter having their cd drives open.

domcleal:
Good point, I was thinking of virtual only for some reason.  It will physically eject on most hardware IIRC - and I did once see the scenario you describe...

:+1: then.

mmoll:
merged, Danke @timogoebel!

PR #248, opened 2016-03-19
miscellaneous cleanups and unifications

mmoll:
- use operating systems names without versions in the header
- reduce diff between kickstart RHEL and non-RHEL
- reduce diff between autoyast SLES and non-SLES
- don't use '-e 0' to suppress errors with yum


sathieu:
looks good to me

kickstart/finish.erb
```diff
@@ -2,15 +2,8 @@
 kind: finish
 name: Kickstart default finish
 oses:
```
domcleal, line 4: missing "RedHat"?

mmoll, line 4: there was no RHEL of any version associated before and then the `redhat_register` snippet should also go into there, I presume?

kickstart/userdata.erb
```diff
@@ -2,15 +2,8 @@
 kind: user_data
 name: Kickstart default user data
 oses:
```
domcleal, line 4: missing RedHat?

mmoll, line 4: same comment as above

autoyast/provision_sles.erb
```diff
@@ -124,6 +124,7 @@ cp /etc/resolv.conf /mnt/etc
 <% if salt_enabled %>
 <%= snippet 'saltstack_setup' %>
 <% end -%>
+
 <% if spacewalk_enabled -%>
```
domcleal, line 128: provision_rhel has this higher so it's able to install the packages if using Spacewalk etc.

autoyast/provision.erb
```diff
@@ -113,6 +115,9 @@ cp /etc/resolv.conf /mnt/etc
         <notification>Setting up Puppet / Foreman ...</notification>
         <source><![CDATA[
 /bin/hostname <%= @host.name %>
+
+<%= snippet('remote_execution_ssh_keys') %>
+
 <% if puppet_enabled -%>
```
domcleal, line 121: missing chef_client?

domcleal:
The two {kickstart,preseed}/userdata.erb files could do with a slight refresh too, they're missing Chef, Salt and remote execution keys.

kickstart/finish.erb
domcleal, line 4: Hm, yes, it would be needed.  I don't mind if you want to leave it off.

autoyast/provision_sles.erb
mmoll, line 128: For SLES that's only to register the system, main reason being that we add puppet/salt repos and packages in the first install step in the XML anyway. I do get the point, but if this needs to be changed that would probably require refactoring which can only be done when having such a setup available for testing.

autoyast/provision.erb
mmoll, line 121: We don't support chef on SUSE at the moment (I also did no research if and how that would be possible)

mmoll:
tushed up the userdata files and added `('built')` for `foreman_url` in a few places.

domcleal:
Merged, thanks @mmoll!

PR #249, opened 2016-03-19
Adding ESX templates

LiamLeane:
ESXi provisioning templates for Foreman.

### DIFF
--- a/esx/PXELinux.erb
+++ b/esx/PXELinux.erb
@@ -1,0 +1,14 @@
+<%#
+kind: PXELinux
+name: ESXi Empty
+oses:
+- ESXi 5
+- ESXi 5.5
+- ESXi 6
+%>
+
+DEFAULT linux
+LABEL linux
+KERNEL undionly.kpxe
+APPEND dhcp && chain <%= foreman_url('iPXE') %>
+IPAPPEND 2
--- a/esx/PXELinux.erb
+++ b/esx/PXELinux.erb
@@ -1,6 +1,6 @@
 <%#
 kind: PXELinux
-name: ESXi Empty
+name: ESXi PXELinux
 oses:
 - ESXi 5
 - ESXi 5.5
--- a/esx/iPXE.erb
+++ b/esx/iPXE.erb
@@ -1,0 +1,13 @@
+#!ipxe
+<%#
+kind: iPXE
+name: ESXi iPXE
+oses:
+- ESXi 5
+- ESXi 5.5
+- ESXi 6
+%>
+ 
+kernel -n mboot.c32 <%= @host.operatingsystem.medium_uri(@host) %>mboot.c32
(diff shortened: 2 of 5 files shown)


mmoll:
@LiamLeane thanks for working on another point on my list (https://speakerdeck.com/mmoll/extending-operating-systems-support-of-foreman#24)! However, like (#198, now known as #250), this also needs implementation of ESXi as an operating system in Foreman core to really yield an unattended installation without much hassle. theforeman/foreman#3331 is the respective core PR for XenServer.

LiamLeane:
Willco, I had plans to add an ESX OS type anyway so there would be a pretty icon :) Ill get to work on adding the OS to Foreman.

Other then the annoying changes required to iPXE and changing boot.cfg which will be required irrespective of changes to Foreman this runs cleanly with RedHat set as a family. I have had a derivative of this running in our QA automation environment for a few months provisioning ~15 instances a day.

mmoll:
I just downloaded an ESXI ISO, the contents are quite baffling... Maybe it's possible to just drop in the ISO as-is and use [memdisk](http://www.syslinux.org/wiki/index.php?title=MEMDISK#ISO_images)?

mmoll:
[sanboot](http://ipxe.org/cmd/sanboot) also looks like something that could be used

LiamLeane:
Added OS type in theforeman/foreman#3351 and updated templates. Also improved readme.

esx/PXELinux.erb
```diff
@@ -0,0 +1,14 @@
...
+oses:
+- ESXi 5
+- ESXi 5.5
+- ESXi 6
+%>
+
+DEFAULT linux
+LABEL linux
+KERNEL undionly.kpxe
+APPEND dhcp && chain <%= foreman_url('iPXE') %>
```
dLobatog, line 13: At this point I get `Loading undionly.kpxe ... OK. Booting kernel failed: Invalid argument.`

This is what that chain resolves to:

```
#!ipxe
kernel -n mboot.c32 http://localhost:9000//mboot.c32
imgargs mboot.c32 -c http://localhost:9000// text ks=http://toshiba.lobatolan.home:3000/unattended/provision?token=31aae981-d3aa-473d-b394-41242613cd95; ignoreHeadless="True"
boot mboot.c32
```

Any ideas? From what I see in the installation media there's no attempt to even GET mboot.c32 so I guess that's not a problem

mleklund: FWIW, I use the following snippet to go straight from PXE to kickstart as I am not running iPXE

```
DEFAULT install
NOHALT 1
LABEL install
 KERNEL boot/esxi60u2/mboot.c32
 APPEND -c boot/esxi60u2/boot.cfg ks=<%= foreman_url("provision")%>
 IPAPPEND 2
```
At least for ESXI 6.0

mmoll:
The core PR was closed after 6 months inactivity, so closing this also.

PR #250, opened 2016-03-19
add XenServer templates

mmoll:
for theforeman/foreman#3331


xenserver/provision.erb
```diff
@@ -0,0 +1,32 @@
+<%#
+kind: provision
+name: XenServer default answerfile
+oses:
+- XenServer
+%>
+<% subnet = @host.subnet -%>
+<% dhcp = subnet.dhcp_boot_mode? -%>
+<% interface_name = @host.primary_interface.identifier.empty? ? 'eth0' : @host.primary_interface.identifier -%>
```
mmoll: @ares is `@host.primary_interface.identifier` correct? this didn't work for me, but could be related to my installation

ares: Since 1.8+ it should return interface object so the snippet would return identifier of primary interface. Before 1.8 `@host.primary_interface` would return identifier itself. So yes, looks good to me, I'd just prefer to use `blank?` instead of `empty?`. When creating host from UI it could be empty string, from API it can be nil.

mmoll: allright, that sounds like my problem, will change that and test, thanks!

mmoll: updated, fixed my problem

xenserver/PXELinux.erb
```diff
@@ -0,0 +1,12 @@
...
+kind: PXELinux
+name: XenServer default PXELinux
+oses:
+- XenServer
+%>
+
+default xenserver
+label xenserver
+kernel mboot.c32
+append <%= @xen %> dom0_max_vcpus=16 dom0_mem=752M com1=115200,8n1 console=com1,vga --- <%= @kernel -%> xencons=hvc console=hvc0 console=tty0 answerfile=<%= foreman_url('provision')%> install --- <%= @initrd %>
```
ekohl: Are `dom0_max_vcpus` and `dom0_mem` settings that should typically be modified? If so, should they be parameters?

xenserver/finish.erb
```diff
@@ -0,0 +1,9 @@
+<%#
+kind: finish
+name: XenServer default finish
+oses:
+- XenServer
+%>
+
+#!/bin/bash
+/usr/bin/wget --output-document=/dev/null <%= foreman_url %>
```
ekohl: Should this be `foreman_url('built')`?

mmoll: yes

xenserver/PXELinux.erb
mmoll: Nope, as this is only used for installation. But good point, I synced the settings from the ISO.

ekohl:
Can't judge the exact Xen details, but otherwise it looks good.

domcleal:
Renders perfectly with the XenServer PR now merged to core, :+1: when Travis passes.

mmoll:
thanks @pujan14 for starting this!

PR #251, opened 2016-03-21
Fixes #13000 - use inheriting_mac to support virtual interfaces

ares:
inheriting_mac was introduced in Foreman 1.9 so it should be safe to rely on it now


mmoll:
@sghai, could you test this and report back, please?

ares:
btw it's very easy to test, just try creating vlan and alias without specifying MAC and then review provisioning template, without patch you see something like `real=...| grep | ...`, with patch you'll see `real=...| grep $mac_of_parent |...`

mmoll:
merged, díky @ares!

PR #254, opened 2016-03-25
Add disable-firewall paramater

brandonweeks:
Adding a firewall to optionally disable the firewall on EL based systems. Internally we are using an iptables based firewall configuration system and would prefer to have firewalld disabled on provisioning.

cc: @mmoll

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -18,6 +18,7 @@ This template accepts the following parameters:
 - salt_master: string (default=undef)
 - ntp-server: string (default="0.fedora.pool.ntp.org")
 - bootloader-append: string (default="nofb quiet splash=quiet")
+- disable-firewall: boolean (default=false)
 %>
 <%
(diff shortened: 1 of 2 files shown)


mmoll:
I'd rather like a parameter like `disable-firewall` to not change the defaults when not having the parameter. :+1: otherwise, I'm having something similar in my private templates.

brandonweeks:
@mmoll done.

mmoll:
:+1:

mmoll:
oh, one thing, could this please also get added to the RHEL template?

mmoll:
and, it's param_e_ter in the commit message ;)

brandonweeks:
done and done.

ekohl:
I'm generally not a fan of negative named variables but in this case I think having an enable-firewall parameter which defaults to true may be confusing as well. The code looks good so :+1: either way.

mmoll:
merged, thanks @brandonweeks!

PR #258, opened 2016-04-03
Fixes #257 - Split the discovery stanza from the global PXE template into a snippet

### LINKED ISSUE #257: Split the discovery stanza from the global PXE template to a snippet
This is needed so we could reuse the template in https://github.com/theforeman/foreman_discovery/pull/265

PR will follow shortly

imriz:


mmoll:
looking good, will there be more modifications required later or is it only about having the snippet available?

imriz:
This is the only change for now

mmoll:
merged with minor changes, thanks @imriz!

PR #259, opened 2016-04-03
Allow for empty gateway or DNS servers in subnet in CoreOS cloudconfig

siwyd:
My CoreOS guest's DNS was not working after provisioning with CoreOS
community template. The problem was an empty second DNS in my
subnet. This caused the following networkd profile in the CoreOS guest:

```
[Match]
MACAddress=52:54:00:de:ad:e6
[Network]
Gateway=10.200.0.1
Address=10.200.0.22/24
DNS=10.200.0.20
DNS=
```

This is a faulty profile. /etc/resolv.conf does not get populated at all.
When you delete the empty DNS string, resolv.conf gets populated.
The second empty DNS string should not be present.

This commit makes sure that empty (optional) subnet parameters will not
be present at all. This includes the following:
- gateway
- primary DNS
- secondary DNS


mmoll:
@johscheuer could you have a look? :eyes:
@HyperBaton in the kickstart snippet, the checks are like `<% if !subnet.gateway.nil? && !subnet.gateway.empty? -%>` could you incooperate that nil check also?

siwyd:
@johscheuer Thanks for the reference to the kickstarter snippet. I've updated with a commit to do a nil check and to keep the syntax similar to that template.

johscheuer:
@mmoll looks good to me. 
@HyperBaton could you squash the commits to one?

siwyd:
@johscheuer Alright, done!

mmoll:
tested, working fine here, thanks @HyperBaton!

siwyd:
@mmol You're welcome! Thanks for merging.

PR #260, opened 2016-04-04
Add better support for FreeBSD template

ngoralski:
Changes
fixes :
- a bug with http_proxy that block freebsd-update to fetch patches.

Enhancement :
- add the freebsd-update process after installing the OS to avoid to install them after installation
- allow disk templating for UFS based freebsd system

### DIFF
--- a/freebsd/disklayout_FreeBSD_mfsBSD-UFS.erb
+++ b/freebsd/disklayout_FreeBSD_mfsBSD-UFS.erb
@@ -1,0 +1,11 @@
+<%#
+kind: ptable
+name: FreeBSD UFS Template
+oses:
+- FreeBSD
+%>
+
+4G freebsd-ufs /,
+2G freebsd-swap,
+2G freebsd-ufs /var,
+auto freebsd-ufs /usr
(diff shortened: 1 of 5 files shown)


mmoll:
@ngoralski I'd like to go with the first/second commit only for now, that what I meant with "generally usable version). Having pkg update and freebsd-update running is something that should be in the templates, but adding a local repository is probably something only very few people do.

I did some work to get in the standard FreeBSD installer with UFS in mmoll/community-templates@dc2d484 but was missing things like 1024k alignment, which is very important for SAN and SSD deployments. I added add a custom ufsinstall script into my personal mfsbsd image (mmoll/mfsbsd@1691333) and am using that in my setup. However, I think that's nothing I like to include for public consumption.

ngoralski:
The local repo can be a local mirror of an official repo or one with self - compiled package.
IMHO, if you go on deploying system with The Foreman / (Puppet|whatever) you need to have a local (repo|mirror) to avoid using your internet connection, saving Bandwith.

About the 1024K alignment you're right, as actually i'm working on my personal esxi infrastructure, so no SAN or SSD storage. We can may be adapt your UFS disk formatting script to use the disk template system, feel free to contact me.

mmoll:
closing this for now, I'd rather wait for the freebsd installer to support 1024k alignment...

PR #263, opened 2016-04-08
kickstart: add deltarpm to package list

mmoll:
this is accelerating the "yum update" run in the postinst


ekohl:
Personally I dislike deltarpm because if you have sufficient bandwidth then it's faster to just fetch the full RPM instead of rebuilding it. We also customized our kickstart to use %packages and add update repos because you can do that with non-RHEL. That gave a much bigger performance boost.

mmoll:
interesting, for me it's with it faster because I'm behind 16MBit/s but my machines are powerful, but I don't mind closing this...

mmoll:
Closing. Other people with limited bandwith can always also add it.

PR #264, opened 2016-04-08
Allow fleet and etcd services to be disabled

siwyd:
Allow disabling of the etcd and fleet service.

This adds one parameter:
- disable_etcd

If unset or anything but 'true', etcd and fleet will be started.

Updated README also.


mmoll:
@HyperBaton instead of `if !...`, `unless ...` is clearer, IMHO.
@johscheuer could you have a look?

siwyd:
@mmol Totally agree. Fixup commited and squashed.

johscheuer:
@mmoll @HyperBaton  LGTM but I would suggest to remove the etcd part completely if etcd is deactivated (just for correctness). So I think just putting around line 7-14 `+<% unless @host.param_true?('disable_etcd') -%>` should do the work.

siwyd:
@johscheuer Yeah I see what you mean... I'll just remove the disable_fleet param and embed everything in the disable_etcd param. Is that okay?

mmoll:
Maybe a parameter like "single-host-mode" or similar is more clear? But I have no idea :)

siwyd:
@mmoll I thought about that too when I first wrote this pull request, but there might be a use case where someone wants to run a single node with etcd, just to test out Kubernetes for instance with a single node. So in retrospect, the naming 'single-host-mode' might be confusing. I'd like to think people who install CoreOS read up a bit on the inner workings and will understand the consequences of disabling etcd.

For me, I just run a single CoreOS node for Docker development and considering I won't use etcd / fleet, it made sense to make an option to disable them alltogether.

johscheuer:
I would like to keep default CoreOS template as simple as possible to make it easy for new users to understand it. Maybe it makes more sense to add some examples into the foreman website see [here](https://github.com/theforeman/theforeman.org/issues/368#issuecomment-201328289) to show users how they can use this template for different use cases like for [Kubernetes deployment](https://github.com/johscheuer/theforeman-coreos-kubernetes).

mmoll:
what's the status here?

johscheuer:
Like I said LGTM I only would recommend to remove the etcd part completely if etcd is deactivated.

@HyperBaton one thing I noticed, could you add `... (default: false)` to the readme?

siwyd:
@mmoll @johscheuer Sorry for the delay, been otherwise occupied. I've updated the commit to have only one parameter, updated readme also. If you guys feel like it's not a good idea to tweak this snippet too much, I'm okay with that. Just wanted to have a way to disable it easily for myself and thought I'd share it.

mmoll:
@HyperBaton No worries, such contributions should just not get lost by neglect. It sounds to me that you're not the only with such a use-case and thus this should get in. :+1:

mmoll:
In my test-setup this leads to no unit being enabled with disable_etcd (that's ok) and ending up in an invalid cloud-config yaml (CoreOS 899.15.0):

``` yaml
write_files:
  - content: |
      #cloud-config

      coreos:
        units:
      users:
        - name: core
          passwd: $5$iXlS/6WL$Y9wEzsKbURxj1Xm72/g3uNxC1J/VBRIFygcX.s/QExB

    path: /home/core/cloud-config.yml
    permissions: '0600'
    owner: core:core
```

`Incorrrect type for "units" (want []struct)`

johscheuer:
@mmoll yeah.. the problem is that etcd is disabled the the docker socket isn't exposed that means there are no unit. We need to sourund [Line 15](https://github.com/theforeman/community-templates/pull/264/files#diff-af02d5b762f49276d3a182fe9c8b3ad6R17) by something like 

``` ruby
<% if @host.param_false?('disable_etcd') || @host.param_true?('expose_docker_socket') -%>
units:
<% end -%>
```

but this looks pretty ugly :( I would also suggest to rename `disable_etcd` to `enable_etcd` which is by default true. So changing the variable name and replace `<% unless @host.param_true?('disable_etcd') -%>` with `<% unless @host.param_flase?('enable_etcd') -%>` should do the job and would simplify (a little bit still a ugly snippet) the part above to:

``` ruby
<% if @host.param_true?('enable_etcd') || @host.param_true?('expose_docker_socket') -%>
units:
<% end -%>
```

@mmoll maybe you have a better idea?

mmoll:
Maybe counting the units like we count apt-repos in the preseed templates might also work.

johscheuer:
This part https://github.com/theforeman/community-templates/blob/develop/preseed/provision.erb#L83-L129 ?

johscheuer:
So we could do something like this:

``` ruby
<% units = 0 %>
<% if @host.param_true?('enable_etcd') -%>
<% repos += 1 -%>
<% end -%>
<% if @host.param_true?('expose_docker_socket') -%>
<% repos += 1 -%>
<% end -%>
#cloud-config
<%#
kind: snippet
name: coreos_cloudconfig
%>
      coreos:
<% if @host.param_true?('enable_etcd') -%>
        etcd2:
<% if @host.params['etcd_discovery_url'] -%>
          discovery: <%= @host.params['etcd_discovery_url'] %>
<% end -%>
          advertise-client-urls: http://<%= @host.ip %>:2379
          initial-advertise-peer-urls: http://<%= @host.ip %>:2380
          listen-client-urls: http://0.0.0.0:2379
          listen-peer-urls: http://0.0.0.0:2380
<% end -%>
<% if units > 0 -%>
        units:
<% end -%>
<% if @host.param_true?('enable_etcd') -%>
          - name: etcd2.service
            command: start
          - name: fleet.service
            command: start
<% end -%>
<% if @host.param_true?('expose_docker_socket') -%>
          - name: docker-tcp.socket
            command: start
            enable: yes
            content: |
              [Unit]
              Description=Docker Socket for the API

              [Socket]
              ListenStream=2375
              BindIPv6Only=both
              Service=docker.service

              [Install]
              WantedBy=sockets.target
          - name: enable-docker-tcp.service
            command: start
            content: |
              [Unit]
              Description=Enable the Docker Socket for the API

              [Service]
              Type=oneshot
              ExecStart=/usr/bin/systemctl enable docker-tcp.socket
<% end -%>
<% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
<% dhcp = @host.subnet.dhcp_boot_mode? && !@static -%>
<% else -%>
<% dhcp = !@static -%>
<% end -%>
<% unless dhcp -%>
          - name: systemd-networkd.service
            command: stop
          - name: static.network
            command: start
            content: |
              [Match]
              MACAddress=<%= @host.mac %>
              [Network]
<% if !@host.subnet.gateway.nil? && !@host.subnet.gateway.empty? -%>
              Gateway=<%= @host.subnet.gateway %>
<% end -%>
              Address=<%= @host.ip %>/<%= @host.subnet.cidr %>
<% if !@host.subnet.dns_primary.nil? && !@host.subnet.dns_primary.empty? -%>
              DNS=<%= @host.subnet.dns_primary %>
<% end -%>
<% if !@host.subnet.dns_secondary.nil? && !@host.subnet.dns_secondary.empty? -%>
              DNS=<%= @host.subnet.dns_secondary %>
<% end -%>
          - name: down-interfaces.service
            command: start
            content: |
              [Service]
              Type=oneshot
              ExecStart=/usr/bin/bash -c 'for i in $(/usr/bin/ls /sys/class/net/); do [ $i == "lo" ] || (/usr/bin/ip link set $i down; /usr/bin/ip addr flush dev $i); done'
          - name: systemd-networkd.service
            command: restart
<% end -%>
<% if @host.params['ssh_authorized_keys'] -%>
      ssh_authorized_keys:
  <% @host.params['ssh_authorized_keys'].split(',').map(&:strip).each do |ssh_key| -%>
      - "<%= ssh_key %>"
  <% end -%>
<% else -%>
      users:
        - name: core
          passwd: <%= root_pass %>
<% end -%>
```

johscheuer:
@mmoll @HyperBaton status?

mmoll:
@johscheuer I'm waiting for a fully working version of this PR.
@HyperBaton if wanted we can also take this up and finish it for you...

siwyd:
Sorry, I didn't have time during the week. I've updated the commit with your suggestions.
@mmoll I also had to add the dhcp parameter as a units counter, otherwise the units header is missing when having a static network and no other units. For this reason, I also had to move some logic up in the snippet, to expose the dhcp variable earlier. It's not 'nice' in my opinion, but I'm not sure how to go about it differently.

Be my guest to alter this yourself, I've added @mmoll and @johscheuer as collaborators on my forked repo. Not sure if that's the way to go about it :) Bit new to all this.

mmoll:
I updated the PR to also ommit the "coreos" entity without units, as that one would be also empty.

siwyd:
Ah jup, didn't catch that.

mmoll:
seems to work fine, bedankt @HyperBaton!

siwyd:
Cool! Sorry if it turned out more difficult than I had anticipated :) But I imagine the logic which was put in now might prove useful in future changes...

PR #265, opened 2016-04-20
Fix puppet4 service not being enabled on RHEL

alexjfisher:
Due to https://github.com/theforeman/community-templates/commit/8eca3cf8665d6193b5b34a27fc118469a8598c3a#diff-8048fb7fe5c29050e20a1a592d25a6c0R35
none of the puppet agent service enabling code was being run after
puppet is installed from the PC1 repository.  After discussing on IRC,
it may be that at least for some platforms the service is enabled by
default on installation.  For both EL6 and EL7 this wasn't the case.

This commit fixes this and has been tested on Oracle Linux 6 and 7.

### DIFF
--- a/snippets/puppet_setup.erb
+++ b/snippets/puppet_setup.erb
@@ -32,6 +32,14 @@ cat > <%= etc_path %>/puppet.conf << EOF
 <%= snippet 'puppet.conf' %>
 EOF
 
+<% if @host.operatingsystem.family == 'Redhat' -%>
+<% if @host.operatingsystem.major.to_i > 6 -%>
+puppet_unit=puppet
+/usr/bin/systemctl list-unit-files | grep -q puppetagent && puppet_unit=puppetagent
+/usr/bin/systemctl enable ${puppet_unit}
+<% end -%>
+/sbin/chkconfig --level 345 puppet on
+<% end -%>
 <% unless @host.param_true?('enable-puppetlabs-pc1-repo') -%>
 <% if @host.operatingsystem.family == 'Debian' -%>
 if [ -f "/etc/default/puppet" ]
@@ -41,12 +49,6 @@ fi
 /usr/bin/puppet agent --enable
 <% elsif @host.operatingsystem.family == 'Freebsd' -%>
 echo 'puppet_enable="YES"' >>/etc/rc.conf
-<% elsif @host.operatingsystem.name == 'Fedora' -%>
-puppet_unit=puppet
-/usr/bin/systemctl list-unit-files | grep -q puppetagent && puppet_unit=puppetagent
-/usr/bin/systemctl enable ${puppet_unit}
-<% elsif @host.operatingsystem.family == 'Redhat' -%>
-/sbin/chkconfig --level 345 puppet on
 <% elsif @host.operatingsystem.family == 'Suse' -%>
 if [ -f "/etc/sysconfig/puppet" ]
 then


snippets/puppet_setup.erb
```diff
@@ -32,7 +34,15 @@ cat > <%= etc_path %>/puppet.conf << EOF
 <%= snippet 'puppet.conf' %>
 EOF
 
-<% unless @host.param_true?('enable-puppetlabs-pc1-repo') -%>
+<% if @host.param_true?('enable-puppetlabs-pc1-repo') -%>
+<% if rhel_compatible -%>
```
ekohl: Given these calls are idempotent I think this is safe for Fedora and we can remove this distinction.

alexjfisher: You mean change to `if @host.operatingsystem.family == 'Redhat'` and rely on `os_major > 6` doing the right thing for Fedora?

I was wary about trying to be too clever and coming up with correct, but difficult to follow logic.  I also wasn't in the position to test other OSes and didn't want to make assumptions.
I agree it's a bit messy though :(

Maybe it'd be best to refactor after other people contribute the correct and tested logic for some of the other OSes?  Dunno.

ekohl: > You mean change to if @host.operatingsystem.family == 'Redhat' and rely on os_major > 6 doing the right thing for Fedora?

Yes. It will break on some old Fedora versions (<= 16 I think), but they aren't supported anymore so that's fine.

alexjfisher:
OK.  In general, whilst I don't normally like knowingly breaking support for older software, the earliest version of fedora that puppetlabs even provides packages for is 20.

mmoll:
@alexjfisher untested, but I think this can be just simplified to:

``` diff
diff --git a/snippets/puppet_setup.erb b/snippets/puppet_setup.erb
index 6c5fb2c..c7eedbf 100644
--- a/snippets/puppet_setup.erb
+++ b/snippets/puppet_setup.erb
@@ -32,6 +32,14 @@ cat > <%= etc_path %>/puppet.conf << EOF
 <%= snippet 'puppet.conf' %>
 EOF

+<% if @host.operatingsystem.family == 'Redhat' -%>
+<% if @host.operatingsystem.major.to_i > 6 -%>
+puppet_unit=puppet
+/usr/bin/systemctl list-unit-files | grep -q puppetagent && puppet_unit=puppetagent
+/usr/bin/systemctl enable ${puppet_unit}
+<% end -%>
+/sbin/chkconfig --level 345 puppet on
+<% end -%>
 <% unless @host.param_true?('enable-puppetlabs-pc1-repo') -%>
 <% if @host.operatingsystem.family == 'Debian' -%>
 if [ -f "/etc/default/puppet" ]
@@ -41,12 +49,6 @@ fi
 /usr/bin/puppet agent --enable
 <% elsif @host.operatingsystem.family == 'Freebsd' -%>
 echo 'puppet_enable="YES"' >>/etc/rc.conf
-<% elsif @host.operatingsystem.name == 'Fedora' -%>
-puppet_unit=puppet
-/usr/bin/systemctl list-unit-files | grep -q puppetagent && puppet_unit=puppetagent
-/usr/bin/systemctl enable ${puppet_unit}
-<% elsif @host.operatingsystem.family == 'Redhat' -%>
-/sbin/chkconfig --level 345 puppet on
 <% elsif @host.operatingsystem.family == 'Suse' -%>
 if [ -f "/etc/sysconfig/puppet" ]
 then
```

alexjfisher:
@mmoll Thanks, that looks much neater!  Tested fine too.

ekohl:
:+1:

mmoll:
merged, thanks @alexjfisher!

PR #266, opened 2016-04-20
Add enable-epel param

alexjfisher:
Defaults to true, but provides a way to disable the use of EPEL repos.

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -14,6 +14,7 @@ This template accepts the following parameters:
 - http-proxy: string (default="")
 - http-proxy-port: string (default="")
 - force-puppet: boolean (default=false)
+- enable-epel: boolean (default=true)
 - enable-puppetlabs-repo: boolean (default=false)
 - enable-puppetlabs-pc1-repo: boolean (default=false)
 - salt_master: string (default=undef)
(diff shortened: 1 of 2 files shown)


mmoll:
:+1: from me, but could any EL user weigh in here? I think the original reason to have that was that Puppet needed packages from EPEL, which will not be required with Puppet 4 AIO packages.

ekohl:
This wil still default to installing EPEL, right? If so :+1:

alexjfisher:
@mmoll Correct, when installing from Puppetlabs PC1, EPEL isn't needed.  During testing, it was proving a bit unreliable too.  More than once, the provisioning hung after failing to install `epel-release`.  A dodgy mirror I guess?  If my servers need EPEL, I use puppet and a [rhn_channel](https://github.com/strider/puppet-rhnreg_ks/blob/master/lib/puppet/type/rhn_channel.rb) resource.

@ekohl Yes, the default behaviour shouldn't change.  I'm pretty sure I remember testing it too! ;)
`unless @host.param_false?` is used [here](https://github.com/theforeman/community-templates/blob/27c79c85de595204e626ecaf604eef17b34bc06d/snippets/freeipa_register.erb#L48) for the parameters that default to true.

mmoll:
merged, thanks @alexjfisher!

PR #269, opened 2016-04-24
Hostgroup name shall be first in pxe menu label

amirfefer:
According to [this](http://projects.theforeman.org/issues/6977) issue, The hostgroup name is more important for the server then the provisioning template name. The hostgroup defines really what it will get and also the content view afterwards.

You can find the  menu labels sorting PR [here](https://github.com/theforeman/foreman/pull/3457)


mmoll:
merged, thanks @amirfefer!

PR #270, opened 2016-04-27
Fixes #14844: Prevent selection of Grub target

imphil:
Ubuntu 16.04 (xenial) requires a new setting to avoid showing a "select
your grub installation location" window during installation.


mmoll:
:+1:

ekohl:
Thanks!

PR #271, opened 2016-04-27
Fixes #14845: Use the first partition in preseed

imphil:
Instead of hardcoding /dev/sda, use the first disk that the debian
installer detects through the `list-devices disk` command. This can still
select the wrong device for the particular setup, but at least does not
install into SD card devices any more (we had that happening a couple of
times).

### DIFF
--- a/preseed/README.md
+++ b/preseed/README.md
@@ -14,7 +14,7 @@ Tested on:
 
 The templates use some Host Parameters to contol the flow of the template. These are:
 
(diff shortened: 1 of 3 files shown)


mmoll:
I need to test this first, but the change itself is looking sensible (and similar to what we do on FreeBSD).

mmoll:
seems to work just fine, danke @imphil!

PR #272, opened 2016-04-27
Consistently use bin_path in puppet templates

imphil:
Just a minor driveby fix, without changing the functionality.


mmoll:
merged, danke @imphil!

PR #273, opened 2016-04-28
Run puppet to apply configuration inside the installer

imphil:
Currently, puppet is only set up inside the installer, but does not run to apply the configuration. 

We have the requirement to have a fully working system after the first reboot, and therefore need to run puppet agent inside the installer. That works great, with the exception of some special cases which require a running service to apply a configuration (i.e. adding a printer requires a running cups daemon). 

This PR therefore does two things:
- If the variable `run-puppet-in-installer` is set (we set it as global variable), puppet actually runs inside the installer. If not, no behavior changes.
- The `$::is_installer` fact can be used to determine if the current puppet run is inside the installer environment or not.

To work around a bug in Puppet, we need to simulate the existence of systemd in the chroot environment of the installer. This is being worked on to be fixed in Puppet, but not yet part of the Puppet 3.x train (and it looks like it never will be). The workaround, however, is simple enough to be included here. See the comment inside the snippet for details.


snippets/puppet_setup.erb
```diff
@@ -58,4 +58,26 @@ fi
...
+# Puppet 3.x?)
+mkdir -p /run/systemd/system
+<% end -%>
+<% end -%>
+
+# export a custom facter fact called 'is_installer=true' to allow puppet 
+# scripts to detect installer environment
+export FACTER_is_installer=true
+
+# passing an non-existent tag like "no_such_tag" to the puppet agent only initializes the node
```
ekohl: Isn't it `a non-existent tag`?

snippets/puppet_setup.erb
```diff
@@ -58,4 +58,26 @@ fi
...
+#
+# Related puppet tickets:
+# https://tickets.puppetlabs.com/browse/PA-136
+# https://tickets.puppetlabs.com/browse/PUP-5577
+#
+# We work around that here until it gets fixed in Puppet (maybe never for 
+# Puppet 3.x?)
+mkdir -p /run/systemd/system
+<% end -%>
+<% end -%>
```
ekohl, line 90: Should this be at the end? As I read it now, it's always executed.

snippets/puppet_setup.erb
imphil: Yes, I'll update the comment.

snippets/puppet_setup.erb
imphil, line 90: It's executed if puppet-run-in-installer == true && ubuntu > 15, and before we actually run puppet. That's intended. Or are we talking past each other?

ekohl, line 90: I think I've been unclear: as I read it now, puppet agent is always executed; even if run-puppet-in-installer is false. That's why I thought this should move to the end of the file.

snippets/puppet_setup.erb
```diff
@@ -58,4 +58,26 @@ fi
 <% end -%>
 <% end -%>
 
-<%= bin_path %>/puppet agent --config <%= etc_path %>/puppet.conf --onetime --tags no_such_tag <%= @host.puppetmaster.blank? ? '' : "--server #{@host.puppetmaster}" %> --no-daemonize
+<% if @host.param_true?('run-puppet-in-installer') -%>
+<% if @host.operatingsystem.name == 'Ubuntu' and @host.operatingsystem.major.to_i >= 15 -%>
```
sathieu: Can you add Debian >=8 too?

snippets/puppet_setup.erb
imphil, line 90: The actual puppet run is prevented by the tag, it's inserted below. 

`<%= @host.param_true?('run-puppet-in-installer') ? '' : '--tags no_such_tag' %>`

It's essentially a conditional version of the way it was before. If you prefer, I can also duplicate the puppet agent call and move it inside the conditional in line 61.

ekohl, line 90: Oh you're right. It appears I still need to wake up ;)

snippets/puppet_setup.erb
imphil: Sure. It probably could be extended for other OSes as well, but I didn't test those and therefore didn't want to include a workaround based on guesswork.

imphil:
I've pushed a new revision with your comments addressed.

ekohl:
:+1:

mmoll:
I'm sceptical if something like this should be in the upstream template repo here. Having the Puppet agent _not_ apply the config was a decision that was made already in the early days because it's known that this is often resulting into problems in the installer environment.

I'm not 100% unwilling to merge this, though, if more people chime in here, but then I'd also like to see that tested at least on CentOS/Fedora, preferably also on Suse and FreeBSD.

imphil:
We've been running this setup for quite some time now, and while we initially expected much larger problems, it works out fine in almost all cases. As I noted, the only problem we had was adding a printer to the system which requires a running cups daemon. This same scenario would come up with for example loading a database fixtures, etc. But in most of those cases, it's no problem if the puppet run partially fails in the installer and then picks up again after the first reboot.

We did this change because we're installing desktop machines, and the users tend to try to log in as soon as the machine reboots for the first time. At that time, the installation is not yet completed and the machines appear to the user as "broken" and we get a lot of support calls. Also, reliably rebooting after the initial installation finished turned out to be more tricky than we expected. Running puppet inside the installer avoids all that.

But taking all the uncertainties into account, I left the default in this patch as it was before and introduced the "run puppet in installer" feature only behind an option, so users need to opt-in. I think that's a sensible option and I'd love to see this included in the default templates to save others the time to figure it out again (it took a colleague of mine quite some time initially to figure out that adding the tag was preventing the puppet run, for example).

Unfortunately I'm not able to test on other OSes, as our setup currently only has Ubuntu running. I'd be happy if any of you running such a setup could give it a try. Probably the puppet systemd fix needs to be replicated for other distributions at least.

ekohl:
We could add some warning in the header that running puppet inside the installer can bring some ~~pain~~ challenges and you should be aware of it. I think that provides a good enough balance to provide this "advanced" feature.

mmoll:
yeah, something like "don't ask us when it breaks" should work...

imphil:
Where would you add the warning? As a comment inside the file? Or can we put it somewhere that the user actually sees it when deploying a node with this setting?

ekohl:
I think inside the template should be sufficient. That's the only place to find this variable.

imphil:
I've added a comment inside the template that this configuration is unsupported.

mmoll:
tested and merged with some rewordings, danke @imphil!

PR #277, opened 2016-05-19
Fixes #15046 - Add bmc_nic_setup snippet

bagasse:
Template part of http://projects.theforeman.org/issues/15046.

### DIFF
--- a/snippets/bmc_nic_setup.erb
+++ b/snippets/bmc_nic_setup.erb
@@ -1,0 +1,89 @@
+<%#
+kind: snippet
+name: bmc_nic_setup
+%>
+<%#
+# bmc_nic_setup snippet
+# This snippet configure BMC interfaces.
+# At this time, only IPMI provider is supported.
+# It was tested on some Dell hardware with CentOS 6 - CentOS 7 and should be compatible with respective
+# upstream RHEL versions.
+#
+# What this snippet does:
+# * Configure the networking part of the interface (ipaddr, netmask, ip gateway or dhcp)
+#
+# * Set the authentication mecanism to 'MD5' for the used privilege level
+#
+# * Create or Update the provided user with provided privileges level. ADMINISTRATOR privilege level is used by default
+#   because ipmitool utility request ADMINISTRATOR privileges level if no privilege level is provided.
+#   If you want to use a lower level of privilges (like OPERATOR), don't forget to add the privilege level to request
+#   with your ipmi command (eg: ipmitool -I lanplus -L operator -U user1 -H 192.168.1.2 chassis status).
+#
+# * Enable the user created/updated
+#
+# What this snippet doesn't do:
+# * Set/Update the VLAN tag on the BMC interface because this configuration depends on the switch port configuration
+#   and the 'ipmitool lan set <channel> vlan id <off|<id>>' command don't work on all Dell hardware tested.
+#
+# host_params:
+# * bmc_ipmi_priv_level: Privilege level to use for the ipmi user.
+#                        'ADMINISTRATOR' is used by default if not provided.
+# * bmc_ipmi_auth_method: Channel authentication types. Possible values: NONE, MD2, MD4, MD5, PASSWORD.
+#                         'MD5' is used by default if not provided.
+# * bmc_ipmi_channel_id: Channel ID for the IPMI configuration.
+#
+#
+# How to test configuration result:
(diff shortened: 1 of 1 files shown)


mmoll:
This is looking very useful, but I don't have any hardware to test this, so only two comments from my side:
- Please make this snippet more OS agnostic. Have a look onto the puppet_setup snippet as an example. I'd like to see that working also on Debian/Ubuntu, idealy also on FreeBSD. Especially regarding the later (no GNU grep or awk) I'd recommend e.g. to use cut instead of awk and `tail -n 1` instead of `tail -1`
- I _think_ there's also hardware on the market that doesn't use 1 as its default channel (@lzap I think you came across some IBM h/w where that was the case), so it might make sense to extract that into a parameter.

bagasse:
@mmoll, thank you for your comment. I've only dell hardware so i didn't test it on other hardware. I've modified the snippet to:
- Manage channel IPMI channel ID and use 1 by default
- Use package manager from os type family (maybe i have to manage dnf for recent fedora releases ?).
- Use portable options of grep, and replaced awk with cut

Modifications tested on CentOS7 at this time. No hardware test server free at this time, it will be difficult to me to test it on some debian or freebsd versions.

lzap:
I also don't have IPMI enabled hardware myself, but since this is a snippet, I think we can merge it.

snippets/bmc_nic_setup.erb
```diff
@@ -0,0 +1,97 @@
+<%#
+kind: snippet
+name: bmc_nic_setup
+oses:
```
mmoll: snippets don't have OS associations, so this should get deleted

snippets/bmc_nic_setup.erb
```diff
@@ -0,0 +1,97 @@
...
+- RedHat 7
+%>
+<%#
+# bmc_nic_setup snippet
+# This snippet configure BMC interfaces.
+# At this time, only IPMI provider is supported.
+# It was tested on some Dell hardware with CentOS 6 - CentOS 7 and should be compatible with respective
+# upstream RHEL versions.
+#
+# What this snippet do:
```
mmoll: does

snippets/bmc_nic_setup.erb
```diff
@@ -0,0 +1,97 @@
...
+# * Set the authentication mecanism to 'MD5' for the used privilege level
+#
+# * Create or Update the provided user with provided privileges level. ADMINISTRATOR privilege level is used by default
+#   because ipmitool utility request ADMINISTRATOR privileges level if no privilege level is provided.
+#   If you want to use a lower level of privilges (like OPERATOR), don't forget to add the privilege level to request
+#   with your ipmi command (eg: ipmitool -I lanplus -L operator -U user1 -H 192.168.1.2 chassis status).
+#
+# * Enable the user created/updated
+#
+# What this snippet don't do:
```
mmoll: doesn't

snippets/bmc_nic_setup.erb
```diff
@@ -0,0 +1,97 @@
...
+os_name   = @host.operatingsystem.name
+
+%>
+
+<% if bmc_nic && bmc_nic.managed? && bmc_nic.provider == "IPMI" && !bmc_nic.username.empty?  && !bmc_nic.password.empty? -%>
+
+# Configure BMC interface
+<% case os_family %>
+<% when 'Redhat' then %>
+yum  -t -y install ipmitool
```
mmoll: one space too much here

snippets/bmc_nic_setup.erb
```diff
@@ -0,0 +1,97 @@
...
+
+privilege_levels = {'ADMINISTRATOR' => 4, 'OPERATOR' => 3}
+user_privilege_level = @host.params['bmc_ipmi_priv_level'] || 'ADMINISTRATOR'
+auth_method = @host.params['bmc_ipmi_auth_method'] || 'MD5'
+channel_id = @host.params['bmc_ipmi_channel_id'] || '1'
+
+os_family = @host.operatingsystem.family
+os_major  = @host.operatingsystem.major.to_i
+os_name   = @host.operatingsystem.name
+
```
mmoll: major and name are not used, AFAICS?

mmoll:
There's no dnf usage in any of the templates, yet, so that's OK

bagasse:
Done.

mmoll:
merged, merci @bagasse!

bagasse:
@mmoll, @lzap, thank you for comments/reviews

PR #278, opened 2016-06-06
Fixes #15046 - Case statement in bmc_nic_setup erb template.

bagasse:
Hi,
Sorry for the noise, but the case statement in erb don't work in the first PR...


mmoll:
Maybe it would be nice, in terms of unification, to use if-statements like in snippets/puppet_setup.erb?

mmoll:
merged, merci @bagasse!

PR #280, opened 2016-06-08
use .present?

mmoll:


ares:
looks good to me, I think there's a rule that we keep templates compatible for 2 releases so once 1.12 is out, I think this can be merged

domcleal:
> I think there's a rule that we keep templates compatible for 2 releases

Not since the repo branch layout's [changed](https://groups.google.com/d/msg/foreman-dev/7do3i4ONFnY/J3_iTJRq56gJ) - now develop only needs to be compatible with Foreman's develop.

domcleal:
Merged, thanks @mmoll - also cherry picked with 8f1c3b1c to 1.12-stable.

ares:
right, thanks

PR #281, opened 2016-06-09
Fixes #15358 - UserData default is insecure

edestecd:
Make the default the secure option.  Also provide a cautionary warning comment about enabling ssh_pwauth.


domcleal:
I agree with the sentiment, but I'm concerned about the default behaviour. I think it's perhaps preferable to configure password auth if the user hasn't set the SSH auth keys parameter as Foreman provides the root password.  If they've provided SSH keys and ssh_pwauth isn't true, then skip the user/password setup?

(Or is that logic too convoluted?)

domcleal:
It'd also be really useful if you could add info about the parameters and their behaviour to the header of the file, mirroring what's happened in the provision templates. It'll make it easier to use as they expand.

edestecd:
I could probably be down with the conditions you describe for the password, so long as I can turn it off in my environment with a param.

domcleal:
Oh definitely, I think switching off password auth if _either_ SSH auth keys are given, or the ssh_pwauth param's set to false would make sense.  It'd automatically be secure when SSH keys are given or when set, but still be usable if somebody only set a password.

edestecd:
@domcleal next revision

domcleal:
:+1: when rebased for me.

edestecd:
rebased

domcleal:
Merged as 002720b, thanks @edestecd.

PR #282, opened 2016-06-09
Fixes #15359 - UserData default should install an automation tool

edestecd:
This also updates the epel snippet and introduces a new snippet to set up the puppetlabs repo, which was lifted form the Kickstart finish.  
The puppetlabs repo only supports fadora and centos.  It needs to also support debain/ubuntu.
Any other templates have these shell commands for debian?

### DIFF
--- a/cloudinit/userdata_cloudinit.erb
+++ b/cloudinit/userdata_cloudinit.erb
@@ -16,6 +16,12 @@ This template accepts the following parameters:
 -%>
 <%
   ssh_pwauth = @host.params['ssh_pwauth'] ? @host.param_true?('ssh_pwauth') : !@host.params['ssh_authorized_keys']
+  rhel_compatible = @host.operatingsystem.family == 'Redhat' && @host.operatingsystem.name != 'Fedora'
+  # safemode renderer does not support unary negation
+  pm_set = @host.puppetmaster.empty? ? false : true
+  puppet_enabled = pm_set || @host.param_true?('force-puppet')
+  salt_enabled = @host.params['salt_master'] ? true : false
+  chef_enabled = @host.respond_to?(:chef_proxy) && @host.chef_proxy
 -%>
 #cloud-config
 hostname: <%= @host.shortname %>
@@ -46,6 +52,28 @@ ssh_authorized_keys:
 
 <% if @host.param_true?('package_upgrade') -%>
 package_upgrade: true
+<% end -%>
+
+runcmd:
+<% if rhel_compatible -%>
+- |
+<%= indent(2) { snippet('epel') } %>
+<% end -%>
+- |
+<%= indent(2) { snippet('remote_execution_ssh_keys') } %>
+<% if chef_enabled -%>
+- |
+<%= indent(2) { snippet('chef_client') } %>
+<% end -%>
+<% if puppet_enabled -%>
+- |
+<%= indent(2) { snippet('puppetlabs_repo') } %>
+- |
+<%= indent(2) { snippet('puppet_setup') } %>
+<% end -%>
+<% if salt_enabled -%>
+- |
+<%= indent(2) { snippet('saltstack_setup') } %>
 <% end -%>
 
 <%# Contact Foreman to confirm instance is built -%>
--- a/snippets/epel.erb
+++ b/snippets/epel.erb
@@ -1,6 +1,10 @@
 <%#
 kind: snippet
 name: epel
-%>
-
-su -c 'rpm -Uvh <%= @host.os.medium_uri(@host, "http://dl.fedoraproject.org/pub/epel/epel-release-latest-#{@host.operatingsystem.major}.noarch.rpm") %>'
+-%>
+<%
+http_proxy = @host.params['http-proxy'] ? " --httpproxy #{@host.params['http-proxy']}" : nil
+http_port  = @host.params['http-proxy-port'] ? " --httpport #{@host.params['http-proxy-port']}" : nil
+os_major   = @host.operatingsystem.major.to_i
+-%>
+rpm -Uvh<%= http_proxy %><%= http_port %> https://dl.fedoraproject.org/pub/epel/epel-release-latest-<%= os_major %>.noarch.rpm
--- a/snippets/puppetlabs_repo.erb
+++ b/snippets/puppetlabs_repo.erb
@@ -1,0 +1,39 @@
+<%#
+kind: snippet
+name: puppetlabs_repo
+-%>
+<%
+http_proxy   = @host.params['http-proxy'] ? " --httpproxy #{@host.params['http-proxy']}" : nil
+http_port    = @host.params['http-proxy-port'] ? " --httpport #{@host.params['http-proxy-port']}" : nil
+proxy_uri    = @host.params['http-proxy'] ? "http://#{@host.params['http-proxy']}:#{@host.params['http-proxy-port']}" : nil
+proxy_string = proxy_uri ? " -e https_proxy=#{proxy_uri}/" : ''
+os_family = @host.operatingsystem.family
+os_major  = @host.operatingsystem.major.to_i
+os_name   = @host.operatingsystem.name
+
+if os_family == 'Redhat'
+  repo_host = 'yum.puppetlabs.com'
+  if os_name == 'Fedora'
+    repo_os = 'fedora'
+  else
+    repo_os = 'el'
+  end
+elsif os_family == 'Debian'
+  repo_host = 'apt.puppetlabs.com'
+  repo_os = @host.operatingsystem.release_name
+end
+
+if @host.param_true?('enable-puppetlabs-repo')
+  repo_name = 'puppetlabs-release'
+elsif @host.param_true?('enable-puppetlabs-pc1-repo')
+  repo_name = 'puppetlabs-release-pc1'
+end
+-%>
+<% if repo_name -%>
+<% if os_family == 'Redhat' -%>
+rpm -Uvh<%= http_proxy %><%= http_port %> https://<%= repo_host %>/<%= repo_name %>-<%= repo_os %>-<%= os_major %>.noarch.rpm
+<% elsif os_family == 'Debian' -%>
+wget -O /tmp/<%= repo_name %>-<%= repo_os %>.deb<%= proxy_string %> https://<%= repo_host %>/<%= repo_name %>-<%= repo_os %>.deb
+dpkg -i /tmp/<%= repo_name %>-<%= repo_os %>.deb
+<% end -%>
+<% end -%>
--- a/kickstart/finish.erb
+++ b/kickstart/finish.erb
@@ -28,23 +28,6 @@ echo 'root:<%= root_pass -%>' | /usr/sbin/chpasswd -e
 <%= snippet 'epel' -%>
 <% end -%>
 
-<% if puppet_enabled -%>
-<% if @host.param_true?('enable-puppetlabs-repo') -%>
-<% if @host.operatingsystem.name == 'Fedora' -%>
-rpm -ivh http://yum.puppetlabs.com/puppetlabs-release-fedora-<%= os_major -%>.noarch.rpm
-<% else -%>
-rpm -ivh http://yum.puppetlabs.com/puppetlabs-release-el-<%= os_major -%>.noarch.rpm
-<% end -%>
-<% end -%>
-<% if @host.param_true?('enable-puppetlabs-pc1-repo') -%>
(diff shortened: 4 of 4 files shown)


snippets/epel.erb
```diff
@@ -1,6 +1,10 @@
...
 kind: snippet
 name: epel
-%>
-
-su -c 'rpm -Uvh <%= @host.os.medium_uri(@host, "http://dl.fedoraproject.org/pub/epel/epel-release-latest-#{@host.operatingsystem.major}.noarch.rpm") %>'
+-%>
+<%
+  http_proxy = @host.params['http-proxy'] ? " --httpproxy #{@host.params['http-proxy']}" : nil
+  http_port = @host.params['http-proxy-port'] ? " --httpport #{@host.params['http-proxy-port']}" : nil
+  os_major = @host.operatingsystem.major.to_i
```
domcleal: no need to add the .to_i here

snippets/puppetlabs_repo.erb
```diff
@@ -0,0 +1,23 @@
+<%#
+kind: snippet
+name: puppetlabs_repo
+-%>
+<%
+  http_proxy = @host.params['http-proxy'] ? " --httpproxy #{@host.params['http-proxy']}" : nil
+  http_port = @host.params['http-proxy-port'] ? " --httpport #{@host.params['http-proxy-port']}" : nil
+  os_major = @host.operatingsystem.major.to_i
```
domcleal: ditto

snippets/puppetlabs_repo.erb
```diff
@@ -0,0 +1,23 @@
...
+-%>
+<% if @host.param_true?('enable-puppetlabs-repo') -%>
+<% if @host.operatingsystem.name == 'Fedora' -%>
+rpm -Uvh<%= http_proxy %><%= http_port %> <%= @host.os.medium_uri(@host, "http://yum.puppetlabs.com/puppetlabs-release-fedora-#{os_major}.noarch.rpm") %>
+<% else -%>
```
domcleal: We'll need support for Debian/Ubuntu in here too I think, or at the _very_ least a check that doesn't try to run `rpm -Uvh` on a non-EL box.

snippets/epel.erb
```diff
@@ -1,6 +1,10 @@
...
+-%>
+rpm -Uvh<%= http_proxy %><%= http_port %> <%= @host.os.medium_uri(@host, "http://dl.fedoraproject.org/pub/epel/epel-release-latest-#{os_major}.noarch.rpm") %>
```
domcleal: It might be worth changing the URLs to HTTPS while you're here, they all support it now and it's important when downloading a release key.

snippets/puppetlabs_repo.erb
```diff
@@ -0,0 +1,23 @@
...
+name: puppetlabs_repo
+-%>
+<%
+  http_proxy = @host.params['http-proxy'] ? " --httpproxy #{@host.params['http-proxy']}" : nil
+  http_port = @host.params['http-proxy-port'] ? " --httpport #{@host.params['http-proxy-port']}" : nil
+  os_major = @host.operatingsystem.major.to_i
+-%>
+<% if @host.param_true?('enable-puppetlabs-repo') -%>
+<% if @host.operatingsystem.name == 'Fedora' -%>
+rpm -Uvh<%= http_proxy %><%= http_port %> <%= @host.os.medium_uri(@host, "http://yum.puppetlabs.com/puppetlabs-release-fedora-#{os_major}.noarch.rpm") %>
```
domcleal: ditto, prefer HTTPS here

snippets/epel.erb
edestecd: I just copied it from another template.  Perhaps we need to fix it there too. (I like consistency)
https://github.com/theforeman/community-templates/blob/develop/kickstart/provision.erb#L27

snippets/epel.erb
edestecd: I agree

snippets/puppetlabs_repo.erb
edestecd: Yup, I mentioned that in the PR description.  I will try and do this soon.

snippets/epel.erb
domcleal: Ah, probably because it's used for a comparison in that template.  Consistency's good, please keep `to_i` here then!

edestecd:
@domcleal Any thoughts on this change?

cloudinit/userdata_cloudinit.erb
```diff
@@ -41,6 +49,28 @@ users:
    lock-passwd: false
    passwd: <%= @host.root_pass %>
 
+ runcmd:
+ <% if rhel_compatible -%>
+ - <%= snippet('epel') -%>
```
domcleal: missing indent?  (And should it be in an indented block _below_ the `-` line like the others?)

cloudinit/userdata_cloudinit.erb
```diff
@@ -41,6 +49,28 @@ users:
...
+ <% end -%>
+ <% if puppet_enabled -%>
+ - <%= snippet('puppetlabs_repo') -%>
```
domcleal: missing indent?

snippets/puppetlabs_repo.erb
```diff
@@ -0,0 +1,39 @@
...
+
+if @host.param_true?('enable-puppetlabs-repo')
+  repo_name = 'puppetlabs-release'
+elsif @host.param_true?('enable-puppetlabs-pc1-repo')
+  repo_name = 'puppetlabs-release-pc1'
+end
+-%>
+<% if repo_name -%>
+<% if os_family == 'Redhat' -%>
+rpm -Uvh<%= http_proxy %><%= http_port %> <%= @host.os.medium_uri(@host, "https://#{repo_host}/#{repo_name}-#{repo_os}-#{os_major}.noarch.rpm") %>
```
domcleal: No need for the `<%= @host.os.medium_uri(@host, "` stuff in these snippets unless it uses the $version type variables that are used in installation media. I think it's a leftover from the EPEL snippet where we used to use the major version variable, but it was removed.  The URL can just be put straight into the template without the ERB `<%= ... %>`.

snippets/puppetlabs_repo.erb
```diff
@@ -0,0 +1,39 @@
...
+<% elsif os_family == 'Debian' -%>
+curl -L -o <%= "/tmp/#{repo_name}-#{repo_os}.deb" %><%= proxy_string %> <%= @host.os.medium_uri(@host, "https://#{repo_host}/#{repo_name}-#{repo_os}.deb") %>
```
domcleal: ditto, re medium_uri

domcleal: Also:
a) I don't know if curl is available on Debian families in a minimal installation. Our preseed templates prefer `wget`.
b) `<%= "/tmp/#{repo_name}-#{repo_os}.deb" %>` can be written into the template as `/tmp/<%= repo_name %>-<%= repo_os %>.deb` to reduce the interpolation.

snippets/epel.erb
```diff
@@ -1,6 +1,10 @@
...
-%>
-
-su -c 'rpm -Uvh <%= @host.os.medium_uri(@host, "http://dl.fedoraproject.org/pub/epel/epel-release-latest-#{@host.operatingsystem.major}.noarch.rpm") %>'
+-%>
+<%
+http_proxy = @host.params['http-proxy'] ? " --httpproxy #{@host.params['http-proxy']}" : nil
+http_port  = @host.params['http-proxy-port'] ? " --httpport #{@host.params['http-proxy-port']}" : nil
+os_major   = @host.operatingsystem.major.to_i
+-%>
+rpm -Uvh<%= http_proxy %><%= http_port %> <%= @host.os.medium_uri(@host, "https://dl.fedoraproject.org/pub/epel/epel-release-latest-#{os_major}.noarch.rpm") %>
```
domcleal: ditto, you can remove medium_uri from here too if you wish.

cloudinit/userdata_cloudinit.erb
```diff
@@ -41,6 +49,28 @@ users:
...
+ <% end -%>
+ - |
+ <%= indent(2) { snippet('remote_execution_ssh_keys') } %>
+ <% if chef_enabled %>
+ - |
+ <%= indent(2) { snippet('chef_client') } %>
+ <% end -%>
+ <% if puppet_enabled %>
```
domcleal: Just combine the two puppet_enabled sections, configure the repo and puppet_setup in the same bit.

domcleal:
Thanks for the update, it's looking good. Could you also rebase it please, the PR isn't mergeable as-is.

If you like, you could also replace the Puppet Labs repo installation in kickstart/finish.erb with a call to your new snippet.

cloudinit/userdata_cloudinit.erb
edestecd: Same as above.  The puppetlabs_repo template only ever returns one line...

cloudinit/userdata_cloudinit.erb
edestecd: http://yaml.org/
I prefer no indentation on arrays like in `Projects:`
However, if you prefer, I can switch all arrays in this template to 2 spaces like in `Related Projects:`

An indent block is not necessary, since the epel template only returns one line.  I'm chomping the newline by using `-%>`

cloudinit/userdata_cloudinit.erb
domcleal: The Debian part of the template can return two lines.

cloudinit/userdata_cloudinit.erb
domcleal: Sorry, it wasn't really the indenting of the array I meant, it was more that I expected the template to render more than one line. I think that's fine then, provided nobody edits it!

cloudinit/userdata_cloudinit.erb
edestecd: Good catch, I had not re-evaluated it since adding Debian.

snippets/puppetlabs_repo.erb
edestecd: OK, I'll remove the `<%= @host.os.medium_uri(@host, "`

cloudinit/userdata_cloudinit.erb
edestecd: XD, I know everyone has an opinion on indentation etc.  It will probably go back and forth.
I wish there was a good linter for yaml to enforce a standard indentation like the `puppet-lint-strict_indent-check` gem and put it to rest.

I will probably just put this one in an indent block to be consistent.  There is always the possibility that someone adds another line to the epel template later.  That would make it more robust.

cloudinit/userdata_cloudinit.erb
edestecd: will do

snippets/puppetlabs_repo.erb
edestecd: Let me check on curl.  I have a minimal Ubuntu 14 and Debian 8 box around.  You are correct they do not come with curl...  And RedHat does not ship with wget.  Argh.

The reason I used curl is because wget seems to be broken when using a proxy and an https url.  I have had many issues with this in the past even on RHEL 7. I can't seem to find any doc on the issue at the moment, but remember reading that it still was not fixed and no one was working on it...  I pretty much hit a wall with wget, so stopped using it.  Maybe its fixed by now?

edestecd: Is interpolation more expensive than erubis template rendering?

domcleal: > Is interpolation more expensive than erubis template rendering?

No idea, and performance doesn't matter - I was just trying to avoid using both in the hope it'd be clearer.

edestecd:
Thanks for the feedback.  I'll rebase and start making the suggested changes.

edestecd:
@domcleal I think that wraps it up

domcleal:
Merged, thanks @edestecd!

edestecd:
Do you have an automated way of syncing these to the foreman project periodically for db seed?

domcleal:
Not automated - it has script/sync_templates.sh in the repo, I'll run it soon.

mmoll:
@edestecd you can always use the foreman_templates plugin to sync from git repos (like this one here) into your Foreman instance (Agreed, that's not db:seed).

edestecd:
Yes, I use the templates plugin for existing installs of foreman.  I just wanted to make sure you didn't need PRs on the foreman project, so that fresh installs get these changes.  Sounds like you have that covered with the script ;)

PR #283, opened 2016-06-09
Fixes #15360 - UserData default should Patch the system and reboot

edestecd:


cloudinit/userdata_cloudinit.erb
```diff
@@ -41,8 +41,19 @@ users:
...
 <%# Contact Foreman to confirm instance is built -%>
 phone_home:
- url: <%= foreman_url('built') %>
- post: []
- tries: 10
+  url: <%= foreman_url('built') %>
+  post: []
+  tries: 10
+
+<% unless @host.param_false?('reboot') -%>
```
domcleal: I'm worried about this behaviour change, and wonder if the default should be not to upgrade/reboot and have it opt-in?

domcleal:
Also documentation for these parameters would be great.

cloudinit/userdata_cloudinit.erb
edestecd: OK, I agree with that.  Seems like more will not want this in a cloud environment.

edestecd:
Where would you like documentation?  At the top of the template, with a line for each param, like others?

domcleal:
Yep, top of the template please.

edestecd:
hmm other templates use dasherized params, while this one uses underscore.
Should we keep it underscore or try to make it consistent by switching the two existing params to dash?

domcleal:
It's probably better to prefer dashes if that's what the majority use already.

edestecd:
Other templates are using the param 'ssh_authorized_keys' and I have seen others with underscore like 'salt_master'.

Its a mixed bag at this point.  I'm not willing to change them all.   Seems like that could break for many that are already using them.

edestecd:
@domcleal  I think I fixed everything

domcleal:
Thanks @edestecd!

PR #287, opened 2016-07-07
Create SSH keys before joining IPA realm

lfisher47:
By default when joining a server to an ipa realm, the ssh public key gets uploaded, but since we are joining the realm before SSH has started and had a chance to create the host keys, running sshd-keygen before the ipa client installation puts keys into IPA and IPA can do the host key management.


snippets/freeipa_register.erb
```diff
@@ -24,6 +24,9 @@ name: freeipa_register
 #
 #   freeipa_opts                Additional options to pass directly to installer
 #
+
+/usr/sbin/sshd-keygen
```
domcleal: This only appears to exist on EL7 and higher, not EL6 or Debian/Ubuntu, which this snippet probably supports. Can you add equivalents?

GregSutcliffe: It's certianly available on Jessie, but you'll need to figure out the path:

```
greg@topaz:~$ which ssh-keygen
/usr/bin/ssh-keygen
```

domcleal: ssh-keygen is not an equivalent command, sshd-keygen regenerates the host keys. A `dpkg-reconfigure openssh-server` would be closer, IIRC.

GregSutcliffe: Ouch, missed the 'd' in sshd. Still, ssh-keygen can specifcy protocols, bits, and output files - presumably it could be called appropriately? Something like `ssh-keygen -b 2048 -t rsa -f /etc/ssh/ssh_host_rsa_key` maybe?

domcleal: I don't think it'd be appropriate here to store the details of what host keys should be present, leave that to the OS tools if at all possible. Key types, lengths and locations will vary between OSes and versions, hardcoding that would cause it to be obsolete quickly.

GregSutcliffe: For interest, I unpacked the openssh-server package - digging into the package, the postinst calls `ssh-keygen -q -f "$file" -N '' "$@"`, iterating over the various keys types. Basically the same, you just get the convenience of it doing all the key types in one command.

ekohl: It looks like sshd-keygen on Fedora also calls ssh-keygen, but I agree we shouldn't emulate the OS tools. On Debian there's `/var/lib/dpkg/info/openssh-server.postinst` which creates the keys. I think the dpkg-reconfigure command is a much better way.

GregSutcliffe: Yeah, I was mostly curious as to how compatible it was - it's good to know what the fallback is if we come across an OS without a decent way to regenerate.

lfisher47:
I'm not familiar enough with Debian to know but is this necessary during build time? It sounds like the command is run at time of install, while in EL it's run during the first run of the service which is after the first boot.

snippets/freeipa_register.erb
```diff
@@ -24,12 +24,16 @@ name: freeipa_register
...
 #   freeipa_opts                Additional options to pass directly to installer
 #
+
 <% if @host.operatingsystem.family == 'Redhat' -%>
   <% if @host.operatingsystem.name == 'Fedora' -%>
     freeipa_client=freeipa-client
   <% else -%>
     freeipa_client=ipa-client
   <% end -%>
+  <% if os_major > 6 -%>
```
domcleal: `os_major` won't/may not be defined here as it's not in this snippet. Use `@host.operatingsystem.major.to_i` or define it in the header.

domcleal:
Merged, thanks @lfisher47!

PR #289, opened 2016-07-29
Static mode in PXELinux template

lzap:
This might not make sense at first sight, but this template is used for Full
Host bootdisk and it currently requires DHCP, but with this patch it should
work in non-DHCP environments (when set for the subnet).

@domcleal what you think?


kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
     options << "modprobe.blacklist=" + @host.params['blacklist'].gsub(' ', '')
   end
   options = options.join(' ')
+  # handle non-DHCP environments (e.g. bootdisk)
+  subnet = @host.subnet
+  if subnet.respond_to?(:dhcp_boot_mode?)
```
domcleal: `subnet` will always respond to `dhcp_boot_mode?` on the develop branch, no conditional required

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
...
+    dhcp = subnet.dhcp_boot_mode? && !@static
+  else
+    dhcp = !@static
```
domcleal: `@static` should be removed, it's obsolete in favour of the subnet configuration

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
...
+  end
+  if dhcp
+    ksurl = foreman_url('provision')
+    legacy_net = net = ''
+  else
+    ksurl = foreman_url('provision') + '&static=yes'
+    mac = @host.mac
```
domcleal: Shouldn't this be the provision interface, not primary?

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
...
+  if subnet.respond_to?(:dhcp_boot_mode?)
+    dhcp = subnet.dhcp_boot_mode? && !@static
+  else
+    dhcp = !@static
+  end
+  if dhcp
+    ksurl = foreman_url('provision')
+    legacy_net = net = ''
+  else
+    ksurl = foreman_url('provision') + '&static=yes'
```
domcleal: static's obsolete, rely on the subnet boot mode

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
...
+    mac = @host.mac
+    bootif = '00-' + mac.gsub(':', '-') if mac
+    ip = @host.ip
```
domcleal: ditto, provision?

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
     options << "modprobe.blacklist=" + @host.params['blacklist'].gsub(' ', '')
   end
   options = options.join(' ')
+  # handle non-DHCP environments (e.g. bootdisk)
+  subnet = @host.subnet
```
domcleal: provision interface?

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
...
+  else
+    ksurl = foreman_url('provision') + '&static=yes'
+    mac = @host.mac
+    bootif = '00-' + mac.gsub(':', '-') if mac
+    ip = @host.ip
+    mask = subnet.mask
+    gw = subnet.gateway
+    dns = subnet.dns_primary
+    net = "ip=#{ip}::#{gw}:#{mask}:::none nameserver=#{dns} BOOTIF=#{bootif}"
+    legacy_net = "ip=#{ip} netmask=#{mask} gateway=#{gw} dns=#{dns} ksdevice=#{mac} BOOTIF=#{bootif}"
```
domcleal: Where is legacy_net used?

kickstart/PXELinux.erb
domcleal: .. in which case, there's no use in pulling `ksurl` out, it can remain inline.

kickstart/PXELinux.erb
```diff
@@ -20,19 +20,40 @@ oses:
...
+  end
 -%>
 
 DEFAULT linux
 
 LABEL linux
     KERNEL <%= @kernel %>
     <% if @host.operatingsystem.name.match(/.*atomic.*/i) %>
-    APPEND initrd=<%= @initrd %> ks=<%= foreman_url('provision') %> repo=<%= @host.operatingsystem.medium_uri(@host) %> ks.device=bootif network ks.sendmac <%= options %>
+    APPEND initrd=<%= @initrd %> ks=<%= ksurl %> repo=<%= @host.operatingsystem.medium_uri(@host) %> ks.device=bootif network ks.sendmac <%= options %>
```
domcleal: missing `net`?

kickstart/PXELinux.erb
lzap: Sure, I took this bit from other template, I can fix them all perhaps.

kickstart/PXELinux.erb
lzap: Yeah.

kickstart/PXELinux.erb
lzap: Haven't realized that we delegate to primary and not provision. Indeed.

lzap:
I have rewritten the patch from scratch, I removed the ugly condition for the APPEND entry and instead I put all required common options into the array. Since kernel command line options have no validation and Anaconda won't fall if you provide an unknown one, I believe the huge condition was not necessary for DHCP mode at all (one exception - Atomic which is handled correctly).

Only for static mode we have some branching and setting appropriate options. It looks much cleaner now. Please have a look, I haven't tested this myself yet. Want to see your opinion first. Looks saner to me.

domcleal:
Please test it before review, thanks.

mmoll:
also, this needs a rebase now :)

PR #292, opened 2016-08-15
Allow users to set ssh keys param manually

dLobatog:
To be merged after #16107 is fixed in REX - https://github.com/theforeman/foreman_remote_execution/pull/199


mmoll:
The r_ex chnage got rejected.

PR #293, opened 2016-08-17
Use integer comparison

UnwashedMeme:
As CoreOS version numbers are now greater than 1000 the lexical comparison fails; convert to integer and compare using that.

### DIFF
--- a/coreos/provision.erb
+++ b/coreos/provision.erb
@@ -5,6 +5,9 @@ name: CoreOS provision
 oses:
 - CoreOS
 %>
+<%
+os_major  = @host.operatingsystem.major.to_i
+%>
 coreos:
   units:
     - name: coreos-bootstrap.service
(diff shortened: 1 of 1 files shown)


mmoll:
:+1: 
Could you extract that to an `os_major` variable in the header, like it's some for some other OSes? That line is already quite long. :)

UnwashedMeme:
I'll admit I might have been looking for a "Rolling-eyes" emoji at that request but this does read a hell of a lot easier. It is especially easier to try and edit inside foreman's in-browser editor so yeah that was a good call.

mmoll:
@UnwashedMeme personally I use the foreman_templates plugin to sync all the templates from this git repo here, so I'm rarely using that editor... ;)
@johscheuer please have a look

coreos/provision.erb
```diff
@@ -14,7 +17,13 @@ coreos:
         [Unit]
         Description=Install coreos to disk
         [Service]
-        ExecStart=/bin/bash -c "/usr/bin/coreos-install -C  <%= @host.operatingsystem.release_name %> -d <%= @host.params['install-disk'] || '/dev/sda' %> -c /home/core/cloud-config.yml <% if @host.operatingsystem.major >= "557" -%>-b <%= @mediapath %><% end -%>  && wget -q -O /dev/null --no-check-certificate <%= foreman_url('built') %> && reboot"
+        Type=oneshot
+        ExecStart=/usr/bin/coreos-install \
+          -C <%= @host.operatingsystem.release_name %> \
+          -d <%= @host.params['install-disk'] || '/dev/sda' %> \
+          -c /home/core/cloud-config.yml <% if os_major >= 557 -%>-b <%= @mediapath %><% end %>
```
johscheuer, line 24: @UnwashedMeme @mmoll is this really a Use Case to support such an old CoreOS version? [557.0](https://coreos.com/releases/#557.0.0) came out January 6, 2015 ...  with `Docker 1.3` + `etcd 0.4.6` personally I think we should just drop this part with `<% if os_major >= 557 -%>` if somebody really wants to use such an old version he just have to remove the part `-b <%= @mediapath %>`.

I mean you use CoreOS because you want a minimal (always) up to date Linux Distro (also after the reboot CoreOS would fetch the newest version of the channel and install this version (I fixed this one https://github.com/theforeman/community-templates/pull/294 )^^).

johscheuer:
👍 to the reformatting and update the systemd service :)

UnwashedMeme:
>  is this really a Use Case to support such an old CoreOS version?

It certainly isn't for me; I tend to agree that it's kind of against the point of CoreOS but, once fixed, the conditional wasn't bothering me and I figured leaving it in was less controversial.

mmoll:
It doesn't hurt to keep it, IMHO

mmoll:
merged, thanks @UnwashedMeme!

PR #294, opened 2016-08-18
Allow setting reboot strategy

johscheuer:
This PR allows users to set the reboot-strategy of their CoreOS instances for example if a user doesn't want automatic updates/reboots he can set the value to `off`, here is the official documentation for the [reboot-strategy](https://coreos.com/os/docs/latest/update-strategies.html).

I also added this option in the Readme.md


mmoll:
danke @johscheuer!

PR #295, opened 2016-08-26
Fixed PXE templates for safemode

lzap:
Template PXE changes to support safemode. Regression introduced by

https://github.com/theforeman/community-templates/commit/cd4c8cdcd1d4bcbbdda43c5fe74c9b2fd4254c8d

with attempt to DRY. New snippet was introduced to carry on common PXE kernel
command line options for particular OS. This really belong to a Core helper.

Note I am dropping snippet in this patch. We need to modify seeds script in
core.

Also I am moving vim modelines to the top, filed patch to templates plugin to
handle that correctly:

https://github.com/theforeman/foreman_templates/pull/28


dLobatog:
This would need changing the ticket number by something that's exclusive to this project, issue numbers shouldn't be shared between projects. Other than that :+1: , works well with the pxe_kernel_options patch in core

mmoll:
I don't think the ticket number needs to get changed... "refs" is correct here, no?

dLobatog:
I think the tests are failing because of the vim stuff is now on top of the file, maybe you can leave that for another PR?

lzap:
> I think the tests are failing because of the vim stuff is now on top of the file, maybe you can leave that for another PR?

Sure, removed modelines completely, it was causing errors when opening
with vim (because of the `%>`). Will add that later.

## 

Later,
 Lukas #lzap Zapletal

dLobatog:
Core PR was merged :smile: 

@mmoll The problem with having a same issue number in the tracker reference commits in 2 different projects, is that when we cherry-pick stuff for a release, you can't cherry-pick stuff from a different repo. Plus the 'release' field becomes useless for one of the 2 projects that share the refs.

lzap:
@domcleal could you please import templates once again into develop? This fixes the safemode issue, there is also one line change

```
default linux
```

which was missed when I was syncing the community-templates PR vs core PR.

Importing will solve both.

lzap:
Oh I was reading your previous comment as "this was merged". Ok, reworder the commit message, this repo does not require redmine, so leaving vanilla.

lzap:
@dLobatog when merging this, can you sync template into foreman core? There is a script there, you need to have one utility installed to have it working: https://github.com/lzap/bin-public/blob/master/prename (not packaged for Fedora fyi)

dLobatog:
@lzap I don't have merge access here, sorry, but yeah it should be merged for 1.13 which is going to be soon

mmoll:
merged, díky @lzap!

PR #296, opened 2016-09-01
Refs #16416 -  Retrieve password_hash from OS

sean797:
Requires theforeman/foreman#3805

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -56,7 +56,7 @@ firewall --disable
 <% else -%>
 firewall --<%= os_major >= 6 ? 'service=' : '' %>ssh
 <% end -%>
-authconfig --useshadow --passalgo=sha256 --kickstart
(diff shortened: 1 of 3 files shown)


sean797:
[test]

kickstart/provision.erb
```diff
@@ -56,7 +56,7 @@ firewall --disable
 <% else -%>
 firewall --<%= os_major >= 6 ? 'service=' : '' %>ssh
 <% end -%>
-authconfig --useshadow --passalgo=sha256 --kickstart
+authconfig --useshadow --passalgo=<%= @host.operatingsystem.password_hash %> --kickstart
```
ekohl: Maybe you can fall back to sha256 if it's not set? That way we stay compatible with older foreman versions.

ekohl:
There's a typo in the commit message

mmoll:
Also, tests need updating. :+1: otherwise

mmoll:
merged as 519490c, thanks @sean797!

PR #297, opened 2016-09-09
RFC - Add optional dns_alt_names to puppet.conf snippet

matonb:
Add optional parameter dns_alt_names to puppet.conf.erb

My use case is when deploying puppet / foreman servers I need to specify dns_alt_names that will be used to access the services.

main_pad is just to keep the values nicely aligned, 'printf' not allowed in safemode ?


snippets/puppet.conf.erb
```diff
@@ -18,12 +18,16 @@ name: puppet.conf
...
 %>
 [main]
-vardir = <%= var_dir %>
-logdir = <%= log_dir %>
-rundir = <%= run_dir %>
-ssldir = <%= ssl_dir %>
+<%- if !@host.params['dns_alt_names'].to_s.empty? -%>
+dns_alt_names = <%= @host.params['dns_alt_names'] %>
+<%- end -%>
+vardir<%= main_pad %>= <%= var_dir %>
```
dLobatog: Aren't these missing `=`?

mmoll:
Am I understanding this correctly, this is only about aligning some values in that configfile?
In general I'd recommend to use theforeman-puppet for all the Puppet configuration, that will also overwrite puppet.conf with a hopefully nice version. ;)

matonb:
@mmoll 
No, not just alignment, the variable 'main_pad' is for alignment,
The real change here is adding `dns_alt_names` if the parameter is defined for the host.

snippets/puppet.conf.erb
matonb: The if statement isn't going in to the generated template ?

mmoll:
@matonb ok, could you please make a minimal version without the alignment, the change itself sounds useful. :)

The paramter itself should maybe be something more telling, like `puppetmaster_dns_alt_names` or smimilar.

matonb:
@mmoll 
I'm happy to adjust the parameter name, but it will affect for any node getting a certificate from a Puppet CA, foreman in my case would also use `dns_alt_names` for example.

I'll force up the change without alignment in a tick :)

matonb:
@mmoll 
How about  `puppetcert_dns_alt_names` instead of `dns_alt_names` ?

mmoll:
@matonb yes, that would be fine :+1:

ekohl:
Another option would be puppet_dns_alt_names.

matonb:
After a brief deliberation, I opted for `puppetcert_dns_alt_names`

snippets/puppet.conf.erb
```diff
@@ -20,6 +20,9 @@ name: puppet.conf
   end
 %>
 [main]
+<%- if !@host.params['puppetcert_dns_alt_names'].to_s.empty? -%>
```
ekohl: I think `unless @host.params` is easier to read.

ekohl:
:+1: besides the minor nit.

snippets/puppet.conf.erb
matonb: Please excuse my roobness :wink:

ekohl:
:+1: from my side, @mmoll?

mmoll:
merged, thanks @matonb!

PR #298, opened 2016-09-10
RFC: Make base package update optional

matonb:
Use case
- Systems not connected to the internet
- User simply doesn't want to update the installed OS packages

`if update_base_packages` would require a change (default to true) to maintain current behavior with previous template releases.

Open to suggestions of another way to implement or add `no_update_base_packages` instead ?


mmoll:
please use e.g. `@host.param_false?('update_base_packages')` there are true/false helpers for these kind of determinators, and that way it stays compatible :)

matonb:
:) Cool thanks for the pointer

matonb:
`@host.param_false?('update_base_packages')`  that should be

`@host.param_true?('update_base_packages')`
or
 `@host.param_false?('no_update_base_packages')`

Which sits better?

mmoll:
In order to retain compatibility (no change when the param is not there, default to update), I'd go for `unless @host.param_false?('update_base_packages')` personally, as it seems that most people do want this to happen, so that param will not exist most of the time.

matonb:
@mmoll sounds good to me, updated the PR

kickstart/provision.erb
```diff
@@ -183,12 +183,14 @@ echo "updating system time"
 echo 'proxy = <%= proxy_uri %>' >> /etc/yum.conf
 <% end -%>
 
+<% unless @host.param_false?('update_base_packages') -%>
```
domcleal: There's already a `package_upgrade` parameter used in cloudinit/userdata_cloudinit.erb, so I'd suggest renaming this parameter. Please also add it to the docs at the top of the file.

matonb: @domcleal I couldn't see what `package_upgrade` actually does from community_templates, I'm guessing it's something to do with `yum **update**` vs `yum **upgrade**`

What would you suggest as an acceptable parameter name to define whether or not `yum update` should happen or not ?

Sure I'll add it to the readme once we've settled on a parameter name.

domcleal: It used in the cloudinit template to run `yum upgrade` (for EL-based distros anyway, cloud-init abstracts it) automatically. I don't think the minor distinction between update and upgrade is important or particularly intentional, they both upgrade the majority of packages (not just refresh metadata).

The parameter should probably be package_upgrade.

matonb: Ok that makes sense

matonb: @domcleal provision/finish.erb hasn't got any of the parameters in the doc section, if this is an omission rather than intentional I can have a look through and a parameter doc block to that file?

domcleal: That's probably an omission, more docs are always welcome.

matonb: Ok no problem I'll add a parameter section.

mmoll:
merged, thanks @matonb!

PR #300, opened 2016-09-26
Enable chainbooting via chain.c32

lzap:
We have a report from RHEV QE that Foreman does not chainload on RHEV 3.6.8 VM
(VDSM 4.17.33) with the following error:

https://postimg.org/image/iyw17zby9/

Using chain.c32 (deployed by default by our installer) workarounds it. It's a
recommended way of chainbooting in SYSLINUX WIKI:

http://www.syslinux.org/wiki/index.php?title=Hardware_Compatibility#LOCALBOOT

This patch sets it as default and keeps the old way as "legacy" option.

Edit: I was also told nested KVM was enabled on this hypervisor.


ekohl (approved):
I don't know PXELinux that well, but it looks correct.

domcleal:
Note that [chain.c32 has no EFI support](https://bugzilla.syslinux.org/show_bug.cgi?id=17), while [LOCALBOOT does have EFI support](http://git.kernel.org/cgit/boot/syslinux/syslinux.git/commit/?id=4f2d3e62b61ebe21498844c0ca346482aca118c9) in 6.0.2.

Should the assignment of "legacy" be the other way around, so prefer LOCALBOOT and have chain.c32 as a workaround option?

lzap:
Sure, added as an alternative. I haven't noticed, no.

domcleal:
Thanks @lzap.

PR #304, opened 2016-10-07
Support Puppet 4 AIO packages on SLES

mmoll:
Because of SLES brokenness (see e.g. puppetlabs/puppetlabs-release#80) the RPM keys have to get imported in an extra step.

### DIFF
--- a/autoyast/provision_sles.erb
+++ b/autoyast/provision_sles.erb
@@ -73,7 +73,11 @@ oses:
       <package>lsb-release</package>
       <package>openssh</package>
 <% if puppet_enabled -%>
+<% if @host.param_true?('enable-puppetlabs-pc1-repo') -%>
+      <package>puppet-agent</package>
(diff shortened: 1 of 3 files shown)


autoyast/provision_sles.erb
```diff
@@ -73,8 +73,12 @@ oses:
       <package>lsb-release</package>
       <package>openssh</package>
 <% if puppet_enabled -%>
+<% if @host.param_true?('enable-puppetlabs-pc1-repo') -%>
+    <package>puppet-agent</package>
```
domcleal: Nitpick, indentation is off here

domcleal:
:+1: otherwise.

autoyast/provision_sles.erb
mmoll: done.

domcleal (approved):

PR #305, opened 2016-10-12
Fixes #16882 - fixed absolute path for Grub1 kickstart

lzap:
Grub1 was giving this error

http://storage7.static.itmages.com/i/16/1012/h_1476270469_9345664_e3371f8da3.png

Unfortunately Grub1 does not allow relative paths like Grub2 and errors out
with `Filename must be either an absolute path or a blocklist`. Therefore the
kernel/image lines must be written this way.

### DIFF
--- a/kickstart/PXEGrub.erb
+++ b/kickstart/PXEGrub.erb
@@ -14,12 +14,12 @@ timeout=<%= @host.params['loader_timeout'] || 10 %>
 title <%= template_name %>
   root (nd)
   <% if @host.operatingsystem.name.match(/.*atomic.*/i) -%>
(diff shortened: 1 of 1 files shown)


mmoll:
@rplevka I guess you tested this? :)

rplevka:
@mmoll, that's correct

mmoll:
Merged and also cherry-picked as ba69ac2 into 1.13-stable, díky @lzap!

lzap:
@domcleal are you going to sync templates for the next minor 1.13 release, or shall I make a core patch for 1.13-stable?

domcleal:
They are always synced during the release process.

lzap:
Just to be absolutely clear - what I mean is if this gets into 1.13.X minor one. I take this as yes, unless you comment on that, thanks :-)

PR #307, opened 2016-10-21
Update PXELinux.erb

narbutas:
Fixing PXELinux template problem regarding missing variables


mmoll:
@lzap could you have a look? this would be required in other files using that variable also, then

domcleal:
It's a helper that should be available on Foreman 1.13.0 or higher when rendering. I don't think it's provided as an instance variable, so I don't see how this populates the data.

Are you using the template with an older version?

PR #309, opened 2016-11-09
Add `disable-uek` parameter

alexjfisher:
If `disable-uek` is set to true, Oracle Unbreakable Kernel packages are
uninstalled in the `%post` section and the host will come up running the
RedHat compatible kernel on reboot.

### DIFF
--- a/kickstart/provision.erb
+++ b/kickstart/provision.erb
@@ -22,6 +22,7 @@ This template accepts the following parameters:
 - bootloader-append: string (default="nofb quiet splash=quiet")
 - disable-firewall: boolean (default=false)
 - package_upgrade: boolean (default=true)
+- disable-uek: boolean (default=false)
 %>
 <%
   rhel_compatible = @host.operatingsystem.family == 'Redhat' && @host.operatingsystem.name != 'Fedora'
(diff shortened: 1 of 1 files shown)


mmoll:
Is there no way to set something in the kickstart?

alexjfisher:
@mmoll Not that I could find.  See this thread.
https://www.redhat.com/archives/spacewalk-list/2015-March/msg00058.html

kickstart/provision.erb
```diff
@@ -207,6 +208,12 @@ fi
 <%= snippet 'saltstack_setup' %>
 <% end -%>
 
+<% if @host.param_true?('disable-uek') -%>
```
mmoll: maybe add `if @host.operatingsystem.name == 'OracleLinux'` here, just to be safe?

alexjfisher: @mmoll Sure.  It's probably not necessary, but certainly wouldn't hurt either.

mmoll: exactly

alexjfisher: done

mmoll:
merged, thanks @alexjfisher!

PR #312, opened 2016-11-14
coreos_cloudconfig missing end

dLobatog:
The latest commit introduced some conditions in this file without the
corresponding end.


mikedevita:
this works for me fyi, it was missing the `<% end %>`

mmoll:
I'm unsure if the end shouldn't go into line 28 instead of the else...

dLobatog:
@mmoll You're right - otherwise if you set the update-strategy param the rendered snippet doesn't make much sense. Updated

mmoll:
merged, gracias @dLobatog!

PR #313, opened 2016-11-21
use puppetlabs-release.deb for PL repo

### LINKED ISSUE #303: Installation of puppet-agent fails
Hello,

since a few days, when I provision a new VM (in this case Ubuntu 16.04.1 LTS) it fails during the "Select and install software" step with the error

> WARNING: The following packages cannot be authenticated!
> puppet-agent

puppet_auth_issue

mmoll:
closes GH-303


preseed/provision.erb
```diff
@@ -131,8 +105,8 @@ d-i apt-setup/local<%= repos %>/key string http://keyserver.ubuntu.com/pks/looku
 # Install minimal task set (see tasksel --task-packages minimal)
 tasksel tasksel/first multiselect minimal, ssh-server, openssh-server
 
-<% if puppet_enabled %>
-<% puppet_package = @host.param_true?('enable-puppetlabs-pc1-repo') ? 'puppet-agent' : 'puppet' -%>
+<% if puppet_enabled && !(@host.param_true?('enable-puppetlabs-repo') || @host.param_true?('enable-puppetlabs-pc1-repo')) %>
+<% puppet_package = 'puppet' -%>
```
ekohl, line 109: Was puppet-agent renamed to puppet in pc-1 as well?

ekohl (commented):
What's the reasoning behind this? You need to update ca-certificates on certain distros to be able to use the puppet repos?

preseed/provision.erb
mmoll: This is (at least for now) only used for the puppet packages from native OS repos.

mmoll:
Debian has ca-certificates not installed by default, wget can't verify the cert chain then and any https download fails.

domcleal (approved):

preseed/provision.erb
ekohl: Does that mean that on older distros that ship puppet 3 by default it's now no longer possible to install puppet 4?

mmoll: When using any of the puppetlabs repos, the respective package is installed in the finish script and not in the preseed step.

ekohl: Should it just be an empty string then? This implementation implies puppet 3 is installed during preseed and upgraded to 4 in the finish script.

mmoll: If puppet is enabled and no puppetlabs repo is enabled, then the native OS package should be used (in whatever version the OS is bringing). In the finish script, there will be a redundant "apt-get install puppet" in that case, but that should be harmless.

If puppet is enabled and any PL repo is enabled, puppet (puppet3) or puppet-agent (puppet4) will be installed .

At least that's what I had in mind.

ekohl (approved):

ekohl:
Thanks!

PR #317, opened 2016-11-22
adjusts package action template for debian and adds suse support

dgoetz:
adds shebang line and removes brackets from function as it caused problems on debian
adds update logic to debian as update only updates metadata
adds suse support using zypper in non-interactive mode


mmoll (dismissed):

mmoll:
Please don't indent things inside ERB blocks here also, as the resulting, rendered shell output will have all the spaces in it.

dgoetz:
Removed the indention and force pushed the amended change.

mmoll:
merged, danke @dgoetz!

PR #322, opened 2016-12-08
Added repo-mirror for puppet and epel repos

jkroepke:
Hi!

we have a repo mirror because our machines does not have internet access. 

Here is a patch add the variable repo-mirror for RHEL based systems.


mmoll:
@jkroepke Please see #267 for comments on a very similar PR. Side note: after #313 this is probably even harder now.

jkroepke:
> Side note: after #313 this is probably even harder now.

Did not see, why it's harder now.

> Please see #267 for comments on a very similar PR.

Okay, let's discuss it there.

mmoll:
see #267

Issue #328, opened 2016-12-16
Preseed passwd/make-user typo


derektamsen:
It appears that the `preseed/provision.erb` template has a typo.

`d-i passwd passwd/make-user boolean false` fails validation checks and should be set as `d-i passwd/make-user boolean false`.

The syntax is also reflected in both the Debian and Ubuntu example preseed files:
- https://www.debian.org/releases/wheezy/example-preseed.txt
- https://help.ubuntu.com/lts/installation-guide/example-preseed.txt

PR #329, opened 2016-12-16
Fix preseed make user typo

derektamsen:
- Typo in d-i line to set `passwd/make-user` false
  - ref: https://www.debian.org/releases/wheezy/example-preseed.txt
  - ref: https://help.ubuntu.com/lts/installation-guide/example-preseed.txt
- Resolves GH-328


ekohl:
Thanks!

Issue #330, opened 2016-12-16
Cannot find a question for console-keymaps-at/keymap


derektamsen:
`debconf-set-selections` showing `Cannot find a question for console-keymaps-at/keymap` error when running `preseed/provision.erb` against newer os. The `console-keymaps-at/keymap` was removed from wheezy and newer per https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=681801.

derektamsen:
I will fix by wrapping the debian installer command in an os version check.

PR #331, opened 2016-12-16
Only set old keymap on Debian Squeeze or older installs

derektamsen:
- only set `console-keymaps-at/keymap seen true` in `preseed/provision.erb` when Debian OS Version is <= Squeeze (6.0).
- Resolves GH-330


ekohl (approved):
Makes sense to me. @mmoll?

mmoll:
While I didn't run into the error, according to the linkd Debian issue, this is looking correct.

mmoll:
merged, thanks @derektamsen!

derektamsen:
Great thanks! I found the errors when working on the travis tests :smile:

Issue #334, opened 2016-12-19
Add preseed param to install security updates post installation


derektamsen:
Debian based oses have an option to immediately install security updates post preseed installation. It would be great to specify this as a host param.

https://help.ubuntu.com/lts/installation-guide/amd64/apbs04.html#preseed-pkgsel

PR #335, opened 2016-12-19
Add "preseed-post-install-upgrade" option with default to none

derektamsen:
- Specify security upgrade options post installation. Choices are none, safe-upgrade, full-upgrade.
- Resolves GH-334


ekohl (approved):
FYI we don't require a separate issue, just a PR is also fine.

derektamsen:
Thanks! Sounds good. I have a couple more that I will just open PRs for.

mmoll:
I'm fine with this, but wonder what the default of the installer is, because I think after a Foreman install of any Debian, I never needed updates, so maybe "none" is not the actual default (or maybe I was just lucky).

derektamsen:
@mmoll I'm not really sure what the default of this is. I think their may also be a bug around the option but can't confirm. I was assuming the default was supposed to be `none` due to the documentation. I was having the opposite issue where I did not want the system to be upgraded and the option seemed to have no affect.

Based on my tests with the key I think the installation will always have the latest installed because if a security mirror is not specified it points to security.ubuntu.com (for ubuntu) via the debian-installer. Then, the installer selects the linux-image which will almost always be installed from the security repo.

From: https://help.ubuntu.com/lts/installation-guide/amd64/apbs04.html#preseed-pkgsel
```
# Whether to upgrade packages after debootstrap.
# Allowed values: none, safe-upgrade, full-upgrade
#d-i pkgsel/upgrade select none
```

mmoll:
it for sure won't hurt, merged, thanks @derektamsen!

Issue #336, opened 2016-12-19
Specify preseed live-installer squashfs path if using iso repos


derektamsen:
When installing Ubuntu 14.04+ from an iso copied to an http mirror the installer will fail while attempting to locate the live-installer. This appears to be a bug in the installation media where the path to the installer is set to a path that does not exist when netbooted. This can be addressed by specifying `d-i live-installer/net-image string http://<%= @preseed_server %><%= @preseed_path %>/install/filesystem.squashfs` in the preseed config.

See:
- http://www.michaelm.info/blog/?p=1378
- https://bugs.launchpad.net/ubuntu/+source/live-installer/+bug/1042906

PR #337, opened 2016-12-19
optionally specify using a live-installer

derektamsen:
- Set to `true` if using an internal mirror copied from an ubuntu installation iso.
- Addresses: https://bugs.launchpad.net/ubuntu/+source/live-installer/+bug/1042906
- Resolves GH-336


mmoll:
merged, thanks @derektamsen!

PR #342, opened 2017-01-09
kickstart finish: root_pass may be nil

timogoebel:
This commit fixes an issue, where `empty?` is called on root_pass which
has a nil value. This causes the template rendering to fail as `empty?`
is undefined for NilClass.


mmoll (approved):

mmoll:
@timogoebel could you test/review #323, if you're using RH/CentOS?

mmoll:
merged, danke @timogoebel!

PR #344, opened 2017-01-16
Change directory structure

ares:
Based on discussion on [foreman-dev](https://groups.google.com/forum/#!topic/foreman-dev/bzfBWEtIMpg) I'm proposing to change directory structure of this repo so one can use foreman_templates for importing and exporting with this repo easily.

I took clean Foreman nightly, imported last version of templates from here, exported with keeping existing metadata. There were few readme file in some directories so I consolidated it in main README.md but maybe we can remove it completely since they are mostly out of date. I kept job template readme in job_templates dir.

For some reason we have one more snippet in core codebase so I added it here too it's [http_proxy snippet](https://github.com/theforeman/foreman/blob/develop/app/views/unattended/snippets/_http_proxy.erb)


ares:
Also doing an export identified we have name mismatch between our seed file and template content in
* https://github.com/theforeman/foreman/blob/698e916ce208b5040b83a908a058c83c94d158ee/db/seeds.d/08-partition_tables.rb#L10
* https://github.com/theforeman/foreman/blob/698e916ce208b5040b83a908a058c83c94d158ee/db/seeds.d/08-partition_tables.rb#L17

We could provide sync names and provide a migration for existing data or let user synchronize with this repo.

ekohl:
The directory names are bit inconsistent. Partition tables are templates as well so I feel that's redundant. I'd suggest:
* jobs
* partition_tables
* provisioning

But I'd be fine with whatever, this structure is already better than the old one and would allow us to easily place labels depending on which directories are changed.

Other than that there's also a merge conflict that needs to be solved.

Regarding the seed file and template content I'm not sure what's the best. We do have different branches for each version so that could help you fix this in a consistent way. Is possible to solve that as a separate issue and get this merged first?

ares:
> The directory names are bit inconsistent. Partition tables are templates as well so I feel that's redundant. I'd suggest:

Ideally each template class would define some "human name" which might take longer to propagate to rex plugin. So maybe we can simply cut off '_template' during the export in foreman_templates, I'll see how it would feel during the rebase. In general :+1: 

> Other than that there's also a merge conflict that needs to be solved.

If you tell me you're willing to merge soon, I'll rebase. Every new PR that's merged here makes this out of date :-)

> Regarding the seed file and template content I'm not sure what's the best. We do have different branches for each version so that could help you fix this in a consistent way. Is possible to solve that as a separate issue and get this merged first?

I'd prefer this approach too.

mmoll:
I thought #343 to get in first, let's see how things develop regarding it today...

ekohl:
> Ideally each template class would define some "human name" which might take longer to propagate to rex plugin. So maybe we can simply cut off '_template' during the export in foreman_templates, I'll see how it would feel during the rebase. In general 👍

I was talking to @GregSutcliffe and he pointed out that we may want to ship foreman hook examples as well. In that case the _templates suffix might be a good thing. Would make more sense to have ```partition_table_templates``` as well? Given they're ERB I consider them templates.

ares:
well, if we put hooks here too then the repo should be renamed I guess :-) but let's reconsider when hooks are included, I'll try to add '_templates' suffix to partition tables then.

ekohl:
I'd prefer to get this merged before #343 given there is no discussion about this and quickly gets out of date.

mmoll:
also true... @ares please rebase :)

ares:
@ekohl rebased

provisioning_templates/snippet/http_proxy.erb
```diff
@@ -0,0 +1,12 @@
+<%#
```
mmoll, line 1: I guess that file slipped in unintentionally? :)

ares:
Renamed `partition_tables` to `partition_tables_templates`. Should be ready now.

provisioning_templates/snippet/http_proxy.erb
ares: nope, that's intentional

> For some reason we have one more snippet in core codebase so I added it here too it's [http_proxy](https://github.com/theforeman/foreman/blob/develop/app/views/unattended/snippets/_http_proxy.erb) snippet

mmoll: Hrm... I guess that's some historical leftover in core, tbh...

ares: I think so too, but I wanted to sync it with this repo, we can remove it separately if that makes sense (some people might use it though)

mmoll: OK, let's pull it in for now, maybe we can even use it to generalize some of the proxy stuff later...

ekohl (approved):
We should update the README because the list of supported OSes is outdated, but that shouldn't hold up this PR.

mmoll:
merged, díky @ares!

PR #346, opened 2017-01-18
Add Preseed default PXEGrub2

sathieu:
See http://projects.theforeman.org/issues/16654

### DIFF
--- a/preseed/PXEGrub2.erb
+++ b/preseed/PXEGrub2.erb
@@ -1,0 +1,40 @@
+<%#
+kind: PXEGrub2
+name: Preseed default PXEGrub2
+oses:
+- Debian
+- Ubuntu
+%>
+#
+# This file was deployed via '<%= template_name %>' template
+#
+# Supported host/hostgroup parameters:
+#
+# blacklist = module1, module2
+#   Blacklisted kernel modules
+#
+# lang = en_US
+#   System locale
+#
+<%
+  options = []
+  if @host.params['blacklist']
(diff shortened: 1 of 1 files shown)


lzap:
Note from the issue: Does not actually work with Debian Grub2.

lzap:
So the reason why Debian templates were not yet implemented was because I did not get to it, I only use Red Hat systems for my development. Now, Fedora-patched grub made a lot of confusion already when tutorials on the web does not work on other systems. But there is an easy way out, luckily!

Grub2 has very powerful scripting capabilities, you can easily make a script (grub.cfg file) that will basically "emulate" Fedora-patched grub behavior so it will read the correct configuration files via MAC address. The configuration scheme used in Fedora and Red Hats is compatible with yaboot which is used on older IBM POWER systems, so it's ideal to keep it as is.

Here is example snippet that will need to be placed as `grub2/grub.cfg`:

http://lists.gnu.org/archive/html/help-grub/2016-10/msg00041.html

Or you can google out couple of others, I haven't tested this one. Can you try this with Grub2 from Debian or Ubuntu? We can then work together in shipping this default grub with Foreman via Puppet installer.

sathieu:
@lzap The problem is: there is no "regexp" function!

mmoll:
merged, merci @sathieu!

ekohl:
We forgot to check the paths. https://github.com/theforeman/community-templates/pull/350 should move it to the correct location.

PR #347, opened 2017-01-19
EFI Boot Debian

nullkarma:


mmoll:
@nullkarma please see #346

nullkarma:
Seems to be a dup. Actually [this requests](https://github.com/theforeman/community-templates/pull/346) does already exist.

PR #348, opened 2017-01-23
Refactor kickstart_networking_setup template

ares:
Needs a `save_to_file` macro from https://github.com/theforeman/foreman/pull/4216 and https://github.com/theforeman/foreman/pull/4222 for unsafe mode

### DIFF
--- a/provisioning_templates/snippet/kickstart_networking_setup.erb
+++ b/provisioning_templates/snippet/kickstart_networking_setup.erb
@@ -1,201 +1,45 @@
 <%#
-kind: snippet
 name: kickstart_networking_setup
-description: this will configure your host networking, it configures your primary interface as well
-    as other configures NICs. It supports physical, VLAN and Alias interfaces. It's intended to be
-    called from %post in your kickstart template. Note that this snippet can be used with Foreman 1.7
-    and later
-%>
-<% subnet = @host.subnet -%>
-<% subnet6 = @host.subnet6 -%>
-<% dhcp = subnet.dhcp_boot_mode? -%>
+snippet: true
+model: ProvisioningTemplate
+kind: snippet
+-%>
+<%- bonded_interfaces = [] -%>
+<%- @host.bond_interfaces.each do |bond| -%>
+<%=   "# #{bond.identifier} interface" %>
+<%=   snippet('kickstart_ifcfg_get_identifier_names', :variables => { :identifier => bond.identifier }) %>
+<%-   ifcfg = snippet('kickstart_ifcfg_bond_interface', :variables => {
+                        :interface => bond,
+                        :subnet => bond.subnet,
+                        :subnet6 => bond.subnet6,
+                        :dhcp => bond.subnet.nil? ? false : bond.subnet.dhcp_boot_mode? }) -%>
+<%=   save_to_file('/etc/sysconfig/network-scripts/ifcfg-$sanitized_real', ifcfg) %>
 
-<% if @host.respond_to?(:has_primary_interface?) %>
-<%# Foreman 1.7 - primary interface contained in @host %>
-# primary interface
-real=`ip -o link | grep <%= @host.mac -%> | awk '{print $2;}' | sed s/://`
-<% if @host.has_primary_interface? %>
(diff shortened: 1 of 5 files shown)


ares:
No one finds this better? :-)

mmoll:
theforeman/foreman#4222 is not merged, yet

ares:
Rebased on top of https://github.com/theforeman/community-templates/pull/359, every call on subnet should verify that subnet is not nil.

ekohl:
Given this is blocked on the save_to_file macro, can you write it without it first so this can be merged?

provisioning_templates/snippet/kickstart_networking_setup.erb
```diff
@@ -1,195 +1,45 @@
...
+<%- bonded_interfaces = [] -%>
+<%- @host.bond_interfaces.each do |bond| -%>
+<%=   "# #{bond.identifier} interface" %>
+<%=   snippet('kickstart_ifcfg_get_identifier_names', :variables => { :identifier => bond.identifier }) %>
+<%-   ifcfg = snippet('kickstart_ifcfg_bond_interface', :variables => {
+                        :interface => bond,
+                        :subnet => bond.subnet,
+                        :subnet6 => bond.subnet6,
+                        :dhcp => bond.subnet.nil? ? false : bond.subnet.dhcp_boot_mode? }) -%>
+<%=   save_to_file('/etc/sysconfig/network-scripts/ifcfg-$sanitized_real', ifcfg) %>
```
ekohl, line 16: To me the ```$sanitized_real``` part is non-obvious. I know how the save_to_file macro works, but it feels like exploiting side effect.

ares:
@ekohl save_to_file macro is merged, theforeman/foreman#4222 is not but it does not block this I think

provisioning_templates/snippet/kickstart_networking_setup.erb
ares: any suggestion how to make it better? would a comment that it's set by the snippet kickstart_ifcfg_get_identifier_names help? we can't get the value in ERB since it's only known when the template is executed on the target

ekohl: You're right so I suppose currently this is the best implementation.

ekohl (approved):
While I approve of this, we should be careful not to make it too much programming. I know a lot of sysadmins prefer relative straight forward templates where a lot of developers prefer abstractions. Those sysadmins sometimes have problems building a mental image of all the related content and end up opening all the related templates while going back and forth between them.

What could help is to make it easier to see the content of another snippet while viewing the template editor. Not sure how that would exactly look, but it's something to keep in mind while moving forward with changing the templates.

lzap:
Nice, this part absolutely deserved improvement. Have you considered creating a macro/function that would render a hash to KEY=value form that is used in rhcfg configuration files? The snippets would look a little bit nicer - instead multiple ERB lines you could simply use multiline block setting up the hash and then just call the function returning all the data. We could also sort them by keys.

mmoll:
this needs a rebase now

ares:
@ekohl I agree, I think this is the most complicated template where I think the abstraction makes future maintenance easier which hopefully justifies it. Maybe we could improve snippet macro to wrap the output with some comments such as

```
# snippet kickstart_network_setup start
$rendered content 
# snippet kickstart_network_setup end
```

which could help understand the flow in preview. Anyway I'll try to avoid unnecessary programming in other templates.

ekohl:
@ares I was more thinking along the lines of changing the editor to understand that ```snippet("name")``` should provide an easy way to see that snippet.

ares:
@lzap that macro could be useful but I think that's a good example of what @ekohl just mentioned. It's probably easier for sysadmins to see snippet such as

```
IPADDR=$something
NETMASK=$something
```

then 
```
snippet 'ifcfg_file', :variables => { :ipaddr => $something, :netmask => $something }
```

anyway we can always improve further in future if it turns out people would prefer this :-)

ares:
I rebased this PR. I'd appreciate to get this in before other changes in kickstart_networking_snippet :-)

mmoll:
did only light testing, but looked good, merged, díky @ares!

PR #349, opened 2017-01-24
enable net.ifnames=1 to allow biosdevnames

Raboo:
In my opinion net.ifnames=1 is a sane default value and should be set if not overridden via 'disable-net-ifnames' parameter.
The reason for my believe is that I feel that more Linux distros are moving towards using biosdevnames like em1, em2, etc.. instead of eth0, eth1.. and this would increase compatibility with newer Linux distributions.

### DIFF
--- a/provisioning_templates/snippet/pxegrub2_discovery.erb
+++ b/provisioning_templates/snippet/pxegrub2_discovery.erb
@@ -3,6 +3,6 @@ kind: snippet
 name: pxegrub2_discovery
 -%>
 menuentry 'Foreman Discovery Image' {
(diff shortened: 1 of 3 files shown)


Raboo:
Hold a moment, there is a problem..

    Error: There was an error creating the PXE Default file: The snippet 'pxelinux_discovery' threw an error: undefined method `param_true?' for nil:NilClass (PXELinux), foreman01: ERF12-9243 [ProxyAPI::ProxyException]: Unable to create default TFTP boot menu ([RestClient::BadRequest]: 400 Bad Request) for proxy http://foreman01:8443/tftp (PXELinux)

Apparently it doesn't build the pxe global default file since I haven't created the parameter. I guess I need to find a way to treat a non existent parameter as false.

Raboo:
Now it should work even if you don't have the disable-net-ifnames set.

lzap:
Unfortunately this will not work, you can't use `@host` or `host_param_true?` function in global context. I suggest to drop a comment only, sorry for confusion. We can't change that either, it's a breaking change and its the reason why this is opt-in only.

Raboo:
You are correct, added a comment instead, a bit long perhaps..

ekohl (approved):
A comment is a good start.

lzap:
Thanks a lot!

PR #351, opened 2017-02-07
Fixes networking preseed snippet

ares:
there's no such method allowed in the safe mode, this one seems to work fine


mmoll:
@ares which Foreman versions does this apply to?

ares:
The change was introduced in Foreman 1.7 - https://github.com/theforeman/foreman/commit/356b2e69f7c7d5dd4e4aac2cf43ac1c7a5930c95

mmoll:
merged, díky @ares!

PR #352, opened 2017-02-07
[Request for feedback] Improve ipv6 support

jovandeginste:
This allows interfaces with only an ipv6 address
I also change the device names to the one I 'want' (ethX)
Lastly, I add some items (hosts ip + expected dns) to /etc/hosts
for convenience.

I had this on my setup to match our 'legacy' installation workflow, decided to upstream it if it's wanted? I can obviously split this out into different commits/PRs, but feedback would be welcome first..


provisioning_templates/snippet/kickstart_networking_setup.erb
```diff
@@ -8,35 +8,43 @@ description: this will configure your host networking, it configures your primar
 %>
 <% subnet = @host.subnet -%>
 <% subnet6 = @host.subnet6 -%>
-<% dhcp = subnet.dhcp_boot_mode? -%>
+
+rm -v -f /etc/sysconfig/network-scripts/ifcfg-eth*
```
ekohl, line 12: In which case is this needed?

provisioning_templates/snippet/kickstart_networking_setup.erb
```diff
@@ -8,35 +8,43 @@ description: this will configure your host networking, it configures your primar
...
 <% if @host.respond_to?(:has_primary_interface?) %>
 <%# Foreman 1.7 - primary interface contained in @host %>
 # primary interface
-real=`ip -o link | grep <%= @host.mac -%> | awk '{print $2;}' | sed s/://`
 <% if @host.has_primary_interface? %>
+real=<%= @host.primary_interface.identifier %>
+
+echo "Writing /etc/sysconfig/network-scripts/ifcfg-$real"
 cat << EOF > /etc/sysconfig/network-scripts/ifcfg-$real
-BOOTPROTO="<%= dhcp ? 'dhcp' : 'none' -%>"
```
ekohl, line 19: Why did you remove this?

provisioning_templates/snippet/kickstart_networking_setup.erb
```diff
@@ -135,27 +142,34 @@ EOF
...
-
-# ifcfg files are ignored by NM if their name contains colons so we convert colons to underscore
+real=<%= interface.identifier %>
 <% if alias_type -%>
 sanitized_real=`echo $real | sed s/:/_/`
 <% else -%>
 sanitized_real=$real
 <% end -%>
 
+cat << EOF >> /etc/hosts
```
ekohl, line 152: I'm not sure everyone wants this. Maybe it should be optional behind a parameter?

ekohl (commented):
There are definitely some bugfixes in here that we should include so thanks for pointing those out. Maybe @ares could have a look as well.

jovandeginste:
I have forgotten to remove the "we don't use dhcp" part, I will fix this first.

provisioning_templates/snippet/kickstart_networking_setup.erb
jovandeginste: This is copied from a copy of a script from some time ago; it does not assume a clean system and runs on pre-installed systems to change their network.

provisioning_templates/snippet/kickstart_networking_setup.erb
jovandeginste: An issue I have with the network management, and I don't know yet how to correctly fix it:
Our mgmt network (usually eth0) is ipv6 only, but for practical reasons (one of those being that pxeboot/kickstart on ipv6 is still a not working for us) we do hand out ipv4 addresses via dhcp.

Basically: at boot time, a server gets an ipv4 address, during kickstart we set a fixed ipv6 address (and _no_ ipv4 address)

I could not model this behaviour in the then current options of Foreman (1.13.2), has this changed? Easiest "fix" was to put the v4 network for this subnet to dhcp, but always set "none" here.

provisioning_templates/snippet/kickstart_networking_setup.erb
jovandeginste: Yes, completely agree; might even be a separate snippet?

provisioning_templates/snippet/kickstart_networking_setup.erb
ekohl: @lzap maybe you have an idea on how to address this?

jovandeginste:
I split this into 3 different PR's (so far): #353, #354, #355.

I will open an issue for the dhcp-during-kickstart.

Should I close this PR?

ares:
It'd be awesome if we'd get https://github.com/theforeman/community-templates/pull/348 in first. Thanks for splitting into smaller PRs, I think this one can be closed now.

jovandeginste:
Closing now; #348 would be nice indeed :-)

PR #354, opened 2017-02-08
Use the device identifiers specified

jovandeginste:
When I specify an identifier in the Interfaces editor, I expect this to
be used as the device name on the server.

This change will remove configuration for auto-detected network
interfaces, and write the device names as specified in the editor.

It will also print the content for verification purposes.


jovandeginste:
I was actually (honestly) wondering why you were not doing this; anyone has an idea about this?

ares:
We didn't want to force user to specify names such as "eth0", "eth1" unless it was needed. Also we wanted to support el7 BIOSNAME identifiers which is "the new way" to name interfaces. With these names, it's hard for user to even specify correct name, since the number is determined based on motherboard slot index (IIRC).

We also do the `$real=ip -o link....` magic because of use case where you reprovision el6 host to el7, so we use MAC to find out what is the name on el7. This was mainly useful with baremetal provisioning with discovery that used to have discovery image based on el6, interfaces would keep old names even if we provisioned el7 there.

ekohl:
So where do we go from here? I think I agree with @ares that this isn't something we want to do by default. The good news is that we refactored a bit and this is now only in ```provisioning_templates/snippet/kickstart_ifcfg_get_identifier_names.erb``` so it should be easy to override.

jovandeginste:
I understand you want to use the new names; I tried but fail to see the use of those (I say this however as someone who kickstarted and installed servers from RH5 till now and likes stability/predictability during installation :-) )

Would some kind of a boolean "use identifier / use bios name" in the interface configuration form be a solution? Or am I really an outlier in this?

ekohl:
I was also considering a parameter and I think it makes sense.

ares:
I can imagine this being a snippet parameter which would be exposed for the kickstart template and passed to the snippet from there. So far no one except you has asked for it but if we keep the parameter defaulting to what we have right now I think that's fine.

Klaas-:
I think this could cause problems on vmware virtual machines, the device names are sometimes not the ones you'd expect them to be because they have a weird way of adressing their paravirtualized devices. If search for the device by mac and then use that identifier it works fine, if you want to set the names of devices like you suggest you'd need to also utilize udev I'd say.

Just in case someone is interested in the vmware issue - it works as designed for them ;) from the case I opened at vmware ages ago:

```
There's only 4 PCI Express root ports in a VM, each with 8 PCI "function"s.  lsisas gets the first, and the first three NICs get the last 3 -- all assigned to function 0.  After that the assignment cycles back to the first root port and starts using function 1 on the root ports.  So, the 4th NIC ends up on the first PCI Express port next to lsisas.
Example: 4 PCI->PCI express bridges devices ( 0.21.0, 0.22.0, 0.23.0, 0.24.0) and each has 8 functions (for example, the bridge device 21 has 8 functions such as 0.21.0, 0.21.1 to 0.21.7). Each PCI express devices will be pluged-in under these functions. 
 
0.21.0    0.22.0   0.23.0   0.24.0
 
------    ------   ------   ------
 
 1          2        3        4
 
 5          6        7        8
 
 9         10
 
So in PXE boot, the selection of boot order would be 1->5->9->2->6 and so on.
```

jovandeginste:
@Klaas- not sure what you mean.

We deploy any kind of servers (vmware vms on esx, hp blades, hp dl's, dell servers, moonshots) and since I am involved we tell Linux (RHEL/CentOS) to assign device 'eth0' to mac address 'xyz'. This has always worked. The only caveat is that you may need to clean up `/etc/udev/rules.d/70-persistent-net.rules` (delete existing entries) when you CHANGE this. During installation this is not an issue, nothing is there to clean up.

This has always been the most intuitive way to configure our servers: I tell the OS before even provisioning what the interface name will be in vlanX, or what interfaces will be members of bond0 (by mac address). I do not care in what order the physical or virtual interfaces are configured (or swapped if anything happens), all I need are the mac addresses. The only issue we ever have is when a network interface needs to be replaced (not really relevant for virtual machines). Then I need to change the mac address configured for that interface. But that is such an edge case scenario ...

Klaas-:
I think I misunderstood what you're trying to achieve. In that case you need to make the identifier a forced parameter for hosts, currently its not enforced to set an identifier and you can just let the system decide the ones it defaults to. That is what I meant in my previous comment - it produces weird results in vmware (if you have more than 4 paravirtual devices).

You could maybe check if @[...].identifier is set and then use that. But that could create issues if you mix setting it and not setting it (ie you use eth1 manually but system has already chosen that for another NIC)

jovandeginste:
@Klaas- I mitigate the double set by removing a bunch of files: https://github.com/theforeman/community-templates/pull/354/files#diff-5e9cf427d517ae8299a6963e3190ccfcR13

Klaas-:
no, I was talking about mixing setting identifier and not setting it within the same host - that could lead to problems if for example you say you set an interface to be eth1 but the os default for another interface, where you have not set an identifier, is also eth1. In general I think this is not something you can just change in the templates without changing foreman handling of network interfaces

lzap:
Hello, disabling naming conventions which are in RHEL is not what we want to do. Interface identifiers in Foreman are not very relevant for provisioning, we use MAC address where we can. There are some edge cases when interface identifier must be used, e.g. for VLAN provisoning as dracut only accepts them on kernel command line, but that's about it.

PR #355, opened 2017-02-08
Suppport interfaces with ipv6 but no ipv4 address

jovandeginste:
This change allows interfaces to have only an ipv6 address, while ipv4
is either blank (no v4 support on that interface) or dhcp.


mmoll:
I didn't test this, but it's looking correct. @ares? @timogoebel?

timogoebel (approved):
LGTM.

ares:
looks good, but shouldn't the same be applied for bonds?

I'd like to get the [refactoring PR](https://github.com/theforeman/community-templates/pull/348) merged first since that deduplicates this but if that won't happen anytime soon, I don't insist on waiting :-)

jovandeginste:
I added code for the bonds -- I also moved ipv6 out of the dhcp part here ...

mmoll:

PR #359, opened 2017-02-13
Fix nil access in kickstart_networking_setup

ananace:
This change fixes issues that might occur when trying to generate kickstart provisioning scripts on hosts without subnets specified.

The exact error this fixes is;
```
Failure parsing Kickstart default: The snippet 'kickstart_networking_setup' threw an error: undefined method 'gateway' for nil:NilClass.
```


dLobatog:
Thanks @ace13 !

ares:
Might have even fixed http://projects.theforeman.org/issues/18305, I'll rebase https://github.com/theforeman/community-templates/pull/348 to include it.

PR #360, opened 2017-02-15
Use the interface's ipv6 address

jovandeginste:


mmoll:

PR #361, opened 2017-02-17
Fixes #18509 - configure hostname in finish template

ares:


mmoll:
can this somehow get merged/unified into the fix_hosts snippet?

provisioning_templates/snippet/set_hostname.erb
```diff
@@ -0,0 +1,15 @@
+<%#
+name: set_hostname
+snippet: true
+model: ProvisioningTemplate
+kind: snippet
+-%>
+hostname <%= @hostname %>
```
ekohl: Should it check if ```hostnamectl``` should be used?

ekohl (commented):
I haven't seen a case where this is needed. Is this a situation where the user hasn't set up DNS?

ares:
> can this somehow get merged/unified into the fix_hosts snippet?

Good point @mmoll, do you think the @host.shortname is preferable over @host.name? That's the only difference. If we agree on one, I can make fix_hosts using set_hostname internally and call the same from kickstart_default_finish.erb

provisioning_templates/snippet/set_hostname.erb
ares: Would it be better to use os major/family foreman facts or rather something like `which hostnamectl`?

ekohl: I think I'd prefer ```which hostnamectl```

mmoll:
@ares I have no idea why, but on my rails console, I'm getting the FQDN for both variables.

ares:
> @ares I have no idea why, but on my rails console, I'm getting the FQDN for both variables.

does the host have a domain associated?

ares:
> I haven't seen a case where this is needed. Is this a situation where the user hasn't set up DNS?

When I have an image with the hostname hardcoded in it and use it for SSH based provisioning, I need modify it to what I set in foreman as a hostname. How do you set hostname in your env? I think host should know its own name without DNS.

mmoll:
> does the host have a domain associated?

Yes, it does... however, in general I think FQDN is always good and (if possible) the shortname can be added as second entry

ekohl:
> When I have an image with the hostname hardcoded in it and use it for SSH based provisioning, I need modify it to what I set in foreman as a hostname.

I didn't consider images and only used kickstarts. In that case Anaconda takes care of it. Thanks for pointing that out :)

ares:
I've updated the `set_hostname` snippet so it uses `hostnamectl` if it's available (it also sets /etc/hostname). I updated the fix_hosts snippet so it uses set_hostname snippet internally and instead of shortname it uses @host.name now. I could also remove it's direct using and replace it with `set_hostname` but I didn't want to change too much. Let me know if you want me to update 

* provisioning_templates/finish/alterator_default_finish.erb 
* provisioning_templates/user_data/kickstart_default_user_data.erb
* provisioning_templates/user_data/preseed_default_user_data.erb

so it uses set_hostname with variable directly. I think we should keep the snippet here for some time, people might rely on it in their custom templates...

@mmoll and @ekohl, I'd appreciate if guys could take another look.

provisioning_templates/snippet/set_hostname.erb
```diff
@@ -0,0 +1,20 @@
...
+  hostname <%= @hostname %>
+  # persist the change to survive  reboot
+  echo "<%= @hostname -%>" > /etc/hostname
+fi
+
+if grep '^127.0.0.1\s*<%= @hostname -%>$' /etc/hosts; then
+  echo "skipping /etc/hosts update, record is already present"
+else
+  echo "adding record to the first line of /etc/hosts"
+  sed -i '1s/^/127.0.0.1   <%= @hostname -%>\n/' /etc/hosts
```
mmoll, line 24: I think generating a new /etc/hosts would be better, especially on clud images you never know what's there. Also, isn't that "localhost" entry also needed?

GregSutcliffe:
This still looks relevant - I do something similar in my image templates :) @ares, are you planning to finish this?

provisioning_templates/snippet/set_hostname.erb
ares: @mmoll what if I prepared my image with custom hosts file with some specific records? sed should be safe enough... I'm not sure about the localhost, but I suppose it does not hurt to change it to
`sed -i '1s/^/127.0.0.1   <%= @hostname -%> localhost\n/' /etc/hosts`

would that be acceptable?

ares: also generating new hosts file means, I wouldn't be able to use the snippet twice within the template

mmoll: I don't know... at least Debian/Ubuntu sets `127.0.1.1 fqdn hostname` per default. I'd rather go for 127.0.0.1 for localhost and the primary IP for the fqdn and hostname...

ares:
@GregSutcliffe I updated the hosts file according to what @mmoll suggested, could you please take a look?

provisioning_templates/snippet/set_hostname.erb
```diff
@@ -0,0 +1,20 @@
...
+  hostname <%= @hostname %>
+  # persist the change to survive  reboot
+  echo "<%= @hostname -%>" > /etc/hostname
+fi
+
+if grep '^127.0.1.1\s*<%= @hostname -%>\s*<%= @shortname -%>$' /etc/hosts; then
+  echo "skipping /etc/hosts update, record is already present"
+else
+  echo "adding record to the first line of /etc/hosts"
+  sed -i '1s/^/127.0.1.1   <%= @hostname -%> <%= @shortname -%>\n/' /etc/hosts
```
GregSutcliffe: This is different to the old, it doesn't have the `localhost.*` entries on the line

GregSutcliffe:
One comment, but looks sensible. @mmoll does this look OK now?

provisioning_templates/snippet/set_hostname.erb
ares: Since we now rely on this being present from OS and with address 127.0.0.1. We nowadd separate record for 127.0.1.1.

ares: Maybe I could move it to second line (sed 2s/) but I don't think it would make a bit difference.

GregSutcliffe: Relying on the OS being correct is what got us into this situation in the first place. If we trusted the contents of `/etc/hosts` we wouldn't need to mess with this at all. Given that we're altering a snippet that may be in use (and in the user_data templates, where hosts could have all sorts of wonky base images), shouldn't we try to preserve behaviour?

ares: I don't think that's the original motivation (at least in my case). The image already contains /etc/hosts, I choose the FQDN when I do the provisioning, the OS can't know what I had in mind during provisioning. Therefore we to set the hostname as part of provisioning.

The question is what was the motivation for "fixing" the hosts file in fix_hosts.erb snippet. If someone knows or consider that localhost and ipv6 addresses always right, I'm happy to incorporate it here. The question is whether this should be part of "set_hostname" though, maybe it should remain in fix_hosts snippets then.

GregSutcliffe: Sure, I get that it's not the motivation for this PR, but you are removing functionality that was present before. I see your point about it not belonging in set_hostname though... but this patch does move from replacing the whole file to changing a single line. That may have consequences for some people who's basic /etc/hosts is incorrect.

I think it's probably an edge case, so I'd go for a merge & a note in the release notes?

ares: makes sense, so let's see if others have some insights... I agree to what you suggested

iNecas: 👍

mmoll:
I didn't test that, but it's looking OK visually. 👍

provisioning_templates/snippet/set_hostname.erb
```diff
@@ -0,0 +1,20 @@
...
+snippet: true
+model: ProvisioningTemplate
+kind: snippet
+-%>
+if [ `command -v hostnamectl` ]; then
+  hostnamectl set-hostname <%= @hostname %>
+else
+  hostname <%= @hostname %>
+  # persist the change to survive  reboot
+  echo "<%= @hostname -%>" > /etc/hostname
```
ekohl: This is correct for Debian and systemd-based systems. Do you want to do something for EL6 and older?

provisioning_templates/snippet/set_hostname.erb
```diff
@@ -0,0 +1,20 @@
...
+fi
+
+if grep '^127.0.1.1\s*<%= @hostname -%>\s*<%= @shortname -%>$' /etc/hosts; then
```
ekohl: The style is a bit inconsistent. Above it's ```[ `command` ]``` and here just ```grep```. I don't mind which style is used, but I like consistency :) Also ```grep -q```?

ekohl (changes requested):

mmoll:
@ares please rebase and have a look into @ekohl's comments, this still can get into 1.18 :)

ares:
rebased and updated, I hope it's ok for everybody now, @ekohl mind to revisit your review?

provisioning_templates/snippet/set_hostname.erb
```diff
@@ -0,0 +1,25 @@
...
+snippet: true
+model: ProvisioningTemplate
+kind: snippet
+-%>
+if command -v hostnamectl; then
+  hostnamectl set-hostname <%= @hostname %>
+else
+  hostname <%= @hostname %>
+# persist the change to survive reboot
+if [ -d "/etc/sysconfig" ]; then
```
ekohl, line 12: I'd test if `/etc/sysconfig/network` exists. I've seen `/etc/sysconfig` exist on some Debian systems because a package placed some files in that directory.

mmoll:
@ares what's the status here?

lzap:
Are you planning getting back to this? Good patch, I think `/etc/hosts` changes should be optional tho because that's really only needed when you don't have proper DNS setup. Foreman can provide DNS management and no changes in this file would be necessary usually.

ares:
Everytime I fix something I get a new request, I'm not sure if I can spend more time on this. I can rebase and update the path at some point I guess.

@lzap how would you like to make this optional? if you read one of the first comments, you'll find the motivation - https://github.com/theforeman/community-templates/pull/361#issuecomment-286050480 Note that this is changing the finish template.

lzap:
It is assumed that hosts managed by Foreman have proper DNS, therefore no changes of `/etc/hosts` is necessary. Some users tend to manage hosts file entries instead of DNS, then configuration management should be used. However, feel free to send a patch to Foreman templates to do this during provisioning, but only as opt-in via parameter. By default, we do not want to change what OS installers set.

File `/etc/resolv.conf` is managed by NetworkManager from CentOS 6, there is no need of changing it directly. Nameservers must be set in ifcfg setting (DNS1, DNS2) or interactively via NetworkManager CLI, TUI or applet.

ares:
Ok, I'll repeat the last time, this is not about properly working DNS in the infrastructure. If you do image based provisioning, you have no idea what the image has in /etc/hosts. That has higher priority than DNS. This would fix it. Setting FQDN in hosts is probably always a good idea. There was overall support for this, not sure what changed.

lzap:
Well, official Fedora/RH images have no `/etc/hosts` entries (localhost only), I suppose: https://github.com/weldr/lorax/blob/master/docs/fedora-openstack.ks

If you create your own image with garbage, it's not our job to fix this. We even ship a tool with RHEL and Fedora that is supposed to clean such images from the garbage!

http://libguestfs.org/virt-sysprep.1.html

Edit: Had to ask the maintainer if `/etc/hosts` is actually subject of the cleanup, however that does not make my attitude any different: https://github.com/libguestfs/libguestfs/issues/44 :-)

lzap:
I am fine with merging *opt-in* deployment of `/etc/hosts`. I'd prefer it to be a separate template and it could be rendered via `curl http://foreman/unattended/script > /etc/hosts`.

PR #362, opened 2017-02-18
move EPEL and Puppet repos/install fully into snippets

mmoll:


mmoll:
This is yet untested, @ares @domcleal @ekohl let me know, what you think.
Also, some conditionals for EL4 based OSes are removed, RHEL4 will be EOLed in six weeks.

ekohl:
Personally I prefer ```%repo``` and ```%package``` over snippets because you get a better progress indicator while installing and it feels cleaner. Sadly that doesn't work with RHEL due to registration and such so this may be cleaner and unified.

mmoll:
of course this does not work with SUSE... However, I'll split out some smaller parts first and come back here in a few days.

ares:
I agree with @ekohl, I think the unification and deduplication justifies the change.

mmoll:
this is ready to be reviewed now.

PR #363, opened 2017-02-20
drop support for EL4 based OSes

mmoll:


mmoll:
According to https://access.redhat.com/support/policy/updates/errata RHEL 4 reaches "End of Extended Life-cycle Support" on March 31, 2017

ekohl (approved):

lzap:
I think this is incorrect, we should really change `os_major > 4` to `os_major > 5` because now we changed the behavior of the templates a lot. @mmoll ?

mmoll:
5, 6 and 7 are >4, so this would always be true? (edit, 5 is of course not >5)

lzap:
Yeah I get it, sure.

PR #364, opened 2017-02-20
Refs #18576 - updates for OpenSUSE Leap

mmoll:


mmoll:
42.1 is the oldest relase currently supported (including Evergreen LTS), according to https://en.opensuse.org/Lifetime

PR #365, opened 2017-02-20
add AutoYaST user-data

mmoll:
OpenSUSE does provide cloud-images and I'm able to use these with libvirt with this user-data template. The SLES part was tested with a SLES 12 SP1 based JeOS (https://www.suse.com/products/server/jeos/) cloud-image.

### DIFF
--- a/provisioning_templates/user_data/autoyast_default_user_data.erb
+++ b/provisioning_templates/user_data/autoyast_default_user_data.erb
@@ -1,0 +1,38 @@
+<%#
+kind: user_data
+name: AutoYaST default user data
+oses:
+- OpenSUSE
+- SLES
+-%>
+<%
+  # safemode renderer does not support unary negation
+  pm_set = @host.puppetmaster.empty? ? false : true
+  puppet_enabled = pm_set || @host.param_true?('force-puppet')
+  salt_enabled = @host.params['salt_master'] ? true : false
+-%>
+#!/bin/bash
+
+<%# Cloud instances frequently have incorrect hosts data %>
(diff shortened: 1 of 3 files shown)


provisioning_templates/user_data/autoyast_default_user_data.erb
```diff
@@ -0,0 +1,38 @@
...
+<% if puppet_enabled %>
+<% if @host.param_true?('enable-puppetlabs-pc1-repo') -%>
+<%= snippet 'puppetlabs_repo' %>
+<% end -%>
+<%= snippet 'puppet_setup' %>
+<% end -%>
+
+<% if salt_enabled %>
+<%= snippet 'saltstack_setup' %>
+<% end -%>
```
ares, line 35: no :heart: for chef? I think it could be the same as in kickstart template https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/provision/kickstart_default.erb#L203-L205

mmoll: AFAICT, nobody ever tested this. If you want to test this with OpenSUSE 42.2 and SLES 12 SP2 (ISO is downloadable after registration and usable as install medium with loopback mounting) and confirm this works, you are more than welcome to add it. However, given all the strange differences I encountered over the years with SUSE based OSes, it wouldn't be surprising if something does not work of of the box.

ares: Fair enough, no need to add it :-) thanks for instructions, maybe one day I will

mmoll:
more comments? :)

PR #366, opened 2017-02-23
preseed: use linux/initrdefi for EFI

mmoll:


mmoll:
@sathieu @lzap could you have a look?

sathieu:
@mmoll linuxefi doesn't work with Debian Kernel. I don't know why, probably because it's not signed.

I think that this should at least be configurable.

Side note: How can we get the boot type from the template?

mmoll:
@sathieu interesting, somebody on IRC had a non-working configuration a few days ago, and that change fixed it. Does it wotk on EFI without the "efi" in the template for you? Is there a difference between Debian and Ubuntu regarding the signage?

sathieu:
Debian jessie kernel is not signed. Debian stretch will (see https://lwn.net/Articles/703001/). Ubuntu is signed since 12.10 or so. I propose using this info as conditional. 

Also note that I've not tested Debian stretch on UEFI yet (I've only tested jessie).

mmoll:
@sathieu I see... would something like the updated PR work?

sathieu:
Looks good to me. Maybe rename efi_string to efi_suffix?

mmoll:
good point, changed.

lzap:
I am ok with the change, thanks.

PR #367, opened 2017-02-23
enable/start puppet explicitly also on Debian 9

mmoll:


mmoll:
see https://anonscm.debian.org/cgit/pkg-puppet/puppet.git/plain/debian/NEWS

mmoll:
@sathieu could you also have a look here?

mmoll:
also, moving the enable/start part after the puppet run to get the cert signed.

sathieu:
Great. Thanks.

PR #369, opened 2017-03-02
Added more options for chainbooting

lzap:
We have more reports from users unable to boot RHEV instances with multiple
disks attached via LOCALBOOT 0 option, the workaround is to use chain
mechanism, but there are also other LOCALBOOT options avaiable which could work
as well.


provisioning_templates/snippet/pxelinux_chainload.erb
```diff
@@ -3,12 +3,25 @@ kind: snippet
...
+LABEL local_chain_hd0
+  MENU LABEL Chainload the first hard drive (hd0)
   COM32 chain.c32
   APPEND hd0
 
+LABEL local_chain_hd1
+  MENU LABEL Chainload the second hard drive (hd1)
+  COM32 chain.c32
+  APPEND hd1
+
```
mmoll: nitpick: extra newline

mmoll:
I'm OK with the changes, but as a user the differences between the options would be mostly unclear to me, maybe some other wording like "Default local boot" and "Boot from primary hard drive" for the first two options would be good.

lzap:
Done!

mmoll:
merged, díky @lzap!

PR #370, opened 2017-03-02
update Atomic Kickstart template

mmoll:
import two changes from Katello:
- add network configuration
- add remote_execution_ssh_keys snippet


mmoll:
@dLobatog  could you have a look?

dLobatog:
These changes have been tested in katello for a while - merging. thanks @mmoll ~!

PR #371, opened 2017-03-07
Add a create_users snippet

ekohl:
Goes along with https://github.com/theforeman/foreman/pull/4290


ares:
would you mind using the same indentation we have e.g. in [kickstart_ifcfg_generic_interface.erb](https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb) ?

I think this snippet has 3 levels of indentation and for me is hard to read, basically

- start every line with either <%- or <%=
- nest inside the erb tag, e.g. `<%-     index = 0 -%>

another thoughts

I'd avoid `if` condition with index, what's bad on having index there even for single key? and why do we need index anyway? if we need to differentiate, wouldn't it be better to use `key.id` instead?

provisioning_templates/snippet/create_users.erb
```diff
@@ -0,0 +1,25 @@
+<%#
+kind: snippet
+name: create_users
+-%>
+# Create all users
+<% [@host.owner].flatten.each do |user| -%>
+<% username = user.login -%>
+useradd <%= username %>
+<% if user.ssh_authorized_keys && user.ssh_authorized_keys.any? -%>
+mkdir <%= username %>/.ssh
```
ares: SSH can be picky about `.ssh` and the key file permissions, should we enforce them here? Also would it be safer to run `mkdir -p`?

ares: I see you're doing it at the end, as long as this is run under root with reasonable umask that's fine, if we want to reuse that later from e.g. REX we should revisit this since this leaves a windows where someone could inject own authorized key

ekohl:
> I'd avoid if condition with index, what's bad on having index there even for single key? and why do we need index anyway? if we need to differentiate, wouldn't it be better to use key.id instead?

This matches what we do in https://github.com/ekohl/puppet-foreman_simple_user. Puppet requires unique key comments so this matches it.

provisioning_templates/snippet/create_users.erb
ekohl: Good point

timogoebel:
> if we need to differentiate, wouldn't it be better to use key.id instead?

I'd try to avoid leaking foreman internal database ids here. I believe a simple index is enough.

ekohl:
We also tried ```.each_with_index``` but that's not allowed in safemode.

provisioning_templates/snippet/create_users.erb
```diff
@@ -0,0 +1,25 @@
+<%#
+name: create_users
+kind: snippet
+snippet: true
+-%>
+<%- users = @host.owner_type == 'Usergroup' ? @host.owner.all_users : [@host.owner] -%>
+<%- users.each do |user| -%>
+<%=   "useradd #{user.login}" %>
+<%-   if user.ssh_authorized_keys && user.ssh_authorized_keys.any? -%>
```
ares: `user.ssh_authorized_keys` will always evaluate to true since it's either empty array or array of keys. If that check was meant to verify the user object responds to that method, `respond_to?` should be used but since templates are versioned I don't think that's needed

ares:
OK, fair enough, thanks for indentation, I find this much more readable. I haven't tested it myself yet, but :+1: Left one comment about minor thing.

provisioning_templates/snippet/create_users.erb
```diff
@@ -0,0 +1,25 @@
...
+<%=     "mkdir -p -m 0700 #{user.login}/.ssh" %>
```
mmoll: there's a tilde missing

provisioning_templates/snippet/create_users.erb
```diff
@@ -0,0 +1,25 @@
...
+<%=     "cat >> ~#{user.login}/.ssh/authorized_keys <<EOF" %>
+<%-     index = 0 -%>
+<%-     user.ssh_keys.each do |key| -%>
+<%-       if index == 0 -%>
+<%=        "#{key.type} #{key.key} #{key.comment}" %>
+<%-       else -%>
+<%=        "#{key.type} #{key.key} #{key.comment} - #{index}" %>
```
mmoll, line 17: I'm not getting that line... should be index be subtracted or printed? if the later, why?

provisioning_templates/snippet/create_users.erb
ekohl: Sharp, will fix.

provisioning_templates/snippet/create_users.erb
ekohl: This makes the names unique. This is done to match https://github.com/ekohl/puppet-foreman_simple_user

provisioning_templates/snippet/create_users.erb
ekohl: I did end up using ```respond_to?``` just to be safe.

provisioning_templates/snippet/create_users.erb
ekohl: Fixed

provisioning_templates/snippet/create_users.erb
```diff
@@ -0,0 +1,25 @@
+<%#
+name: create_users
+kind: snippet
+snippet: true
+-%>
+<%- users = @host.owner_type == 'Usergroup' ? @host.owner.all_users : [@host.owner] -%>
+<%- users.each do |user| -%>
+<%=   "useradd #{user.login}" %>
+<%-   if user.respond_to?(ssh_authorized_keys) && user.ssh_authorized_keys.any? -%>
```
mmoll: this needs to be `respond_to?(:ssh_authorized_keys)` with the colon

provisioning_templates/snippet/create_users.erb
```diff
@@ -0,0 +1,25 @@
...
+<%-       if index == 0 -%>
+<%=        "#{key.type} #{key.key} #{key.comment}" %>
+<%-       else -%>
+<%=        "#{key.type} #{key.key} #{key.comment} - #{index}" %>
+<%-       end -%>
+<%-       index += 1 -%>
+<%-     end -%>
+<%=     "EOF" %>
+<%=     "chown -R #{user.login}:#{user.login} ~#{user.login}/.ssh" %>
+<%=     "chmod go= ~#{user.login}/.ssh" %>
```
mmoll: i think there should also be a `chmod 0600 ~{user.login}/.ssh/authorized_keys`

ekohl: Actually it should be -R

mmoll:
I didn't yet test this, but it's looking fine now and it's only a snippet which is not connected in any other templates. merging.

PR #372, opened 2017-03-08
Remove a redundant grep

ekohl:


ares:
Good catch, works for me :+1:

ekohl:
I tried to remove sed as well, but that didn't make it any better.

mmoll:
@ekohl is that applyable to the preseed snippet as well?

ekohl:
@mmoll it is so I've updated it.

mmoll:
merged, bedankt @ekohl!

PR #373, opened 2017-03-08
Allow ipv6-only interfaces

jovandeginste:
A generic interface can have an ipv6 address without having an ipv4
address


ekohl (approved):
Untested, but it looks correct to me.

mmoll:

PR #374, opened 2017-03-08
Extra parameters when interface is named vlanX

jovandeginste:
When you have a vlan with id `5` on top of a device `bond0`, you may call the vlan
interface either `bond0.5` or simply `vlan5`. The latter case is now
supported via this PR, which adds the following paramters:

```bash
VLAN_NAME_TYPE=VLAN_PLUS_VID_NO_PAD
PHYSDEV=bond0        # whatever the `attached_to` is
```

You interface identifier should be `vlan5` in this example.

For more information:
https://www.tummy.com/blogs/2006/11/30/configuring-vlans-under-fedorarhelcentos/


ares:
I guess this is safe as it only applies if the identifier is specified this way, just note that when we parse facts from puppet we rely on identifier value to detect the interface type, if the host would be unknown to foreman, "vlan5" wouldn't be probably identified as vlan interface and the tag 5 would not be parsed either. Maybe the parsing in foreman could be updated so ^vlan\d+ would be identified as vlan type of interface too.

jovandeginste:
@ares I'm certainly pro correctly identifying unknown hosts in Foreman and am certainly willing to provide code and feedback for this.

I haven't looked at that part yet - at this point I'm still working on provisioning via Foreman, after having provided Foreman with the relevant configuration.

ares:
Sounds good. I haven't tested but :+1: for the concept

provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
```diff
@@ -42,6 +42,10 @@ DEFROUTE=<%= primary %>
 <%- end -%>
 <%- if @interface.virtual? && ((!@subnet.nil? && @subnet.has_vlanid?) || @interface.vlanid.present?) -%>
 <%=   "VLAN=yes" %>
+<%-   if @interface.identifier.match(/^vlan/) -%>
```
ares: would it make sense to check for `/vlan\d+/`? just in case someone used vlan5.10 (not sure what would that mean :-)

mmoll:
@jovandeginste what's the status here?

jovandeginste:
I agree with @ares, found another addition needed. Will update the PR

jovandeginste:
@mmoll @ares can any of you confirm that the changes in `provisioning_templates/snippet/kickstart_networking_setup.erb` do what I expect it to do? I didn't upgrade my setup to the split-snippet yet, so everything is still a single erb here...

mmoll:
@jovandeginste Did you upgrade in the meanwhile? ;)

jovandeginste:
Is there an easy way to do this? I'm pretty sure I didn't touch any of the built-in templates...

mmoll:
If you're on Foreman 1.15, foreman_templates is probably the easiest way to sync in your changes.

jovandeginste:
Meanwhile I managed to upgrade and convert our provisioning scripts to use the new networking scripts (modulo all the PR's I submitted ;-))

Everything seems to work from what I can tell here - we successfully deployed a number of different hardware and network configurations.

mmoll:

PR #377, opened 2017-03-24
Add support for static networking in the autoyast templates.

hboetes:
I noticed the Suse based installs didn't support static networking, unlike debian and redhat based systems, and I needed it for my project.

Do note I basically applied the same diff twice to two different templates.

### DIFF
--- a/provisioning_templates/provision/autoyast_default.erb
+++ b/provisioning_templates/provision/autoyast_default.erb
@@ -4,13 +4,16 @@ kind: provision
 name: AutoYaST default
 oses:
 - OpenSUSE
-%>
+-%>
 <%
   # safemode renderer does not support unary negation
   pm_set = @host.puppetmaster.empty? ? false : true
   puppet_enabled = pm_set || @host.param_true?('force-puppet')
   salt_enabled = @host.params['salt_master'] ? true : false
-%>
+
+  primary_interface_identifier = @host.primary_interface.identifier.blank? ? 'eth0' : @host.primary_interface.identifier
+  primary_interface_subnet = @host.primary_interface.subnet
+-%>
 <!DOCTYPE profile>
 <profile xmlns="http://www.suse.com/1.0/yast2ns" xmlns:config="http://www.suse.com/1.0/configns">
   <general>
@@ -22,7 +25,23 @@ oses:
   <networking>
     <dns>
       <hostname><%= @host.name %></hostname>
+<% unless primary_interface_subnet.dhcp_boot_mode? -%>
+      <dhcp_hostname config:type="boolean">false</dhcp_hostname>
+      <dhcp_resolv config:type="boolean">false</dhcp_resolv>
+<% if @host.domain -%>
+      <domain><%= @host.domain -%></domain>
+<% end -%>
+<% if primary_interface_subnet.dns_primary.present? -%>
+      <nameservers config:type="list">
+        <nameserver><%= primary_interface_subnet.dns_primary -%></nameserver>
+<% if primary_interface_subnet.dns_secondary.present? -%>
+        <nameserver><%= primary_interface_subnet.dns_secondary -%></nameserver>
+<% end -%>
+      </nameservers>
+<% end -%>
+<% else -%>
       <resolv_conf_policy>auto</resolv_conf_policy>
+<% end -%>
 <% if @host.domain -%>
       <searchlist config:type="list">
         <search><%= @host.domain %></search>
@@ -30,13 +49,42 @@ oses:
 <% end -%>
     </dns>
     <interfaces config:type="list">
+<%
+  @host.managed_interfaces.each do |interface|
+    next if !interface.managed? || interface.subnet.nil? || interface.ip.nil?
+    dhcp = interface.subnet.dhcp_boot_mode?
+-%>
+<% if dhcp -%>
       <interface>
         <bootproto>dhcp</bootproto>
-        <device>eth0</device>
+        <device><%= interface.primary ? primary_interface_identifier : interface.identifier -%></device>
         <startmode>auto</startmode>
         <usercontrol>no</usercontrol>
       </interface>
+<% else -%>
+      <interface>
+        <bootproto>static</bootproto>
+        <device><%= interface.primary ? primary_interface_identifier : interface.identifier -%></device>
+        <ipaddr><%= interface.ip -%></ipaddr>
+        <netmask><%= interface.subnet.mask -%></netmask>
+        <startmode>onboot</startmode>
+        <usercontrol>no</usercontrol>
+      </interface>
+<% end -%>
+<% end -%>
     </interfaces>
+<% unless primary_interface_subnet.dhcp_boot_mode? -%>
+    <routing>
+      <routes config:type="list">
+        <route>
+          <destination>default</destination>
+          <device><%= primary_interface_identifier -%></device>
+          <gateway><%= primary_interface_subnet.gateway -%></gateway>
+          <netmask><%= primary_interface_subnet.mask -%></netmask>
+        </route>
+      </routes>
+    </routing>
+<% end -%>
   </networking>
   <ntp-client>
     <configure_dhcp config:type="boolean">false</configure_dhcp>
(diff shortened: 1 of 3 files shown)


provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,54 @@ oses:
   <networking>
     <dns>
       <hostname><%= @host.name %></hostname>
+      <% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
+      <dhcp_hostname config:type="boolean">false</dhcp_hostname>
+      <dhcp_resolv config:type="boolean">false</dhcp_resolv>
+      <domain><%= @host.domain %></domain>
+      <nameservers config:type="list">
+        <nameserver><%= @host.subnet.dns_primary %></nameserver>
+        <nameserver><%= @host.subnet.dns_secondary %></nameserver>
```
Klaas-: I think we should check for presence of those nameservers, at least the secondary one - check the other templates: something like this:
<% if @host.subnet.dns_secondary.present? -%>
        <nameserver><%= @host.subnet.dns_secondary %></nameserver>
<% end -%>

provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,54 @@ oses:
...
+    <% end -%>
     </dns>
     <interfaces config:type="list">
       <interface>
-        <bootproto>dhcp</bootproto>
         <device>eth0</device>
+        <ethtool_options>autoneg on</ethtool_options>
+        <% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
+          <bootproto>static</bootproto>
+          <broadcast></broadcast>
```
Klaas-: I don't think you'll need this line

provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,57 @@ oses:
...
         <search><%= @host.domain %></search>
       </searchlist>
-<% end -%>
+    <% end -%>
     </dns>
     <interfaces config:type="list">
       <interface>
-        <bootproto>dhcp</bootproto>
         <device>eth0</device>
+        <ethtool_options>autoneg on</ethtool_options>
```
mmoll: is this line really needed?

provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,57 @@ oses:
...
         <usercontrol>no</usercontrol>
       </interface>
     </interfaces>
+    <% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
+    <routing>
+    <ip_forward config:type="boolean">false</ip_forward>
+    <routes config:type="list">
+      <route>
+        <destination>default</destination>
+        <device>eth0</device>
```
mmoll: I guess (but we do have the same problem for DHCP also, anyway...) that's not always correct with persistent device naming in newer versions?

provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,57 @@ oses:
...
+        <gateway><%= @host.subnet.gateway %></gateway>
+        <netmask>-</netmask>
```
mmoll: what does the dash mean?

provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,57 @@ oses:
...
-<% end -%>
+    <% end -%>
     </dns>
     <interfaces config:type="list">
       <interface>
-        <bootproto>dhcp</bootproto>
         <device>eth0</device>
+        <ethtool_options>autoneg on</ethtool_options>
+        <% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
+          <bootproto>static</bootproto>
```
ekohl: The XML indenting will look weird here.

provisioning_templates/provision/autoyast_default.erb
```diff
@@ -22,21 +22,57 @@ oses:
...
       </searchlist>
-<% end -%>
+    <% end -%>
     </dns>
     <interfaces config:type="list">
       <interface>
-        <bootproto>dhcp</bootproto>
         <device>eth0</device>
+        <ethtool_options>autoneg on</ethtool_options>
+        <% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
```
ekohl: If it responds to boot mode but boot mode is set to DHCP then you will still end up with static networking.

ekohl (changes requested):

provisioning_templates/provision/autoyast_default.erb
hboetes: I c&p this line from the kickstart template, what do you propose I should use?

provisioning_templates/provision/autoyast_default.erb
hboetes: I don't know, but I find it all over the docs. https://www.suse.com/documentation/sles-12/book_autoyast/data/createprofile_network.html

provisioning_templates/provision/autoyast_default.erb
hboetes: That depends on the default configuration in autoyast. autoneg is what most people will want to have.

mmoll: Isn't autoneg the default? So this line can get ommited?

provisioning_templates/provision/autoyast_default.erb
mmoll: ah, I see 👍

provisioning_templates/provision/autoyast_default.erb
hboetes: Fixed in my latest commit which supports multiple interfaces

provisioning_templates/provision/autoyast_default.erb
hboetes: I noticed as well. What can I do about it?

mmoll:
Regarding `dhcp_boot_mode`, the `respond_to?` is only for identifying if the Foreman in question does support the distinction. As the templates are now versioned, this can/should be ommited.

Regarding the distinction between DHCP and static mode, here is the simple code used in the preseed networking snippet:
```
<% @host.managed_interfaces.each do |interface| -%>
  <% next if !interface.managed? || interface.subnet.nil? || interface.primary -%>
  <% subnet = interface.subnet -%>
  <% dhcp = subnet.nil? ? false : subnet.dhcp_boot_mode? -%>
  <% if dhcp -%>
    [do dhcp things here]
  <% else -%>
    [do static things here]
  <% end -%>
<% end -%>
```

provisioning_templates/provision/autoyast_default.erb
mmoll: indenting it correctly? ;)

provisioning_templates/provision/autoyast_default.erb
hboetes: ok, consider it gone.

mmoll:
Again, `<% if @host.subnet.respond_to?(:dhcp_boot_mode?) -%>` does _not_ check if that subnet is in DHCP or static mode.

You copypasted some code from the kickstart snippet, but in your the current code there's no usage of the variables `virtual`, `vlan`, `alias-type` and even `dhcp` of the respective subnet. Please double-check your code and also make sure, DHCP installs work as intended.

mmoll (changes requested):
this needs to get into a working state first

hboetes:
If you guys want me to join all these commits, I'll gladly oblige.

Also, I'd like to note that travis-ci only tests kickstart files.

mmoll:
@hboetes thanks for following up on this, looking much better now.

I squashed all your commit together into one and then added one commit with some refactorings.

This is still untested and I don't know when I will have time to test this with real machines, in the meanwhile I'd ask @ares to have a look. 👀

Klaas-:
I think there is another issue here we should address: the device identifiers are not enforced by foreman at the moment, if they aren't set the network interfaces won't be configured  properly (at least if I didn't mess up my test case :) ). I'm not sure how one would work around that - could we just set a device name based on a loop-counter or can we do it like the rhel networking snippet does - set no device identifiers and use the mac to identify the device?

mmoll:
I did think about that, but at the end this is already an improvement over the current situation, where eth0 is also hardcoded. I hope to test this in the next days and merge it, further improvements shoudl go into further PRs.

mmoll:
merged, thanks @hboetes!

PR #379, opened 2017-03-27
Add inst.ks.sendmac option

lzap:
For more info:

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Installation_Guide/chap-anaconda-boot-options.html

when you scroll down to 20.1.1. Deprecated and Removed Boot Options.


ekohl (approved):
I don't think these hurt, but does Foreman actually use these HTTP headers?

mmoll:
What ewoud says. But merging for 1.15

lzap:
It does.

PR #382, opened 2017-03-30
validate autoyast XML

mmoll:


mmoll:
@hboetes Have a look here, this is not perfect, as I didn't find a XSD to validate the autoyast result against, but still better than nothing.

PR #384, opened 2017-04-15
Add proxy support for suse linux.

hboetes:
The big advantage is that all remote files are being installed from a caching proxy, including rpms from puppetlabs, making installs a lot faster, especially on slower connections.

### DIFF
--- a/provisioning_templates/PXELinux/autoyast_default_pxelinux.erb
+++ b/provisioning_templates/PXELinux/autoyast_default_pxelinux.erb
@@ -5,8 +5,14 @@ oses:
 - SLES
 - OpenSUSE
 %>
+<%
+  http_proxy_enabled = @host.params['http-proxy'] ? true : false
+  http_proxy = @host.params['http-proxy']
+  http_proxy_port = @host.params['http-proxy-port']
+-%>
+
 DEFAULT linux
 
 LABEL linux
     KERNEL <%= @kernel %>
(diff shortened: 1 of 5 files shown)


mmoll:
Please rebase on current develop, so that only one commit is left in this PR.

I did try something similar with apt-cacher-ng and did not implement it at the end, because I couldn't manage that only requests for repositories are proxied... I didn't try it the last two years or so, maybe this requires only little effort now.

mmoll:
Also, I'd recommend to use a new branch on your fork for each PR.

provisioning_templates/PXELinux/autoyast_default_pxelinux.erb
```diff
@@ -5,8 +5,14 @@ oses:
 - SLES
 - OpenSUSE
 %>
+<%
+  http_proxy_enabled = @host.params['http-proxy'] ? true : false
```
ekohl, line 9: Are empty strings true or false? I forgot.

ekohl (commented):
@mmoll there are also situations where you want everything to be proxied. Cases where you might not have direct internet access otherwise.

provisioning_templates/PXELinux/autoyast_default_pxelinux.erb
hboetes: @mmol yeah, it looks a bit messy. I’ll figure it out.
@ekohl good question, will test on tuesday.
@mmoll The nice thing about this solution it uses the proxy only during installation. I used about the same code in the provisioning template, setting up `<proxy>…</proxy>` and that didn’t do anything. But after installation I can use puppet to set up the system like I want.

PR #385, opened 2017-04-15
Add support for a proxy during provisioning for autoyast

hboetes:
ok, after a bit of fighting with git I got a clean commit, @mmoll 
As mentioned before, I still have to test what happens if the proxy is unset. That will happen on tuesday.


hboetes:
As promised, I just checked what happens when `http-proxy` is undefined and the proxy string is not included in the output, as desired.

ekohl (approved):

mmoll:
@hboetes I didn't test this, but I guess it'll be OK... I'd rather use something like (untested pseuo-code, but you'll get the idea):
```
http_proxy_string = @host.params['http-proxy'] ? "proxy=<%= @host.params['http-proxy'] %>:<%= @host.params['http-proxy-port'] %>" : ''

[...]

  APPEND initrd=<%= @initrd %> ramdisk_size=65536 install=<%= @mediapath %> autoyast=<%= foreman_url('provision') %> textmode=1 <%= http_proxy_string -%>
```

hboetes:
I was looking for that construct but I couldn’t find it in the docs nor an example. Thanks for the heads up, will change the commit and test it soonish.

hboetes:
Here is the new and improved pull request: https://github.com/theforeman/community-templates/pull/388

PR #388, opened 2017-05-15
Add support for a proxy for SLES installs

hboetes:
This simple patch enables using a proxy during SLES autoyast installs. This is the second version of this pull request, which uses a ternary expression as requested by @mmoll .

### DIFF
--- a/provisioning_templates/PXELinux/autoyast_default_pxelinux.erb
+++ b/provisioning_templates/PXELinux/autoyast_default_pxelinux.erb
@@ -5,7 +5,7 @@ oses:
 - SLES
 - OpenSUSE
 %>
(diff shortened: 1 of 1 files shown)


mmoll:
I think there should be a "http://" before the string (https://www.suse.com/documentation/sled-12/book_sle_deployment/data/sec_i_yast2_startup.html -> "Using a Proxy During the Installation"). Also, the OpenSUSE template should also be updated, looking good otherwise.

Please update your current branch of this PR instead of opening new ones.

hboetes:
Hi @mmoll 

- I added the http:// part to the url, good catch.
- I would prefer to keep using the same branch but I somehow had some problems caused by the previous branch pull request and decided to start from scratch. 

Sincerely.

mmoll:
merged, thanks @hboetes!

PR #389, opened 2017-05-18
WIP - Rework Container Linux (CoreOS)

johscheuer:
Hi,

since Container Linux had many changes how to provision it I started rework of the templates here (to make them up to date). Some information about the changes are here: https://github.com/johscheuer/theforeman-coreos-kubernetes/issues/11#issuecomment-302329543

The main changes for this template are:

- Use [ignition](https://coreos.com/ignition/docs/latest/) instead of deprecated cloud-config
- We don't need two templates for CoreOS anymore, since we can pass the ignition file directly to the boot process -> https://coreos.com/blog/introducing-ignition.html the part `PXE & bare metal`
- ignition is supported since CoreOS `1010.1.0` and is the preferred/more reliable provision tool
- cloud-config will probably be deprecated soon
 
Currently this PR doesn't contain all changes (WIP), the following todo's are open:

- [ ] translate cloud-config to ignition file
- [ ] testing


GregSutcliffe:
@johscheuer still working on this?

SndR85:
This isn't going to work without the transpiler between CoreOS and the provisioning script. Who is going to translate the yaml to Ignition json?

I think Foreman should be able to render the ignition json with the input of yaml (human friendly). See also https://github.com/coreos/container-linux-config-transpiler. For a workaround I think a kind of proxy (which translates the yaml to Ignition, with the transpiler) will also do.

Alternatively you could paste the Ignition in the provisioning template, but it is hard to edit, as it doesn't support newlines. I had done it this way, by building the yaml version, then translate it to json. But the ERB tags were lost, so I had to recover those to have those working again. It wasn't very friendly. That's why I think Foreman (or some proxy) should translate after the ERB parser.

I have created quick a tool which could do such job: https://github.com/scornelissen85/foreman-transpiler. This proxy will translate the yaml to json with the transpiler tool

provisioning_templates/provision/coreos_provision.erb
```diff
@@ -1,44 +1,109 @@
...
+  else
+    dhcp = !@static
+  end
+  unless dhcp
+    units += 1
+  end
+-%>
+<% if units > 0 -%>
+      coreos:
+<% if @host.params['reboot-strategy'] -%>
```
lzap, line 25: Please replace all `@host.params` with `host_param()` as in https://github.com/theforeman/community-templates/pull/484

johscheuer:
Currently I don't have time to finish this. Anyone who is interested in this can finish this. I will ask some friends if they find some time to finish this.

PR #390, opened 2017-05-19
Added missing finish options

lzap:
I think these guys were missing these options, this is untested but I was grepping our codebase for finish statements and these came up.


ekohl:
Does the foreman show a deprecation for unspecified ```foreman_url```? Personally I think it should because it's ambiguous and I like this change.

mmoll:
I think nobody uses these nowadays, but this is correct.

mmoll:
merged, díky @lzap!

lzap:
@ekohl we can even simply error out, deprecation warning first, yeah.

PR #392, opened 2017-05-30
fix authorized_keys in create_users snippet

zeromind:
use .ssh_key instead of .key, as we already use .type
and .key contains both .type and .ssh_key

### DIFF
--- a/provisioning_templates/snippet/create_users.erb
+++ b/provisioning_templates/snippet/create_users.erb
@@ -12,9 +12,9 @@ snippet: true
 <%-     index = 0 -%>
 <%-     user.ssh_keys.each do |key| -%>
 <%-       if index == 0 -%>
-<%=        "#{key.type} #{key.key} #{key.comment}" %>
+<%=        "#{key.type} #{key.ssh_key} #{key.comment}" %>
 <%-       else -%>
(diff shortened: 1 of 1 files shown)


mmoll:
@ekohl could you take a look here?

ekohl:
I think https://github.com/theforeman/foreman/blob/cf4687537819e1862d530049a7f29d0e86ba16fe/app/models/ssh_key.rb#L47-L53 is the relevant part and that would mean this shouldn't be needed. @timogoebel?

timogoebel:
@ekohl: Iirc `to_export_hash` is only relevant for the ENC data. The template code uses the code directly.

To get the correct string, either use `"#{key.type} #{key.ssh_key} #{key.comment}"` or `"#{key. to_export}"`.

ekohl:
Right, so this change is correct.

ekohl (approved):

timogoebel (approved):

ekohl:
Thanks @zeromind!

ares:
This should probably be cherry-picked to 1.15-stable and synced to Foreman 1.15.2, otherwise ssh keys can't be deployed.

ares:
sorry for the noise, it was already CPed to 1.15-stable as 58de81dc54e22a297f39ddaca79341d15d7dcda0

PR #393, opened 2017-06-06
Include grub.cfg-01-$mac files in global template

lemrouch:
because GRUB2 on Debian does not include them automatically.


lzap (approved):
I haven't tested but sure, thanks.

PR #395, opened 2017-06-08
Primary is not provision

hboetes:
Without this patch the Debian installer will try to install from the primary interface, which in my case not configured for such purposes.

I have a caching proxy mirror setup which hosts all packages, it also is the management access for foreman, puppet, monitoring, etc.

I do not want my hosts to access the internet to get packages and updates, this would cause a massive bandwidth overhead.

### DIFF
--- a/provisioning_templates/provision/preseed_default.erb
+++ b/provisioning_templates/provision/preseed_default.erb
@@ -21,7 +21,7 @@ d-i keyboard-configuration/xkb-keymap seen true
 d-i console-keymaps-at/keymap seen true
 <% end -%>
 
-<% subnet = @host.subnet -%>
+<% subnet = @host.provision_interface -%>
 <% if subnet.respond_to?(:dhcp_boot_mode?) -%>
   <% dhcp = subnet.dhcp_boot_mode? && !@static -%>
 <% else -%>


mmoll:
@ares could you have a look?

ares:
The provision interface is used only for PXE boot, the rest of the installation is intended to go through the primary interface. In your use case, I think you should be able to setup an installation medium so that it installer use local proxy, it works for me using apt-cacher-ng. For rpm-based distros you can use Katello plugin. The use case you mention for other communication such as puppet/monitoring sounds valid but I suppose that should be implemented as different flag, something like "maintenance interface". That could be used for puppet agent configuration for example.

I'm afraid the change can't be merged as is, it would change the behavior for all existing users. If you want to keep using it, I suggest you clone the template so you don't lose your change on upgrade.

hboetes:
I'm aware of the fact you use the primary interface for provisioning and the provisioning interface for pxe boot. I use the provisioning interface for provisioning and management, and the primary interface for client acces. If the rest of the world disagrees with me then so be it.

My setup works fine, thank you very much.

ares:
If you want to get more opinions I suggest you start thread on [foreman-users](https://groups.google.com/forum/#!forum/foreman-users) list. I hope [this update](https://github.com/theforeman/foreman/commit/3501ac9da1112e05a3571a10dbfe9610ca3b1c54) of inline help will minimize the confusion in future.

PR #397, opened 2017-06-08
Fix the TYPE's of ifcfgs

jovandeginste:
The network setup tools (at least for RHEL and CentOS) check the TYPE of
interfaces in some cases; they test for `bonding`, not `bond` or `Bond`.

Also, a vlan on top of a bond should also have this `TYPE=bonding`; the
vlan can not be activated if the `TYPE` is not set. The error I got in
this case was:

```
Error: Connection activation failed: Connection 'Vlan vlan123' is not
available on the device vlan123 at this time.
```

(Interface `vlan123` was configured on top of interface `bond0`, on top
of interfaces `eth0` and `eth1`).

This error was resolved as soon as I added `TYPE=bonding` to
`ifcfg-vlan123` - adding `TYPE=Bond` dit *NOT* work.

I have no idea whether the `TYPE=Alias` is used for anything, I could
not find it but would assume consistency from the side of the
developers...

### DIFF
--- a/provisioning_templates/snippet/kickstart_ifcfg_bond_interface.erb
+++ b/provisioning_templates/snippet/kickstart_ifcfg_bond_interface.erb
@@ -9,7 +9,7 @@ kind: snippet
                   :subnet => @subnet,
                   :subnet6 => @subnet6,
                   :dhcp => @dhcp }) -%>
-TYPE=Bond
+TYPE=bonding
 BONDING_OPTS="<%= @interface.bond_options -%> mode=<%= @interface.mode -%>"
 BONDING_MASTER=yes
 NM_CONTROLLED=no
--- a/provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
+++ b/provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
@@ -48,10 +48,10 @@ DEFROUTE=<%= primary %>
 <%-   if @interface.identifier.match(/^vlan\d+/) -%>
 <%=     "VLAN_NAME_TYPE=VLAN_PLUS_VID_NO_PAD" %>
 <%=     "PHYSDEV=#{@interface.attached_to}" %>
-<%-     if bonding_interfaces.include?(interface.attached_to) -%>
-<%=       "TYPE=bonding" %>
-<%-     end -%>
+<%-   end -%>
+<%-   if bonding_interfaces.include?(interface.attached_to) -%>
+<%=     "TYPE=bonding" %>
 <%-   end -%>
 <%- elsif @interface.virtual? && !@subnet.nil? && !@subnet.has_vlanid? && @interface.identifier.include?(':') -%>
-<%=   "TYPE=Alias" %>
+<%=   "TYPE=alias" %>
 <%- end -%>


mmoll:
@ares could you have a look?
@jovandeginste was this also tested on CentOS 6.x?

jovandeginste:
@mmoll Yes

ekohl:
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/sec-Create_a_Channel_Bonding_Interface.html states:

> The MAC address of the bond will be taken from the first interface to be enslaved. It can also be specified using the HWADDR directive if required. If you want NetworkManager to control this interface, remove the NM_CONTROLLED=no directive, or set it to yes, and add **TYPE=Bond** and BONDING_MASTER=yes.

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/sec-Network_Bonding_Using_the_Command_Line_Interface.html#sec-Create_a_Channel_Bonding_Interface als has an example where ```TYPE=Bond``` so this is weird. I don't currently have an EL7 machine at hand to verify this.

jovandeginste:
@ekohl I didn't test with NetworkManager.

jovandeginste:
@ekohl @mmoll apparently we _did_ test with NetworkManager. I was not aware at the time. Anyone else has practical experience with bonding can chime in?

ekohl:
Sorry, no experience with bonding here. All I can do is read the documentation.

jovandeginste:
Did some lookups on the ol' Google - the documentation may be lacking.
https://www.google.be/search?q=redhat+"type%3Dbonding"

Some results:
* https://wiki.linuxfoundation.org/networking/bonding https://www.kernel.org/doc/Documentation/networking/bonding.txt, section  "Using DHCP with initscripts"
* http://h20564.www2.hpe.com/hpsc/doc/public/display?docId=emr_na-c03578192
* https://www.redhat.com/archives/linux-cluster/2012-June/msg00020.html

mmoll:
@jovandeginste please rebase
@xprazak2 as you seem to work on the bonding topic also, could you have a look?

xprazak2:
I have no experience with bonds on RHEL6, sorry :(. My recent fix was related to how we keep track of interfaces attached to bond when Puppet uploads facts.

jovandeginste:
@mmoll rebased

mmoll:
I guess it won't make a difference, but please change it to "Bonding" with capital B, I can't really test this, but think this is good and will merge then. 👍

jovandeginste:
@mmoll just tested on live servers with `Bonding` instead of `bonding`, still works. I'll retry with `bond` for good measure - just to be sure.

jovandeginste:
So I am very confused now - it works with `TYPE=Bond` on both the bond and tagged vlans on top of it, on both el6 and el7. I will try again with the situation that certainly did nót work, which was not having the `TYPE` on the vlan (but have it on the bond itself)

jovandeginste:
Even this works now. I don't understand anymore.

lzap:
This is indeed confusing. When NetworkManager is enabled, it expects TYPE=Bond:

https://developer.gnome.org/NetworkManager/stable/nm-settings-ifcfg-rh.html

There is no case sensitivity info, so I'd rather stick with capital letter. Now, when you are using sysv init scripts then bonding TYPE does not exist. At least in the documentation (/usr/share/doc/initscripts/sysconfig.txt) and even in examples you set Ethernet in all cases:

```
cat /usr/share/doc/initscripts/examples/networking/ifcfg-bond-* | grep TYPE
TYPE=Ethernet
TYPE=Ethernet
TYPE=Ethernet
TYPE=Ethernet
```

lzap:
We have completely redesigned our network scripts recently and it is possible that bonding on sysvinit was not tested.

ares:
@jovandeginste any progress on this? Should we keep it open?

jovandeginste:
Since no-one else experiences issues here, and the documentation is lacking/inconsistend, I suggest to revisit this issue at a later time. I'll try to find a slot where I can reinstall a bunch of different servers, with different case for these keys, and find out what works when...

lzap:
I am going to stick with NM official documentation, the correct TYPE is "Bond". If you feel it should have been changed in code or docs, please file a ticket to nm project.

PR #398, opened 2017-06-09
An interface may only have an ipv6 subnet attached

jovandeginste:
This will still generate the network interface if only an ipv6 subnet is
attached to a given (regular) interface.


mmoll (approved):

ekohl:
Thanks!

PR #399, opened 2017-06-14
Remove SSH key type from create_users

dLobatog:
In Foreman, you cannot write the SSH key without the key type.
This content is already in `ssh_key`, so there is no need to put it
twice using `key.type`. In fact this would not work in .authorized_keys

screenshot from 2017-06-14 14-13-23


dLobatog:
Just saw https://github.com/theforeman/community-templates/pull/392 , nevermind!

PR #402, opened 2017-06-20
Fix bond deployments

xprazak2:
Needs a change in [Foreman](https://github.com/theforeman/foreman/pull/4583) to work.

### DIFF
--- a/provisioning_templates/snippet/kickstart_networking_setup.erb
+++ b/provisioning_templates/snippet/kickstart_networking_setup.erb
@@ -36,8 +36,9 @@ kind: snippet
 <%-   next if !interface.managed? || (interface.subnet.nil? && interface.subnet6.nil?) -%>
 <%-   next if bonded_interfaces.include?(interface.identifier) -%>
 
+<%-   interface_identifier = @host.bond_interfaces.map { |i| i.identifier }.include?(interface.attached_to) ? interface.identifier : nil %>
 <%=   "# #{interface.identifier} interface" %>
-<%=   snippet('kickstart_ifcfg_get_identifier_names', :variables => { :interface => interface }) -%>
+<%=   snippet('kickstart_ifcfg_get_identifier_names', :variables => { :interface => interface, :identifier => interface_identifier }) -%>
 <%-   ifcfg = snippet('kickstart_ifcfg_generic_interface', :variables => {
                         :interface => interface,
                         :subnet => interface.subnet,


provisioning_templates/snippet/kickstart_networking_setup.erb
```diff
@@ -36,8 +36,9 @@ kind: snippet
 <%-   next if !interface.managed? || (interface.subnet.nil? && interface.subnet6.nil?) -%>
 <%-   next if bonded_interfaces.include?(interface.identifier) -%>
 
+<%-   interface_identifier = @host.bond_interfaces.map(&:identifier).include?(interface.attached_to) ? interface.identifier : nil %>
```
ares: the map must be rewritten to `.map { |i| i.identifier }` syntax, the shorter was removed from safe_mode

provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
```diff
@@ -40,7 +40,7 @@ DEFROUTE=<%= primary %>
 <%-     end -%>
 <%-   end -%>
 <%- end -%>
-<%- if @interface.virtual? && ((!@subnet.nil? && @subnet.has_vlanid?) || @interface.vlanid.present?) -%>
+<%- if @interface.virtual? && ((!@subnet.nil? && @subnet.has_vlanid?) || (@interface.tag.present? && @interface.vlanid.present?)) -%>
```
ares: why the tag must be present? the vlanid is taking tag into consideration, phrasing the code

```
    def vlanid
      # Determine a vlanid according to the following cascading rules:
      # 1. if the interface has a tag, use that as the vlanid
      # 2. if the interface has a v4 subnet with a non-blank vlanid, use that
      # 3. if the interface has a v6 subnet with a non-blank vlanid, use that
      # 4. if no reasonable vlanid was determined, then return an empty string
      #
      # In case the v4 and v6 subnet are both present, they should have the same
      # vlanid. If they have a different vlanid, this is probably an error in the
      # user's database. There is no way of determining the real vlanid, so we
      # pick the v4 one unless it turns out to be blank.

      return self.tag unless self.tag.blank?
      return self.subnet.vlanid if self.subnet && self.subnet.vlanid.present?
      return self.subnet6.vlanid if self.subnet6 && self.subnet6.vlanid.present?
      return ''
    end
```

ares: actually testing the deployment without this part of change seems to work well

ares (changes requested):
nearly there :-)

xprazak2:
I updated as suggested.

ares (approved):

ares:
The linked PR just got in so this one is also good to go, I'll create sync PR since it would not work in nightly without this change while the redmine issue would indicate it's resovled...

Thanks @xprazak2!

PR #404, opened 2017-06-28
Trigger kickstart networking provisioning with only IPv6

dLobatog:
If a host has managed IPv6 networks, but no IPv4, the kickstart will
not even try to configure it. This change allows to configure if:

Only IPv4 is present
Only IPv6 is present
Both IPv4 and IPv6 are present

cc @jovandeginste might be useful for you, I needed the changes you made to the kickstart networking snippet for this to make any sense.

### DIFF
--- a/provisioning_templates/provision/kickstart_default.erb
+++ b/provisioning_templates/provision/kickstart_default.erb
@@ -137,7 +137,7 @@ exec < /dev/tty3 > /dev/tty3
 #changing to VT 3 so that we can see whats going on....
 /usr/bin/chvt 3
 (
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
+<% if subnet.respond_to?(:dhcp_boot_mode?) || @host.subnet6.respond_to?(:dhcp_boot_mode?) -%>
 <%= snippet 'kickstart_networking_setup' %>
 <% end -%>
 
--- a/provisioning_templates/provision/kickstart_rhel_default.erb
+++ b/provisioning_templates/provision/kickstart_rhel_default.erb
@@ -114,7 +114,7 @@ exec < /dev/tty3 > /dev/tty3
 #changing to VT 3 so that we can see whats going on....
 /usr/bin/chvt 3
 (
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
+<% if subnet.respond_to?(:dhcp_boot_mode?) || @host.subnet6.respond_to?(:dhcp_boot_mode?) -%>
 <%= snippet 'kickstart_networking_setup' %>
 <% end -%>
 


provisioning_templates/provision/kickstart_default.erb
```diff
@@ -137,7 +137,7 @@ exec < /dev/tty3 > /dev/tty3
 #changing to VT 3 so that we can see whats going on....
 /usr/bin/chvt 3
 (
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
+<% if subnet.respond_to?(:dhcp_boot_mode?) || @host.subnet6.nil? && @host.subnet6.respond_to?(:dhcp_boot_mode?) -%>
```
jovandeginste: How does this work? You test if `subnet6` is nil, and then call `respond_to?` on it?

Should this be:
```ruby
if subnet.respond_to?(:dhcp_boot_mode?) || (@host.subnet6.present? && @host.subnet6.respond_to?(:dhcp_boot_mode?))
```

And since `NilClass` (hopefully) doesn't `respond_to?` `:dhcp_boot_mode?` (but the call exists), this can actually just be:

```ruby
if subnet.respond_to?(:dhcp_boot_mode?) || @host.subnet6.respond_to?(:dhcp_boot_mode?)
```

provisioning_templates/provision/kickstart_rhel_default.erb
```diff
@@ -114,7 +114,7 @@ exec < /dev/tty3 > /dev/tty3
 #changing to VT 3 so that we can see whats going on....
 /usr/bin/chvt 3
 (
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
+<% if subnet.respond_to?(:dhcp_boot_mode?) || @host.subnet6.nil? && @host.subnet6.respond_to?(:dhcp_boot_mode?) -%>
```
jovandeginste: Same as the other comment, I think :-)

jovandeginste (commented):
I added some comments, I'm not sure how the code works..?

dLobatog:
@jovandeginste Thanks - It should've been `!@host.subnet6.nil?`, but I shortened it thanks to your suggestion 😄

jovandeginste:
@dLobatog Can we not make `subnet6` available just like `subnet`, instead of having to call `@host.subnet6`? Would be more consistent...

dLobatog:
@jovandeginste It's not that `subnet` is available, it's just assigned to that local variable at the beginning of the script: https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/provision/kickstart_default.erb#L47

jovandeginste:
Yes I was just now reading the rest of the file. Are there no changes needed there too?

dLobatog:
I don't think so, the point is just that the script is triggered to configure networking - the other call `network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select(&:present?).join(',')}" %> --hostname <%= @host %><%= os_major >= 6 ? " --device=#{@host.mac}" : '' -%>` I have not tried with IPv6 so I prefer not to change it yet.

jovandeginste:
@dLobatog fair enough! I was just thinking that it might be useful to do something of this kind:

```ruby
subnet = @host.subnet || @host.subnet6
```

But maybe that's for another time ;-)

mmoll:
merged, gracias @dLobatog!

PR #405, opened 2017-07-04
Ansible Tower provisoning callback

dLobatog:
Service to pull and run provisioning scripts from Ansible tower
post-installation. The snippet is compatible only with systemd-enabled
servers for the moment

props & shoutouts to @wzzrd

### DIFF
--- a/provisioning_templates/provision/kickstart_default.erb
+++ b/provisioning_templates/provision/kickstart_default.erb
@@ -192,6 +192,15 @@ yum -t -y remove kernel-uek*
 sed -e 's/DEFAULTKERNEL=kernel-uek/DEFAULTKERNEL=kernel/g' -i /etc/sysconfig/kernel
 <% end -%>
 
+# This is RHEL7 specific code, as it uses systemctl, replace with
+# chkconfig or an @boot cron if you run RHEL5 or RHEL6
+<% if realm_compatible && host_param_true?('ansible_tower_provisioning') %>
+<%= save_to_file('/etc/systemd/system/ansible-callback.service',
+                 snippet('ansible_tower_callback_service')) %>
+# Runs during first boot, removes itself
+/usr/bin/systemctl enable ansible-callback
+<% end -%>
+
 sync
 
 # Inform the build system that we are done.
--- a/provisioning_templates/provision/kickstart_rhel_default.erb
+++ b/provisioning_templates/provision/kickstart_rhel_default.erb
@@ -165,6 +165,15 @@ fi
 <%= snippet 'saltstack_setup' %>
 <% end -%>
 
+# This is RHEL7 specific code, as it uses systemctl, replace with
+# chkconfig or an @boot cron if you run RHEL5 or RHEL6
+<% if realm_compatible && host_param_true?('ansible_tower_provisioning') %>
+<%= save_to_file('/etc/systemd/system/ansible-callback.service',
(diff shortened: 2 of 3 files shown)


wzzrd:
Yay!

ekohl:
Why such a complicated solution to do one curl call? We do the foreman post-installation HTTP call during provisioning (like [here](https://github.com/theforeman/community-templates/blob/5ca7947f02a4989fd21ecd4f9654df705c1577b3/provisioning_templates/provision/kickstart_default.erb#L198)).

dLobatog:
@ekohl I guess the benefit of this is that you would be able to repeat this any time by just restarting the service

ekohl:
To me it doesn't feel like a service. You could also write a ```/usr/local/bin/ansible-tower-callback``` and execute that during provisioning. It'd have the same benefits and not be limited to systemd. While most Linux distro's are switching over, the BSDs we support won't.

wzzrd:
So the point of doing it this way is that the callback to Tower needs to happen *after* the reboot after kickstarting. 

The callback is going return immediately and will trigger an Ansible run from Tower, so we don't want to reboot right after the cURL call returns, because that will likely make the Ansible run from Tower fail. The host will probably either unreachable during reboot or shutting down after kickstarting.

Reason I used a service for this was that services have been used in this way before, and it was an easy solution when I really needed it. 

I am sure we can improve upon this and rewrite the snippet into something that is more portable though.

ekohl:
Ah, so it's more than "just registration". I wasn't aware of the callback. In that case it does make sense.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -191,6 +191,17 @@ yum -t -y remove kernel-uek*
 sed -e 's/DEFAULTKERNEL=kernel-uek/DEFAULTKERNEL=kernel/g' -i /etc/sysconfig/kernel
 <% end -%>
 
+# This is RHEL7 specific code, as it uses systemctl, replace with
+# chkconfig or an @boot cron if you run RHEL5 or RHEL6
+<% if realm_compatible && @host.param_true?('ansible_tower_provisioning') %>
+cat > /etc/systemd/system/ansible-callback.service << EOF
```
mmoll: can `save_to_file` be used here?

dLobatog: indeed

dLobatog (commented):
Updated to use `save_to_file`, thanks for the tip @mmoll

mmoll:
this needs to use `host_param_true?` now.

mmoll:
merged after updating to new macros. gracis @dLobatog & bedankt @wzzrd!

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -192,6 +192,15 @@ yum -t -y remove kernel-uek*
 sed -e 's/DEFAULTKERNEL=kernel-uek/DEFAULTKERNEL=kernel/g' -i /etc/sysconfig/kernel
 <% end -%>
 
+# This is RHEL7 specific code, as it uses systemctl, replace with
```
ekohl, line 195: Nit that's a bit late: The comment should be inside the if statement so you won't end up with just a comment when it's false. I've submitted https://github.com/theforeman/community-templates/pull/410 to fix this.

PR #408, opened 2017-07-17
Use kernel-detected mac addresses instead of current configuration

jovandeginste:
We scan the mac addresses of interfaces according to the kernel,
instead of relying on what's currently configured.

This might solve issues if you use bonding during installation
(when your two ethernet adapters have the same mac address). The
order here is to first seek bonding slaves, and then consider all
other devices.


lzap:
Can you add `2>/dev/null` to the grep command, because on some platforms you can hit:

```
# grep -l xxxxxxxxxx /sys/class/net/*/{bonding_slave/perm_hwaddr,address} | awk -F '/' '// {print $5}' | head -1
grep: /sys/class/net/*/bonding_slave/perm_hwaddr: No such file or directory
enp30s0
```

jovandeginste:
@lzap Something like this?

lzap (approved):
Something like this, @ares any objections?

ares:
No objections, I haven't tested it but looks good to me.

lzap:
Thanks a lot.

PR #410, opened 2017-08-01
Place the comment inside the if statement

ekohl:


mmoll:
true. merged, bedankt @ekohl!

PR #415, opened 2017-08-24
Added ReX reboot snippet

lzap:
SSIA

We saw behavior when the task is reexecuted again, we are still investigating the root cause of this.


mmoll:
I saw something similar and I think the cause was that the task was still in the proxy-side tasks DB as "to execute".

provisioning_templates/snippet/remote_execution_reboot.erb
```diff
@@ -0,0 +1,16 @@
...
+# Proper reboot of a host. It is not recommended to use "shutdown" command
+# directly as it kills SSH connection and Remote Execution may mark the job
+# as failed which might lead to reexecution in the worst case.
+#
+# Parameters:
+#
+# remote_execution_minutes_to_reboot: number of minutes to reboot (default 1)
+#
+
+<% min = host_param('remote_execution_minutes_to_reboot') || 1 %>
```
ekohl, line 15: Does it make sense to do this in seconds? A minute is a long time and I can imagine some users might want to drop to 30 seconds.

iNecas:
I don't think this code itself has the relation to the re-execution of the commands, compared to the fact that there were other factors in the reproducing environment (such as periodic restarts for smart_proxy_dynflow_core), that could lead to it. There already is a template for that here https://github.com/theforeman/community-templates/blob/develop/job_templates/power_action_-_ssh_default.erb. Anyway, I would deffer with this until we have confirmed that this is the reason for the observed behavior, as I'm not convinced that this would have any effect on that at the moment.

lzap:
I did not know that ReX carries own templates in its git repo. Closing.

ares:
@lzap just to clarify, templates in REX repo should be in sync with what we have here, it's the same model that Foreman uses. If this snippet is useful, it belongs here.

PR #416, opened 2017-08-26
add support for Puppet 5 agent installs

mmoll:


ekohl (approved):
Sad that the subdir is needed.

PR #417, opened 2017-08-29
Fixed PXEGrub2 MAC file loading

lzap:
We have reports that the debian workaround causes issues on VMWare (RH Grub2 build). This should help.


mmoll:
looks sensible, merged, díky @lzap!

PR #418, opened 2017-08-31
Fixes #20814 - Added insmod regexp, removed test

lzap:
So my previous patch actually did not work, test statement searches for local files and not remote files, therefore the file was never sourced. It looks like `source` command fail silently, so the test statement is not necessary.

But we found a different thing - some grub2 binaries (namely the one we ship in foreman-bootloaders) do not ship with `regexp` command compiled in and this must be inserted from a module `regexp.mod`. The `insmod` statement does that, otherwise the regexp commandd fails silently. We will need to add modules into foreman-bootloaders.


lzap:
The `insmod` fails silently if module is not present, so this won't regress.

provisioning_templates/PXEGrub2/pxegrub2_global_default.erb
```diff
@@ -7,12 +7,15 @@ name: PXEGrub2 global default
...
+# cases, make sure "regexp.mod" file is present on the TFTP. For more info see:
+# https://bugzilla.redhat.com/show_bug.cgi?id=1370642#c70
+insmod regexp
 regexp --set=1:m1 --set=2:m2 --set=3:m3 --set=4:m4 --set=5:m5 --set=6:m6 '^([[:xdigit:]]{1,2})\:([[:xdigit:]]{1,2})\:([[:xdigit:]]{1,2})\:([[:xdigit:]]{1,2})\:([[:xdigit:]]{1,2})\:([[:xdigit:]]{1,2})' "$net_default_mac"
 mac=${m1}-${m2}-${m3}-${m4}-${m5}-${m6}
 configfile=/grub2/grub.cfg-01-$mac
-if [ -f "$configfile" ]; then
-  source "$configfile"
-fi
+source "$configfile"
```
lzap, line 18: @dLobatog watch out when backporting - the `if [ -f "$configfile" ]; then` was added yesterday but I found it was incorrect. So just ignore that, the only relevant bit here is "insmod regexp"

dLobatog: @lzap We don't do backporting with community templates - they are synced on every minor release

dLobatog (approved):
@lzap Seems reasonable to me, although I haven't tried it, just read some of the BZs associated. 

Will we have to do anything when a release of EL7 that fixes the ':' bug?

dLobatog:
Thanks @lzap !

lzap:
Thanks, I can confirm this won't cause any problems, `insmod` just prints an error if module cannot be found but Grub2 continues execution the next statement.

PR #419, opened 2017-09-15
Complement to Bug #20814

lero:
Because of the buggy GRUB2 in RHEL/CentOS 7.4, we also need to remove the last ':' from the MAC address for Foreman Discovery cmdline.


mmoll:
@lzap could you have a look?

lzap:
Thanks, yeah bummer. Could you move the mac variable creation to global template, name it something like "mac_colons" or similar and then use it here but fallback to "net_default_mac" when not set? Just in case someone uses this snippet outside of global template context.

lero:
Sorry for the delay. I think we can close this one because RedHat already fixed their grub2 package: https://access.redhat.com/errata/RHBA-2017:2950

PR #420, opened 2017-09-15
container linux: install vmware specific image

timogoebel:
... installs the VMWare specific CoreOS Container Linux image which contains `vmware-tools`.

https://coreos.com/os/docs/latest/installing-to-disk.html


ekohl (approved):

mmoll:
merged, danke @timogoebel!

PR #421, opened 2017-09-18
container linux: specify desired version

timogoebel:
This specifies the desired CoreOS Container Linux version explicitly.


mmoll:
merged, danke @timogoebel!

PR #422, opened 2017-09-26
Fix BOOTIF MAC address notation in preseed_default_ipxe

akhirai:
This patch fixes MAC address notation convention of BOOTIF parameter.
Debian-installer doesn't recognize BOOTIF parameter with colon separated values.
The following is an message recorded in /var/log/installer/syslog.

```
netcfg[2157]: INFO: Could not find valid BOOTIF= entry in /proc/cmdline
```

If the secondary NIC (e.g. eno2) is a provisioning network, installation fails at the time of network detection.
This patch changes the notation to 01 prefix hyphen separated values.


lzap:
Thanks!

PR #425, opened 2017-09-29
Fix include issue when using puppetlabs 5 repo

derektamsen:
I noticed an issue when using the puppet labs 5 repo the repository was not being setup on Debian/Ubuntu and Centos/RHEL systems. This ensures the `puppetlabs_repo` snippet is included when using `enable-puppetlabs-puppet5-repo`. Otherwise puppet is attempted to be installed from the distrobutions repo instead of the puppetlabs repo.


mmoll:
thanks a lot @derektamsen!

PR #427, opened 2017-10-02
Allow EPEL repo URL base to be changed

treydock:
This would be handy for sites that prefer to use a local mirror.


mmoll:
seems to work just fine, merged, thanks @treydock!

PR #428, opened 2017-10-02
Allow puppetlabs repo base URL to be modified

treydock:
This would be useful for sites that prefer to use local mirrors.

### DIFF
--- a/provisioning_templates/snippet/puppetlabs_repo.erb
+++ b/provisioning_templates/snippet/puppetlabs_repo.erb
@@ -3,6 +3,7 @@ kind: snippet
 name: puppetlabs_repo
 -%>
 <%
+repo_base    = host_param('puppetlabs-repo-base') ? host_param('puppetlabs-repo-base') : nil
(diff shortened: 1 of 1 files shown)


mmoll:
I think this doesn't work for quite some situations, e.g. when having something like http://my.mirr.or/pupeptlabs/apt and http://my.mirr.or/puppetlabs/yum. See GH-267 for more discussion on that topic.

treydock:
Should work, my local mirror is https://fqdn/puppetlabs and so that'd be value of puppetlabs-repo-base.  That discussion looks to suggest more parameters isn't always better which makes sense.

mmoll:
@treydock Well, my mirror is not organzied in that way... ;)

I'm not opposed to adding more parameters, it's not nice, but the result of the fact that we don't have complex data structures available to set for the templates.

tbrisker:
This PR has been inactive for a long time now, closing. Feel free to reopen it if it is still relevant. thanks @treydock !

PR #430, opened 2017-10-05
Refs #21204 - default PXE entry from settings

lzap:
This unifies how default entry is being selected via new setting helper function. Note grub1 only supports number, there is a conversion. Method "index" should be allowed by safemode. For grub2 labels are created using "id" option:

https://www.gnu.org/software/grub/manual/grub/grub.html#default

See https://github.com/theforeman/foreman/pull/4891 for more info.


lzap:
Mis-pushed an old commit, rebased. Do not merge until the core PR is merged of course :-)

mmoll:
please rebase :)

lzap:
Will do once core is merged.

provisioning_templates/PXEGrub2/pxegrub2_default_local_boot.erb
```diff
@@ -4,7 +4,7 @@ name: PXEGrub2 default local boot
 %>
 <%# Used to boot provisioned hosts, do not associate or change the name. %>
 
-set default=0
+set default=<%= global_setting("default_pxe_item_local", "local_chain_hd0") %>
```
iNecas, line 8: Shouldn't we use the `index` here as well?

provisioning_templates/PXEGrub/pxegrub_global_default.erb
```diff
@@ -3,8 +3,12 @@ kind: PXEGrub
 name: PXEGrub global default
 %>
 <%# Used to boot unknown hosts, do not associate or change the name. %>
-
-default=0
+<%
+  # Grub1 only supports numeric default statement, this array allows conversion to number
+  items = ["local", "discovery"]
+  default_item = items.index(global_setting("default_pxe_item_global", "local"))
```
iNecas: What happens, if the item is not found? Shouldn't we raise some error with explanation what went wrong?

provisioning_templates/PXEGrub2/pxegrub2_global_default.erb
```diff
@@ -4,7 +4,7 @@ name: PXEGrub2 global default
 %>
 <%# Used to boot unknown hosts, do not associate or change the name. %>
 
-default=0
+default=<%= global_setting("default_pxe_item_global", "local") %>
```
iNecas: seems like index is also expected here.

provisioning_templates/PXEGrub2/pxegrub2_default_local_boot.erb
iNecas: Scratch that: I'm reviewing top-bottom, I see the changes in the snippets now :)

provisioning_templates/PXEGrub/pxegrub_global_default.erb
lzap: I think a sane approach is to add `|| 0` which works with both PXELinux and Grub. Let me add that and rebase.

ares:
core is merged, @iNecas were your comments addressed?

iNecas:
Yes

lzap:
What are we waiting for then? :-)

Docs: https://github.com/theforeman/theforeman.org/pull/1026

mmoll:
merged, díky @lzap!

PR #431, opened 2017-10-09
Enable Puppet service for Ubuntu

### LINKED ISSUE #414: Puppet service not started on Ubuntu
When provisioning an Ubuntu box (e.g. 17.04 Zesty) then the `puppet` service is not being started (enabled). This may be due to some code in [provisioning_templates/snippet/puppet_setup.erb](https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/puppet_setup.erb#L112-L145), which probably unsuccessfully tries to make this work for Ubuntu.

Specifically, `facter` never yields a value for neither `os_name` nor `os_major`, which are [used in the `if`](https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/puppet_setup.erb#L113-L119) that tries to work around Puppet issues [PA-136](https://tickets.puppetlabs.com/browse/PA-136) and [PUP-5577](https://tickets.puppetlabs.com/browse/PUP-5577) (as quoted in the comment). Not sure if this observation points into the right direction.

On a fresh Ubuntu Zesty node the output of `facter` only contains two properties starting with`os`:
```ruby
$ facter | grep ^os
os => {"name"=>"Ubuntu", "family"=>"Debian", "release"=>{"major"=>"17.04", "full"=>"17.04"}, "lsb"=>{"distcodename"=>"zesty", "distid"=>"Ubuntu", "distdescription"=>"Ubuntu 17.04", "distrelease"=>"17.04", "majdistrelease"=>"17.04"}}
osfamily => Debian
```
(Note: The node under test was rebooted twice after provisioning; once after the OS installation, as foreseen, and the second time after manually running `service puppet start` and a subsequent installation of software by the Puppet agent.) 

Software being used/installed on a fresh node is:
```bash
$ puppet --version
4.8.2
$ facter --version
2.4.6
```
Is this problem already known?

clecherbauer:
With Ubuntu the Puppet service is not activated at the completion of the setup. These changes ensure Puppet is activated just as with Debian.

Fixes #414.


ekohl (approved):

mmoll:
merged, danke @ChristianLecherbauer!

PR #432, opened 2017-10-10
fix preseed partitioning expert recipe gsub

derektamsen:
Modified the gsub search string for the preseed partman expert recipe as it rendered incorrectly. I am not sure if this is the best regex to fix the issue. However, it renders correctly with the example `boot-root` recipe.

Prior to the fix the example below:
```
boot-root ::
        1 1 1 free
                $gptonly{ }
                $primary{ }
                $bios_boot{ }
                method{ biosgrub }
        .
```
would result in:
```
d-i partman-auto/expert_recipe string \
boot-root ::
 \
        1 1 1 free
 \
                $gptonly{ }
 \
                $primary{ }
 \
                $bios_boot{ }
 \
                method{ biosgrub }
 \
        .
 \
```

after update renders as:
```
d-i partman-auto/expert_recipe string \
boot-root :: \
        1 1 1 free \
                $gptonly{ } \
                $primary{ } \
                $bios_boot{ } \
                method{ biosgrub } \
        . \
```


mmoll:
merged, thanks @derektamsen!

PR #435, opened 2017-10-13
allow partitioning a preseeded system without a swap partition

derektamsen:
This allows partitioning a Debian and Ubuntu system without a swap partition. By default Ubuntu will warn during installation if a swap partition is not specified. By setting `partitioning-allow-noswap` `true` the warning will be suppressed and a system can be configured without a swap partition.

https://github.com/turnkeylinux/di-live/blob/7c72192d58466e824b112b037ae6489b289881b2/d-i/source/partman-basicfilesystems/debian/partman-basicfilesystems.templates#L46


mmoll:
merged, thanks @derektamsen!

PR #436, opened 2017-10-13
allow specifying disk label during preseed

derektamsen:
Add the ability to manually specify a disk label when formatting a disk for debian or ubuntu during preseed. This is necessary when using an expert recipe as the label seems to default to `msdos` instead of being automatically calculated. This allows setting the label via `partitioning-disk-label` if for example `gpt` was needed.

https://lists.debian.org/debian-user/2012/05/msg02687.html

_Also fixed a spelling error in the readme._


mmoll:
@derektamsen please rebase :)

derektamsen:
Sure no problem.

mmoll:
merged, thanks @derektamsen!

PR #437, opened 2017-10-18
Add gpg key for puppet

hboetes:
Puppetlabs started signing packages, since the key was missing from the provisioning template the repo was rejected by the installed. This patch adds the key.


mmoll:
please rebase onto the current develop branch and squash everything into one commit, then force-push 📍

hboetes:
yes, this is getting messy

hboetes:
ok, all fixed. Thanks @mmoll !

mmoll:
I didn't test this, but it can't be more broken than before, so let's pull that in...

mmoll:
merged, thanks @hboetes!

hboetes:
I have no idea how it happened, but the key is wrong. :-(
It should be 7F438280EF8D349F and not 2ABFA143A0E46E11, which is the open suse key id.

PR #438, opened 2017-10-19
Prevent from grub2 MAC addres bug in 7.4

lzap:
Only grub2 is affected, not grub1 or pxelinux.


rplevka (approved):
I confirm this works fine and it's indeed needed in order for FDI to send out a correct mac address value

mmoll:
@lzap could you open an issue in the upstream redmine and connect that BZ?

lzap:
FYI RHEL 7.4 finally got an update now: https://access.redhat.com/errata/RHBA-2017:2950

The BZ was closed as dupe, not sure what you mean we do not use Red Mine for templates. Removed the incorrect Fixes link tho.

mmoll:
Oh, I thought there would be an upstream issue... anyway, is this still relevant after the update?

lzap:
Is this really mandatory, I don't see RM issues very often in the main branch.

mmoll:
nope, not mandatory.

lzap:
The purpose of this patch is to do the workaround for discovery as well. Although it has been fixed in RHEL already, I suggest to have the workaround there (we already have this for localboot template).

mmoll:
merged, díky @lzap!

PR #441, opened 2017-11-07
For some mysterious reason I pasted the suse fingerprint and not the puppetlabs fingerprint.

hboetes:
For some mysterious reason I pasted the suse fingerprint and not the puppetlabs fingerprint.


mmoll:
merged, thanks @hboetes!

PR #442, opened 2017-11-21
fixes #21694 - Add MTU support for kickstart installations

bagasse:
Template parts for [#21694](http://projects.theforeman.org/issues/21694).

Adding MTU support for kickstart templates. I didn't see any resources about MTU support at provisioning time on other distribution (except for Suse)

### DIFF
--- a/provisioning_templates/provision/atomic_kickstart_default.erb
+++ b/provisioning_templates/provision/atomic_kickstart_default.erb
@@ -15,7 +15,7 @@ timezone --utc <%= host_param('time-zone') || 'UTC' %>
 <% dhcp = !@static -%>
 <% end -%>
 
-network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select{ |item| item.present? }.join(',')}" %> --hostname <%= @host %><%= " --device=#{@host.mac}" -%>
+network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select{ |item| item.present? }.join(',')} --mtu=#{subnet.mtu.to_s}" %> --hostname <%= @host %><%= " --device=#{@host.mac}" -%>
 
 # Partition table should create /boot and a volume atomicos
 <% if @dynamic -%>
--- a/provisioning_templates/provision/kickstart_default.erb
+++ b/provisioning_templates/provision/kickstart_default.erb
@@ -52,7 +52,8 @@ skipx
 <% else -%>
 <% dhcp = !@static -%>
 <% end -%>
-network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select{ |item| item.present? }.join(',')}" %> --hostname <%= @host %><%= os_major >= 6 ? " --device=#{@host.mac}" : '' -%>
+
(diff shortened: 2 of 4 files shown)


provisioning_templates/provision/kickstart_default.erb
```diff
@@ -50,7 +50,10 @@ skipx
 <% else -%>
 <% dhcp = !@static -%>
 <% end -%>
-network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select{ |item| item.present? }.join(',')}" %> --hostname <%= @host %><%= os_major >= 6 ? " --device=#{@host.mac}" : '' -%>
+<% if subnet.respond_to?(:mtu) && !subnet.mtu.blank? -%>
+<% mtu = subnet.mtu.to_s %>
```
ares: This makes mtu empty string in case it's nil, in line below it adds mtu param `if mtu` which would be true in this case, since empty string are evaluated as true. The same applies for user leaving the text field empty. I wonder if it wouldn't be better to always have mtu value set on subnet as required field and default to 1500.

provisioning_templates/provision/atomic_kickstart_default.erb
```diff
@@ -13,8 +13,11 @@ timezone --utc <%= host_param('time-zone') || 'UTC' %>
 <% else -%>
 <% dhcp = !@static -%>
 <% end -%>
+<% if subnet.respond_to?(:mtu) && !subnet.mtu.blank? -%>
```
ares: Since templates are now bramched per Foreman version I think the `respond_to?` check is unnecessary and should be removed.

provisioning_templates/provision/kickstart_default.erb
ares: Actually it would not be an issue since there's check for blank on line above, but still, wouldn't it be better if we couls rely on mtu would always be there?

ares (changes requested):

bagasse:
Change templates to reflect that MTU will be always present.

lzap (approved):
Well done, excellent thanks.

lzap:
Will let this on @ares since he started the review, LGTM.

ares:
waiting on https://github.com/theforeman/foreman/pull/5017 to test

mmoll:
@bagasse please rebase

bagasse:
@mmoll done.

bagasse:
The foreman part [#21694](http://projects.theforeman.org/issues/21694) was merged yesterday.

mmoll:
I'll look into this over the weekend :+1:

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -52,7 +52,8 @@ skipx
 <% else -%>
 <% dhcp = !@static -%>
 <% end -%>
-network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select{ |item| item.present? }.join(',')}" %> --hostname <%= @host %><%= os_major >= 6 ? " --device=#{@host.mac}" : '' -%>
+
+network --bootproto <%= dhcp ? 'dhcp' : "static --ip=#{@host.ip} --netmask=#{subnet.mask} --gateway=#{subnet.gateway} --nameserver=#{[subnet.dns_primary, subnet.dns_secondary].select{ |item| item.present? }.join(',')} --mtu=#{subnet.mtu.to_s}" %> --hostname <%= @host %><%= os_major >= 6 ? " --device=#{@host.mac}" : '' -%>
```
mmoll, line 56: shouldn't the MTU get set always here?

mmoll:
@bagasse please check the `network` line in the two kickstart templates, working fine otherwise :+1:

provisioning_templates/provision/kickstart_default.erb
bagasse: @mmoll, as described in [RFC2132](https://www.ietf.org/rfc/rfc2132.txt), the MTU can be set via DHCP.

eg, for ISC DHCPD:
```
option interface-mtu 9000;
```

So I think that if a dhcp server is used to provide network configuration for hosts on a network, the MTU will be set via the dhcp server too.

mmoll: I see, let's bring this in then :+1:

mmoll:
merged, merci @bagasse!

PR #443, opened 2017-12-03
Use single quotes around passwords to avoid shell interpretation

stdietrich:
Special characters in the password might be interpreted by the shell. This might lead to a failed registration at RHEL.
By using single quotes around the password, this will be avoided.


mmoll (approved):

mmoll:
makes sense visually... any RH people care to test?

provisioning_templates/snippet/redhat_register.erb
```diff
@@ -142,18 +142,18 @@ name: redhat_register
...
     <% end %>
   <% end %>
   <% if host_param('subscription_manager_username') && host_param('subscription_manager_password') %>
     <% if host_param('subscription_manager_pool') %>
-      subscription-manager register --name="<%= @host.name %>" --username="<%= host_param('subscription_manager_username') %>" --password="<%= host_param('subscription_manager_password') %>"
+      subscription-manager register --name="<%= @host.name %>" --username="<%= host_param('subscription_manager_username') %>" --password='<%= host_param("subscription_manager_password") %>'
       subscription-manager attach --pool="<%= host_param('subscription_manager_pool') %>"
     <% else %>
-      subscription-manager register --name="<%= @host.name %>" --username="<%= host_param('subscription_manager_username') %>" --password="<%= host_param('subscription_manager_password') %>" --auto-attach
+      subscription-manager register --name="<%= @host.name %>" --username="<%= host_param('subscription_manager_username') %>" --password='<%= host_param("subscription_manager_password") %>' --auto-attach
```
ares: Looks good to me but if you're touching this, would you mind also fixing name and username? I guess there are more strict validations for these, but still...

stdietrich:
I have added single quotes around the user- and hostnames and updated the branch accordingly.

ekohl (approved):
Looks good. @ares can you have another look?

ares:
Tested what I could and works well, thanks @stdietrich, there's another call on line 167 but it does not contain username/password so that should be OK. Merging.

PR #444, opened 2017-12-22
Refs #16740 - info was removed in 1.17

ares:
This needs to be synced for 1.17 as well, info is no longer available


ares:
@xprazak2 FYI

mmoll:
merged, díky @ares!

PR #445, opened 2017-12-29
Add model metadata attribute

ares:
Foreman adds model attribute to metadata during export, which makes it easier to import the template later just from file without further knowledge. It's useful in foreman_templates. I updated all templates in this repo to contain this attribute too, plus added a test that ensures all new templates have this attribute filled. As a small bonus, I updated failure message for this and similar tests.

It would be great to get this in early since the PR will get out of date easily.

### DIFF
--- a/job_templates/package_action_-_ssh_default.erb
+++ b/job_templates/package_action_-_ssh_default.erb
@@ -1,6 +1,7 @@
 <%#
 kind: job_template
 name: Package Action - SSH Default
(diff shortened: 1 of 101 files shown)


ares:
Second commit also adds snippet attribute to snippets, normally exported by Foreman

ekohl (approved):

mmoll:
merged, díky @ares !

PR #447, opened 2018-01-11
Use os_family instead of os_name to verify if it's a Suse based OS

sbernhard:
In case enable-puppetlabs-X-repo is set we need to adjust the path.


ekohl (approved):

mmoll:
AIO packages are only available for SLES, the change was made intentionally back then.

sbernhard:
Good your point. Do you think, that the SLES packages would work on OpenSUSE, too?

mmoll:
Maybe. But even if that's the case I wouldn't want to pull this into the canonical upstream templates.

mmoll:
Closing for now, if there are OpenSUSE AIO RPMs or guarantees about binary compatibilties between SLES and OpenSUSE, this can get reworked.

PR #448, opened 2018-01-19
Extracted built command into snippet

lzap:
Yeah, this does exactly what's in the title. Beware! The snippet assumes there is a shell (note "or" syntax). There are few more wget/curl leftovers when it was obvious there is either no shell or it was not possible to do multiline replacement.

*Please* verify all instances if shell is allowed, looks to me it is the case.


provisioning_templates/snippet/built.erb
```diff
@@ -0,0 +1,10 @@
+<%#
+kind: snippet
+name: built
+model: ProvisioningTemplate
+snippet: true
+-%>
+# exit build mode (try with curl, wget and busybox)
+/usr/bin/curl -o /dev/null --insecure --silent '<%= foreman_url('built') %>' || \
```
ares: what happens if user has both curl and wget installed? would it be better to first detect the available tool? something like https://github.com/theforeman/foreman_chef/blob/master/app/views/foreman/unattended/snippets/_chef_client_omnibus_bootstrap.erb#L8-L12

and even better it would be to have it in core renderer :-)

ares:
since some calls contain `--insecure` args, it would be good to check whether the URL contains https or http as it may cause problems on http paths

provisioning_templates/snippet/built.erb
lzap: Added detection. I see no reason polluting core with such a powerful ERB system we have.

lzap:
But what is added benefit of having more complex code and one extra option which does not hurt?

I rebased and added `--no-proxy` options.

provisioning_templates/snippet/built.erb
ares: I have few, testing, updates (one day we'll want to add third tool), easier to call one macro in any template than writing these lines etc, note this is not the only place where we use wget/curl

ares:
I think the benefit is having support for curl that does not allow using --insecure with http urls. At least that was my understanding of reporters problem.

lzap:
Looks like you insist, amended the change you want.

mmoll:
@lzap please rebase :)

lzap:
And done!

lzap:
Guys...

provisioning_templates/snippet/built.erb
```diff
@@ -0,0 +1,27 @@
...
+-%>
+<%#
+  Sends 'built' request to exit build mode. Only use in multi-line shell context.
+  Tries curl, if not found then wget and finally wget from busybox (limited options).
+-%>
+<%
+  url = foreman_url('built')
+  if url.start_with?('https')
+    curl_opts = "--insecure"
+    wget_otps = "--no-check-certificate"
```
mmoll: `wget_opts` please ;)

lzap:
Rebased.

mmoll (approved):
Untested, but as it looks good and I don't have that much time to really test all the OSes anyway, I'd say, bring this in...

lzap:
@tbrisker I assume that 1.20 templates are already branched, can we merge this in?

tbrisker:
@lzap correct.

mmoll:
merged, díky @lzap!

PR #449, opened 2018-01-30
Kickstart: Add --no-proxy to informing Foreman

dgoetz:
I ran into the problem last week that Kickstart provisioning with proxy set does not inform Foreman when it is built as it tries to use the proxy also for this request. I think adding `--no-proxy` to it is ok as having a proxy between the provisioned system and Foreman is not a common scenario and probably will not work at all. We could also add another parameter to not break such a scenario, but I would like to avoid this for simplicity. I am not sure if other operating systems would benefit from a similar fix.

### DIFF
--- a/provisioning_templates/provision/kickstart_default.erb
+++ b/provisioning_templates/provision/kickstart_default.erb
@@ -199,7 +199,7 @@ sync
 
 # Inform the build system that we are done.
 echo "Informing Foreman that we are built"
-wget -q -O /dev/null --no-check-certificate <%= foreman_url('built') %>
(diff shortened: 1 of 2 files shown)


ekohl:
Related PR: https://github.com/theforeman/community-templates/pull/448

lzap:
I don't understand. When you set proxy in kickstart then you obviously want all utilities from the live system to use it.

In any case, don't merge I will incorporate the change into my refactoring ticket #448 if we agree it should be there.

dgoetz:
Typically you want to use no_proxy exceptions for local addresses like Foreman would be one which you can not set in kickstart as far as I know.

lzap:
That's right, I have put this into my PR #448

PR #451, opened 2018-02-03
Update CoreOS Container Linux PXE template

dLobatog:
As you can see in https://coreos.com/os/docs/latest/booting-with-pxe.html

An update to the provisioning template should also follow, so that it uses ignition (or ct for transpiling the yaml)... any preferences @timogoebel ?

### DIFF
--- a/provisioning_templates/PXELinux/coreos_pxelinux.erb
+++ b/provisioning_templates/PXELinux/coreos_pxelinux.erb
@@ -6,7 +6,13 @@ oses:
 - CoreOS
 %>
 DEFAULT coreos
+PROMPT 1
+TIMEOUT 15
+
+DISPLAY boot.msg
 
 LABEL coreos
-    KERNEL <%= @kernel %>
-    APPEND initrd=<%= @initrd %> cloud-config-url=<%= foreman_url('provision')%>
+  MENU default
+  KERNEL <%= @kernel %>
+  INITRD <%= @initrd %>
+  APPEND coreos.first_boot=1 coreos.config.url=<%= foreman_url('provision') %>


timogoebel:
@dLobatog: Thanks. Ignition is not meant to be written directly in JSON, but instead, they use a [transpiler](https://github.com/coreos/container-linux-config-transpiler) to transpile YAML into JSON. The transpile step also does validations, so you have a first safeguard against errors.
I have developed a [ruby wrapper](https://github.com/theforeman/foreman_omaha/blob/5afd7ef958462cc396c9e7c1ef7cf7ef5e5fbe95/app/services/foreman_omaha/container_linux_config_transpiler.rb#L4) around the transpiler which I think makes sense to support in Foreman core. It just raises the question if we want to package `ct` with Foreman or not. I think it would be great if you can write the template entirely in YAML and Foreman core can transpile it transparently into JSON. We should decide on whether we want to do the transpile step transparently (e.g. without the user having to call a method `transpile_container_linux_config`) in its template or explicitly with such a template macro.

lzap:
What a great idea @timogoebel I think this could be valuable feature in core if we generalize it. I can imagine that provisioning template would have a field "external converter" and when provided render would go into this utility. This could be leveraged for this purpose and we could push the YAML format into community-templates but this opens doors to other things.

I can imagine running ksvalidate or other tools at render time so we get much more detailed errors on the server. The contract could be very simple:

* script or binary (full path)
* arbitrary amount of options (I'd prefer execing without shell to prevent execution attacks)
* converter is ignored if binary is not found (not executable, no permissions, SELinux blocks it) - warning is logged
* stdin - rendered template input
* stdout - the result
* stderr - errors printed to foreman logs/audit
* exit code - obvious

timogoebel:
@lzap: Do you have an idea how we could run these tools in a "safe mode" environment? `ct` allows you to load files from disk, in theory, someone could load `/etc/foreman/database.yml` and the database password could be leaked.

cc: @ares

ares:
few quick ideas: one-shot container? tmp dir with chroot, we'd need to put dependencies into it? REX job so you can offload it to other system (ks validator on fedora, debconf-set-selections on debian)

I see the existing redmine tickets btw at https://projects.theforeman.org/issues/3873 or https://projects.theforeman.org/issues/2310, validation should happen on demand, perhaps when user clicks preview button and selects the host

lzap:
Good point, chroot with extra SELinux (when possible) could do it. Maybe running it via Smart Proxy so you can easily offload it to separate VM outside of network. ReX is a good idea, but we'd perhaps need to pre-render the template the moment host enters build mode, we don't want it to render it when unattended controller triggers that's easy DoS. Even spawning small container is doable perhaps via systemd-nspawn (minimum dependencies) and interesting idea. Alternative (crazy) idea - have it compiled into javascript (there are few Go to JS compilers).

I kinda like this scenario:

* having a smart-proxy module to be able to offload it to separate proxy
* proxy spawning the tool in chroot or container
* SELinux policy around it
* pre-render template when host enters build mode

mmoll:
@dLobatog @timogoebel I guess this is still relevant and OK to get merged?

timogoebel:
@mmoll: Imho this still needs the ignition part to work correctly.

provisioning_templates/PXELinux/coreos_pxelinux.erb
```diff
@@ -6,7 +6,13 @@ oses:
 - CoreOS
 %>
 DEFAULT coreos
+PROMPT 1
+TIMEOUT 15
+
+DISPLAY boot.msg
```
lzap, line 12: This `boot.msg` file, is it part of a different patch in our installer? Because it won't be present by default. I think you may not want to include this command here.

tbrisker:
This PR has been inactive for a long time now, closing. Feel free to reopen it if it is still relevant. thanks @dLobatog !

PR #452, opened 2018-02-09
Add Ansible templates

dLobatog:
The templates included here simply mimic the equivalent SSH one. In
addition to that, there are 2 templates to import roles (from galaxy or
git)

cc @ares

### DIFF
--- a/job_templates/ansible_roles_-_ansible_default.erb
+++ b/job_templates/ansible_roles_-_ansible_default.erb
@@ -1,0 +1,20 @@
+<%#
+name: Ansible Roles - Ansible Default
+job_category: Ansible Playbook
+description_format: Run Ansible roles
+snippet: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - name: Display all parameters known for the Foreman host
+      debug:
+        var: foreman_params
+  roles:
+<%- if @host.all_ansible_roles.present? -%>
+<%=   @host.all_ansible_roles.map { |role| "    - #{role.name.strip}" }.join("\n") %>
+<%- end -%>
--- a/job_templates/ansible_roles_-_install_from_galaxy.erb
+++ b/job_templates/ansible_roles_-_install_from_galaxy.erb
@@ -1,0 +1,32 @@
+<%#
+name: Ansible Roles - Install from Galaxy
+job_category: Ansible Galaxy
+description_format: Install roles '%{ansible_roles_list}' from Galaxy
+snippet: false
+template_inputs:
+- name: ansible_roles_list
+  required: true
+  input_type: user
+  description: "List of roles in Ansible Galaxy to install, separated by commas, e.g:
+    \"mysql, nginx\"\r\n\r\nThe default location is the 'roles_path' configured on
+    /etc/ansible/ansible.cfg, you may override it by filling the 'location' input.
+    Click on \"Advanced\" to see it."
+  advanced: false
+- name: location
+  required: false
+  input_type: user
+  description: A particular directory where you want the downloaded roles to be placed.
+  advanced: true
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: ansible-galaxy install {{ item }} <% "-p #{input('location')}" if input('location').present? %>
+      register: out
+      with_items:
+        - <%= input('ansible_roles_list') %>
+    - debug: var=out
--- a/job_templates/ansible_roles_-_install_from_git.erb
+++ b/job_templates/ansible_roles_-_install_from_git.erb
@@ -1,0 +1,28 @@
+<%#
+name: Ansible Roles - Install from git
+job_category: Ansible Roles Installation
+description_format: Clone roles from git repository to %{location}
+snippet: false
+template_inputs:
+- name: git_repository
+  required: true
+  input_type: user
+  description: "URL to the git repository containing the roles, e.g:\r\nhttps://github.com/theforeman/foreman_role_1"
+  advanced: false
+- name: location
+  required: true
+  input_type: user
+  description: For example, '/etc/ansible/roles/foobar' . Look at '/etc/ansible/ansible.cfg'
+    roles_path option to find what is your roles_path
+  advanced: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: git clone <%= input('git_repository') %> <%= input('location') %>
+      register: out
+    - debug: var=out
--- a/job_templates/package_action_-_ansible_default.erb
+++ b/job_templates/package_action_-_ansible_default.erb
@@ -1,0 +1,45 @@
+<%#
+name: Package Action - Ansible Default
+job_category: Ansible Packages
+description_format: 'Package %{name}: %{state}'
+snippet: false
+template_inputs:
+- name: state
+  required: true
+  input_type: user
+  description: Whether to install (present, latest), or remove (absent) a package.
+  options: "present\r\nabsent\r\nlatest"
+  advanced: false
+- name: name
+  required: true
+  input_type: user
+  description: "Package name, or package specifier with version, like name-1.0.\r\nBe
+    aware that packages are not always named the same and this module will not 'translate'
+    them per distro."
+  advanced: false
+- name: pre_script
+  description: A script to run prior to the package action
+  input_type: user
+  required: false
+  advanced: true
+- name: post_script
+  description: A script to run after the package action
+  input_type: user
+  required: false
+  advanced: true
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+# For Windows targets use the win_package module instead.
+---
+- hosts: all
+  pre_tasks:
+    - shell: "<%= input('pre_script') %>"
+  tasks:
+    - package:
+        name: <%= input('name') %>
+        state: <%= input('state') %>
+  post_tasks:
+    - shell: "<%= input('post_script') %>"
--- a/job_templates/power_action_-_ansible_default.erb
+++ b/job_templates/power_action_-_ansible_default.erb
@@ -1,0 +1,28 @@
+<%#
+name: Power Action - Ansible Default
+job_category: Power
+description_format: "%{action} host"
+snippet: false
+template_inputs:
+- name: action
+  required: true
+  input_type: user
+  description: Action to perform on the service
+  options: "restart\r\nshutdown"
+  advanced: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: |
+        echo <%= input('action') %> host && sleep 3
+        <%= case input('action')
+          when 'restart'
+            'reboot'
+          else
+            'shutdown -h now'
+          end %>
--- a/job_templates/puppet_run_once_-_ansible_default.erb
+++ b/job_templates/puppet_run_once_-_ansible_default.erb
@@ -1,0 +1,21 @@
+<%#
+name: Puppet Run Once - Ansible Default
+job_category: Puppet
+description_format: 'Run Puppet once with "%{puppet_options}"'
+snippet: false
+template_inputs:
+- name: puppet_options
+  required: false
+  input_type: user
+  description: Additional options to pass to Puppet
+  advanced: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: |
+        puppet agent --onetime --no-usecacheonfailure --no-daemonize <%= input("puppet_options") %>
--- a/job_templates/run_command_-_ansible_default.erb
+++ b/job_templates/run_command_-_ansible_default.erb
@@ -1,0 +1,22 @@
+<%#
+name: Run Command - Ansible Default
+job_category: Ansible Commands
+description_format: Run %{command}
+snippet: false
+template_inputs:
+- name: command
+  required: true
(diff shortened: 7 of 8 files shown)


job_templates/ansible_roles_-_ansible_default.erb
```diff
@@ -0,0 +1,18 @@
...
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - debug: msg="Parameters provided: {{ foreman_params }}"
+  roles:
+    <% if @host.all_ansible_roles.present? -%>
+    - <%= @host.all_ansible_roles.map { |role| role.name.strip }.join("\n    - ") %>
```
ares: since you're already doing map, how about `map { |role| "    - #{role.name.strip}" }.join("\n")` to avoid prefix duplication? then you could also write it without hardcoding whitespaces into erb, see https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/kickstart_networking_setup.erb for example of what I mean

job_templates/service_action_-_ansible_default.erb
```diff
@@ -0,0 +1,61 @@
...
+---
+- hosts: all
+  tasks:
+  - service:
+      name: "{{ name }}"
+      state: "{{ state }}"
+      <% if input('enabled').present? %>enabled: "{{ enabled }}"<% end %>
+      <% if input('args').present? %>args: "{{ args }}"<% end %>
+      <% if input('pattern').present? %>pattern: "{{ pattern }}"<% end %>
+      <% if input('sleep').present? %>sleep: "{{ sleep }}"<% end %>
```
ares: if we are using `<% input` anyway, why it's combined with `{{ args``? the input macro can be previewed, isn't it better to use it in templates we ship especially if they already use erb?

ares (commented):
2 comments before starting to test

job_templates/service_action_-_ansible_default.erb
dLobatog: The point was to try to show to Ansible users how they can use `{{ }}` but I'll change it

job_templates/ansible_roles_-_ansible_default.erb
```diff
@@ -0,0 +1,18 @@
+<%#
```
ares, line 1: for this template it seems to be rendered as

```
---
- hosts: all
  tasks:
    - debug: msg="Parameters provided: {{ foreman_params }}"
  roles:
        - geerlingguy.ntp
```

note the extra whitespaces

and then when executing it, I'm getting
```
  1:
ERROR! Syntax Error while loading YAML.
   2:
   3:
   4:
The error appears to have been in '/tmp/d20180215-18251-1hpmmut/foreman-playbook-fb6eafb9-ea4c-44ac-8069-4bcd3aa25a98.yml': line 4, column 38, but may
   5:
be elsewhere in the file depending on the exact syntax problem.
   6:
   7:
The offending line appears to be:
   8:
   9:
tasks:
  10:
- debug: msg="Parameters provided: {{ foreman_params }}"
  11:
^ here
  12:
We could be wrong, but this one looks like it might be an issue with
  13:
missing quotes. Always quote template expression brackets when they
  14:
start a value. For instance:
  15:
  16:
with_items:
  17:
- {{ foo }}
  18:
  19:
Should be written as:
  20:
  21:
with_items:
  22:
- "{{ foo }}"
  23:
  24:
exception type: <class 'yaml.scanner.ScannerError'>
  25:
exception: mapping values are not allowed in this context
  26:
in "<unicode string>", line 4, column 38
  27:
Exit status: 4
```

job_templates/ansible_roles_-_install_from_git.erb
```diff
@@ -0,0 +1,29 @@
+<%#
```
ares, line 1: this does not work, the input is called git_repositories_list while the template relies on {{ git_repositories }}

```
   1:
PLAY [all] *********************************************************************
   2:
   3:
TASK [Gathering Facts] *********************************************************
   4:
ok: [abel-debolt.example.tst]
   5:
[WARNING]: Failure using method (v2_runner_on_ok) in callback plugin
   6:
(<ansible.plugins.callback.foreman.CallbackModule object at 0x7f167162db10>):
   7:
Could not find the TLS certificate file, invalid path:
   8:
/etc/foreman/client_cert.pem
   9:
  10:
  11:
TASK [command] *****************************************************************
  12:
fatal: [abel-debolt.example.tst]: FAILED! => {"msg": "'git_repositories' is undefined"}
  13:
to retry, use: --limit @/tmp/d20180215-18251-hfs6cv/foreman-playbook-d7142be5-130b-4a26-9da3-d15006c157ec.retry
  14:
  15:
PLAY RECAP *********************************************************************
  16:
abel-debolt.example.tst : ok=1 changed=0 unreachable=0 failed=1
  17:
  18:
[WARNING]: Failure using method (v2_playbook_on_stats) in callback plugin
  19:
(<ansible.plugins.callback.foreman.CallbackModule object at 0x7f167162db10>):
  20:
Could not find the TLS certificate file, invalid path:
  21:
/etc/foreman/client_cert.pem
  22:
  23:
Exit status: 2
```

job_templates/ansible_roles_-_install_from_galaxy.erb
```diff
@@ -0,0 +1,23 @@
+<%#
```
ares, line 1: It installs the role without namespace (probably not fault of this template) so that geerlinguy.ntp ends up imported as ntp.  This can cause problems if you sync two roles with same name from different repos, anything you can think of to keep the namespace?

job_templates/puppet_run_once_-_ansible_default.erb
```diff
@@ -0,0 +1,21 @@
+<%#
+name: Puppet Run Once - Ansible Default
+job_category: Puppet
+description_format: "%{action} host"
+snippet: false
+template_inputs:
+- name: puppet_options
+  required: false
+  input_type: user
+  options: Additional options to pass to Puppet
```
ares: this is not a valid option, this should have been description I guess, options should not be set as it's free text input

job_templates/run_command_-_ansible_default.erb
```diff
@@ -0,0 +1,21 @@
...
+  advanced: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: "{{ command }}"
```
ares: here I think it would be extremely useful to use <%= input... %> so you can see the command in job invocation page and in preview, the new job invocation UI will make it more obvious but with current UI, it would be a huge win I think

ares (changes requested):
Awesome stuff! Could you please also add install package template?

job_templates/ansible_roles_-_install_from_galaxy.erb
dLobatog: I don't see that, actually I see the namespace.role in /etc/ansible/roles, maybe you have an older version of ansible-galaxy/ansible?

job_templates/ansible_roles_-_install_from_git.erb
dLobatog: Should be fixed now

dLobatog (commented):
@ares I've updated this according to your comments, and added the `Package action` template.

job_templates/ansible_roles_-_ansible_default.erb
```diff
@@ -0,0 +1,20 @@
...
+%>
+
+---
+- hosts: all
+  tasks:
+    - name: Display all parameters known for the Foreman host
+      debug:
+        var: foreman_params
+  roles:
+<% if @host.all_ansible_roles.present? -%>
```
ares: if you add - in opening erb tag, you'll get same lenght of erb tag and have code aligned to even number of whitespaces and then also the line @host.all_ansible ... can be properly aligned

```erb
<%- if @host.all_ansible_roles.present? -%>
<%=   @host.all_ansible_roles.map { |role| "    - #{role.name.strip}" }.join("\n") %>
<%- end %>
```

job_templates/ansible_roles_-_install_from_git.erb
ares: I see this has been now changed to single resource (name and example in description) but the template still accepts multiple items, what is the desired usage? If it allows multiple repos, should the input name be plural?

dLobatog (commented):
@ares Updated the Ansible Roles template so that the ERB is aligned.

job_templates/ansible_roles_-_install_from_git.erb
dLobatog: It only works with a single repo. The reason is that you can't simply `git clone myrepo /etc/ansible/roles`. That returns an error `fatal: destination path '/etc/ansible/roles' already exists and is not an empty directory`. The location must be something like `/etc/ansible/roles/something` so that we're able to clone.

I wondered how to do this parsing the input, but as an example:

1. `git_repositories = puppet-foreman
openshift-ansible
kubernetes-ansible`
2. There's only one location that can be specified so far, what should it be? `location = /etc/ansible/roles/openshift`
3. Since there's only one location, it doesn't make sense for all repositories to use the same one.

I'm just realizing we could use:

```yaml
  with_items:
    - { git_repository: 'somerepo', location: 'somelocation' }
    - { git_repository: 'somerepo', location: 'somelocation' }
```

and build this structure with 2 inputs that accept multiple values.

dLobatog (commented):
@ares I've updated the description of the `install from git` PR to be more clear about it's use - only 1 repo.

dLobatog:
@ares I wonder if this can be merged with `install from git repository` installing only from 1 repo, so we can pull this into a release and then work on the multiple repos template?

ares:
That is fine, but in that case, I'd remove `with_items` from the template and use git repository parameter directly

dLobatog:
@ares Updated to use the `git_repository` input directly instead of `with_items.`

job_templates/ansible_roles_-_install_from_git.erb
```diff
@@ -0,0 +1,28 @@
...
+  advanced: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: git clone <%= input('git_repository') %> <%= input('location') %>
```
ares, line 26: just curious, not necessary to fix in this PR, is there a reason why we don't use [native git module ](http://docs.ansible.com/ansible/latest/git_module.html) and fallback to shell command?

job_templates/ansible_roles_-_install_from_galaxy.erb
ares: interesting, does not happen on clean dev env, please ignore this :-)

job_templates/ansible_roles_-_install_from_git.erb
dLobatog: Actually no reason but ignorance (I just discovered it), I'm not sure how that'd work for multiple items which is what I had in mind originally

job_templates/power_action_-_ansible_default.erb
```diff
@@ -0,0 +1,28 @@
...
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: |
+        echo <%= input('action') %> host && sleep 3
+        <%= case input('action')
+          when 'restart'
+            'reboot'
```
ekohl, line 25: You could also use ```shutdown -r now``` for more consistency with halting.

dLobatog: @ekohl It's less consistent with the equivalent SSH template - what about changing that in both templates after it's merged if it's important?

job_templates/puppet_run_once_-_ansible_default.erb
```diff
@@ -0,0 +1,21 @@
+<%#
+name: Puppet Run Once - Ansible Default
+job_category: Puppet
+description_format: "%{action} host"
```
ares: action is not interpolated, I guess it was copied from other template?

job_templates/package_action_-_ansible_default.erb
```diff
@@ -0,0 +1,31 @@
+<%#
+name: Package Action - Ansible Default
+job_category: Ansible Packages
+description_format: 'Package %{name}: %{state}'
+snippet: false
+template_inputs:
```
ares, line 6: the ssh template has two additional advanced inputs - pre_script and post_script, this was very asked feature, could you add theme here too? they would result in two commands that would be executed only if passed

ares (changes requested):
last two small things, otherwise good to to be merged, I'd leave things like mutliple roles and shutdown refactoring for separate PRs

dLobatog (commented):
@ares Updated to fix 'Puppet Run Once description' and add "pre_script" and "post_script" to the 'Package Action" playbook.

ares:
one last thing, pre and post script do not work if they contain multiple commands because yaml format gets broken, it needs to be wrapped in quotes, this seems to work https://gist.github.com/ares/d8db21191dcaab497c913c0e38a24d35

then good to merge

job_templates/power_action_-_ansible_default.erb
ekohl: Consistency with other templates is more important, but I think systemd wants to deprecate ```halt``` and ```reboot``` in favor of ```shutdown``` so it might be good to pick this up in a separate PR.

dLobatog:
@ares Updated so that it works with multiple commands & rebased the PR so that it's only one commit.

ares (approved):

ares:
Awesome work, thanks @dLobatog and also thanks @ekohl for second pair of eyes. Merging!

job_templates/power_action_-_ansible_default.erb
ekohl: Looks like I was not the only one: https://github.com/theforeman/community-templates/pull/456

PR #454, opened 2018-02-15
Chainbooting discovery via iPXE

lzap:
We were working with a customer to boot discovery via iPXE and this is the result. This template needs the iPXE to be put into public html directory, but it can be still a useful item to have.


mmoll:
merged, díky @lzap!

PR #456, opened 2018-02-24
Fixes #22679 - job template for reboot to use shutdown -r +1

pmoravec:
Signed-off-by: Pavel Moravec <pmoravec@redhat.com>

### DIFF
--- a/job_templates/power_action_-_ssh_default.erb
+++ b/job_templates/power_action_-_ssh_default.erb
@@ -16,7 +16,7 @@ template_inputs:
 echo <%= input('action') %> host && sleep 3
 <%= case input('action')
       when 'restart'
-        'reboot'
+        'shutdown -r +1'
       else
         'shutdown -h now'
       end %>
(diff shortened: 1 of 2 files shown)


job_templates/power_action_-_ssh_default.erb
```diff
@@ -16,7 +16,7 @@ template_inputs:
 echo <%= input('action') %> host && sleep 3
 <%= case input('action')
       when 'restart'
-        'reboot'
+        'shutdown -r +1'
```
ekohl, line 19: How does +1 compare to now?

ekohl:
Could you also change https://github.com/theforeman/community-templates/blob/fdae1ff37493e8ce794720c841ac296a99a2ed52/job_templates/power_action_-_ansible_default.erb#L25

job_templates/power_action_-_ssh_default.erb
pmoravec: "shutdown -r now" is basically the same like "reboot" - either command can trigger shutting down network *before* the command terminates, causing the "ssh" command from remote server will not receive any return value and will be waiting for it frever.

pmoravec:
updated the ansible job template accordingly.

dLobatog (approved):

iNecas (approved):

iNecas:
Tested on centos7 and rhel5, works well on both.

dLobatog (commented):
Thanks @pmoravec !

PR #457, opened 2018-02-26
Add feature to default roles playbook

dLobatog:
This playbook is used for the ansible_run_host. Currently I'm adding this line on every sync, but it should not be necessary.


ares:
:+1: thanks @dLobatog, merging

PR #459, opened 2018-03-02
Katello REX features through Ansible

dLobatog:
Using the `package` or the `yum` module here is *not* possible except for the "Install Package" "Remove Package" templates.

* 'group install' only takes packages in the format '@something' 
* there is no update-minimal, group remove, group update.. https://github.com/ansible/ansible/issues/11498


job_templates/install_package_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,16 @@
...
+template_inputs:
+- name: package
+  required: true
+  input_type: user
+  advanced: false
+provider_type: Ansible
+kind: job_template
+%>
+
+<%= render_template('Package Action - SSH Default', :state => 'latest', :name => input('package')) %>
```
ares: should this use Package Action - Ansible Default?

ares: btw it's unrelated but `Package Action - Ansible Default` seems to be broken, if pre and post script inputs are empty, ansible fails with traceback like 

```
{"changed": false, "module_stderr": "Shared connection to dell-pet410....com closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n File \"/tmp/ansible_ebIAgm/ansible_module_command.py\", line 213, in <module>\r\n main()\r\n File \"/tmp/ansible_ebIAgm/ansible_module_command.py\", line 147, in main\r\n if args.strip() == '':\r\nAttributeError: 'NoneType' object has no attribute 'strip'\r\n", "msg": "MODULE FAILURE", "rc": 0}
```

job_templates/install_errata_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,18 @@
+<%#
+name: Install Errata - Katello Ansible Default
+snippet: false
+model: JobTemplate
+job_category: Katello
```
ares: since REX won't be able to deal with multiple providers for the same job category, I'm afraid we'll need to create separate category called `Katello via Ansible` for now, otherwise all these job template would only be usable through Katello features.

job_templates/install_group_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,16 @@
...
+template_inputs:
+- name: package
+  required: true
+  input_type: user
+  advanced: false
+provider_type: Ansible
+kind: job_template
+%>
+
+<%= render_template('Run Command - Ansible Default', :command => "yum group install #{input('package')}") %>
```
ares: many files does not have EOF character, it should be added I think

job_templates/restart_services_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,18 @@
...
+- name: helper
+  required: true
+  input_type: user
+  description: A comma separated list of commands to run to restart services
+  advanced: false
+provider_type: Ansible
+kind: job_template
+%>
+
+<%= render_template('Run Command - Ansible Default', :command => input(:helper).split(",").map {|split| "#{split}" }.join("\n")) %>
```
ares: this seems to produce invalid yaml as the run command contains `---` as the first line, result is (note --- in the middle)
```
---
- hosts: all
  tasks:
    - command: goferd
      register: out
    - debug: var=out
---
- hosts: all
  tasks:
    - command: katello-tracer-upload
      register: out
    - debug: var=out
```

ares: either it should be combined otherwise or this should be removed https://github.com/theforeman/community-templates/blob/develop/job_templates/run_command_-_ansible_default.erb#L17

ares (changes requested):
I left few comments, :+1: in general

job_templates/install_group_-_katello_ansible_default.erb
dLobatog: Sorry, I'm not sure what do you mean here? What do EOF characters have to do with `yum group install`?

ares: nothing :-) just that EOL was missing and this was last line in the file

dLobatog (commented):
@ares Any more changes I should be making here? There's a separate PR to fix the pre/post script problem when installing a package you mentioned in a comment.

ares:
code wise looks good, let me retest, thanks!

job_templates/restart_services_-_katello_ansible_default.erb
```diff
@@ -0,0 +1,17 @@
...
+- name: helper
+  required: true
+  input_type: user
+  description: A comma separated list of commands to run to restart services
+  advanced: false
+provider_type: Ansible
+kind: job_template
+%>
+
+<%= render_template('Run Command - Ansible Default', :command => "#{input(:helper).split(",")},katello-tracer-upload".map {|split| "#{split}" }.join("\n")) %>
```
ares: getting undefined method map on string

ares (changes requested):
found one problem with the service restart, also needed to apply https://github.com/theforeman/community-templates/pull/469 in order for package action to work

dLobatog (commented):
@ares Updated, it was calling 'split' too early instead of right before the `.map`

ares:
This still does not work because of multilines are not properly formatted by run command template. here's an example with "echo hello" input

```yaml
# For Windows targets use the win_package module instead.
---
- hosts: localhost
  tasks:
    - command: echo "test"
katello-tracer-upload
      register: out
    - debug: var=out
```

I opened a PR that improves the command at https://github.com/theforeman/community-templates/pull/472, we can merge this after that one is in

ares (approved):
Thanks, all seems working now :-)

PR #450, opened 2018-02-02
Fixes #15634 - RancherOS support

selmison:

PR #464, opened 2018-03-22
Fixes #15634 - RancherOS support

zozoh94:
I improved Semilson template to fix his own pull request : [](https://github.com/theforeman/community-templates/pull/450)

### DIFF
--- a/provisioning_templates/PXELinux/rancheros_pxelinux.erb
+++ b/provisioning_templates/PXELinux/rancheros_pxelinux.erb
@@ -1,0 +1,12 @@
+<%#
+kind: PXELinux
+name: RancherOS PXELinux
+model: ProvisioningTemplate
+oses:
+- RancherOS
+-%>
+DEFAULT rancheros
+
+LABEL rancheros
+  KERNEL <%= @kernel -%> rancher.state.dev=LABEL=RANCHER_STATE rancher.state.autoformat=[<%= host_param('install-disk') || '/dev/sda' %>] rancher.cloud_init.datasources=['url:<%= foreman_url('provision')-%>']
+  APPEND initrd=<%= @initrd -%>
--- a/provisioning_templates/snippet/rancheros_cloudconfig.erb
+++ b/provisioning_templates/snippet/rancheros_cloudconfig.erb
@@ -1,0 +1,52 @@
+<%#
+kind: snippet
+name: rancheros_cloudconfig
+model: ProvisioningTemplate
+snippet: true
+-%>
+#cloud-config
+hostname: <%= @host.shortname %>
+rancher:
+  network:
+    dns:
+      nameservers:
+<% if @host.subnet.dns_primary.present? or @host.subnet.dns_secondary.present? -%>
+<% if @host.subnet.dns_primary.present? -%>
+        - <%= @host.subnet.dns_primary %>
+<% end -%>
+<% if @host.subnet.dns_secondary.present? -%>
+        - <%= @host.subnet.dns_secondary %>
+<% end -%>
+<% end -%>
+      search:
+        - <%= @host.domain %>
+    interfaces:
+      eth*:
+        dhcp: true
+<% unless @host.subnet.respond_to?(:dhcp_boot_mode?) -%>
+      eth0:
+        address: <%= @host.ip -%>/<%= @host.subnet.cidr %>
+        gateway: <%= @host.subnet.gateway %>
+<% end -%>
+  console: debian
+ssh_authorized_keys:
+<%-
+users = @host.owner_type == 'Usergroup' ? @host.owner.all_users : [@host.owner]
+users.each do |user|
+  if user.respond_to?(:ssh_authorized_keys) && user.ssh_authorized_keys.any?
+    index = 0
+    user.ssh_keys.each do |key|
+      if index == 0
+-%>
+<%="  - #{key.type} #{key.ssh_key} #{key.comment}"%>
+<%-
+      else
+-%>
+<%="  - #{key.type} #{key.ssh_key} #{key.comment} - #{index}"%>
+<%-
+      end
+      index += 1
+    end
+  end
+end
+-%>
--- a/partition_tables_templates/empty.erb
+++ b/partition_tables_templates/empty.erb
@@ -1,0 +1,7 @@
+<%#
+kind: ptable
+name: Empty
+model: Ptable
+oses:
+- RancherOS
+-%>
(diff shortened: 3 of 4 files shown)


partition_tables_templates/rancheros_default_fake.erb
```diff
@@ -0,0 +1,8 @@
+<%#
+kind: ptable
+name: RancherOS default fake
+model: Ptable
+oses:
+- RancherOS
+%>
+# Not required for RancherOS provisioning.
```
lzap: This might be useful for the future, can you rename the file and rename it to simply "Empty". Then keep the RancherOS association, but do not put any text, it must render to empty string. To make sure there is not even a newline add the minus sign in the end:

```erb
<%#
kind: ptable
name: Empty
model: Ptable
oses:
 - RancherOS
-%>
```

provisioning_templates/PXELinux/rancheros_pxelinux.erb
```diff
@@ -0,0 +1,12 @@
...
+name: RancherOS PXELinux
+model: ProvisioningTemplate
+oses:
+- RancherOS
+-%>
+DEFAULT rancheros
+
+LABEL rancheros
+  KERNEL <%= @kernel -%> rancher.password='changeme' rancher.state.dev=LABEL=RANCHER_STATE rancher.state.autoformat=[/dev/xvda] rancher.cloud_init.datasources=['url:<%= foreman_url('provision')-%>']
+  APPEND initrd=<%= @initrd -%>
```
lzap: Foreman can provide root password via `<%= root_pass %>` but it will be in Linux crypt form (`$5$xxxx$yyyyyyy`), not sure if RancherOS accepts it in this form. Putting password via plaintext is something you do not want to do, this file is word-readable via TFTP! I suggest to delete whole `rancher.password='changeme'` part if the OS boots without it.

Device should be parametrized to make this template usable, your will only work for Xen, use this: `<%= host_param('autoformat_device') %>` or similar.

provisioning_templates/snippet/rancheros_cloudconfig.erb
```diff
@@ -0,0 +1,52 @@
+#cloud-config
```
lzap: Please put this line below the comment, it will work.

lzap (changes requested):
Couple of changes to have better default templates.

provisioning_templates/PXELinux/rancheros_pxelinux.erb
zozoh94: Rancher don't support crypted passwor I tried. So I will delete this parameter. I will be the responsability of the user to either have public keys confgured or add this parameter.

partition_tables_templates/rancheros_default_fake.erb
zozoh94: In a "empty.erb" file ?

provisioning_templates/PXELinux/rancheros_pxelinux.erb
lzap: Thanks, since it supports cloud-init this is not an issue actually.

provisioning_templates/PXELinux/rancheros_pxelinux.erb
```diff
@@ -0,0 +1,12 @@
...
+kind: PXELinux
+name: RancherOS PXELinux
+model: ProvisioningTemplate
+oses:
+- RancherOS
+-%>
+DEFAULT rancheros
+
+LABEL rancheros
+  KERNEL <%= @kernel -%> rancher.state.dev=LABEL=RANCHER_STATE rancher.state.autoformat=[/dev/xvda] rancher.cloud_init.datasources=['url:<%= foreman_url('provision')-%>']
```
lzap: The `/dev/xvda` is still unparametrized, can you add a param and default to this when not set?

lzap (changes requested):
Last comment about the hardcoded device name and we are good to go!

provisioning_templates/PXELinux/rancheros_pxelinux.erb
lzap: The line is a bit off now, can you add two leading spaces?

This would work as well: `<%= host_param('autoformat_device') || '/dev/vda' %>`

provisioning_templates/PXELinux/rancheros_pxelinux.erb
mmoll: The `install-disk` parameter should be re-used here, IMHO

provisioning_templates/PXELinux/rancheros_pxelinux.erb
mmoll: The `install-disk` parameter should be re-used here, IMHO

provisioning_templates/PXELinux/rancheros_pxelinux.erb
zozoh94: So install-disk or autoformat_device ?

mmoll: install-disk please, that variable is used in some other templates.

mmoll:
thanks, I try to test this over the weekend :+1:

lzap (approved):
Thanks I am not going to test this end to end, so if you don't get to it let's just merge.

zozoh94:
Wait before merging I will do some last tests to see if cloud-config work well.

lzap:
Thanks appreciated, let us know how it went. Preferably squash all commits into one before testing, so we all have the same SHA value.

zozoh94:
I merged commits in just one by rebasing called : Fixes #15634 - RancherOS support. I tried to put rancher password in cloud-config file a last time but unfortunately it seems impossible but everything works well with authorized_keys. I also figure out that there is no binary for wget inside RancherOS for template finishing but there is a curl one so I remplace the long wget command with "curl -sk". I also added a default autoformat_device which I forgot when I replaced autoformat_device param with install-disk.

So I think it's ok you can merge !

zozoh94:
Wait the finish template seems not working. I will find a fix.

zozoh94:
It's ok I delete the finish template because Rancher work with a provisioning each time there is a reboot. Rancher works on RAM and will just mount the device marked with the label RANCHER_STATE to keep docker images and persist cloud-config file.

mmoll:
superseded by GH-464

mmoll:
this is for sure not perfect, e.g. that eth* and eth0 is hardcoded, but it seems to work, so let's get this in and improve later :+1:

mmoll:
merged, thanks @zozoh94 and @selmison!

PR #465, opened 2018-03-28
Align regexp with puppet installer

lzap:
This is what our puppet installer set, so this patch aligns it. It is also shorter, so I'd prefer this version. Both should work anyway.


mmoll:
merged, díky @lzap!

PR #466, opened 2018-03-28
Add Katello support for RHVH/oVirt Node provisioning

iNecas:
Leverages `repository_url` helper to generate full url to
Katello-provided squashfs img, based on relative path.

Requires https://github.com/Katello/katello/pull/7215

The process I was following was:

1. extract the squashfs.img from the rhvh rpm
2. create custom product (name:rhv) and repository (content type: file, name:rhvh)
3. upload the squashfs.img to the repository (hammer repository upload-content --organization 'Default Organization' --name rhvh --product rhv --path ~/squashfs.img)
4. used 'custom/RHV/rhvh' as 'liveimg_content_path' host parameter.

With this patch (and the katello PR), this will lead to using http://katello.example.com/pulp/isos/Default_Organization/Library/custom/RHV/rhvh/squashfs.img for liveimg

### DIFF
--- a/provisioning_templates/provision/kickstart_ovirt.erb
+++ b/provisioning_templates/provision/kickstart_ovirt.erb
@@ -9,7 +9,6 @@ oses:
 <%#
 This template accepts the following parameters:
 - liveimg_name: string (default=squashfs.img)
-- liveimg_content_path: string - a relative path to the Katello-managed content
 - lang: string (default="en_US.UTF-8")
 - selinux-mode: string (default="enforcing")
 - keyboard: string (default="us")
@@ -23,15 +22,15 @@ root folder (or specified via 'liveimg_name' parameter). See oVirt Node document
 or RHV Installation Manual, section 5.2. Advanced installation.
 
 By default, the template expects the squashfs.img to be present inside
-the installation media. When using Katello for content management,
-it's possible to set 'liveimg_content_path' parameter, that is allows
-to specify relative path to the file and the template will generate an
-absolute path to the file (taking into account also info about the
-content proxy relevant for the host)
+the installation media. When using Katello for content management (the
+kt_activation_key parameter is set), the liveimg_name is used to
+specify relative path to the file and the template use repository_url
+helper to generate an absolute path to the file (taking into account
+also info about the content proxy relevant for the host)
 
 For example, in case the squashfs.img is uploaded inside custom
 product named 'oVirt' and repository 'hypervisor', the
-liveimg_content_path would be 'custom/ovirt/hypervisor/squashfs.img'.
+liveimg_name would be 'custom/ovirt/hypervisor/squashfs.img'.
 In this case, this repository would need to be part of the content
 view the host is assigned to. It's also possible to provide full url,
 in which case it would be used without a change.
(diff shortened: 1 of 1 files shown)


provisioning_templates/provision/kickstart_ovirt.erb
```diff
@@ -20,9 +21,32 @@ This kickstart will only work with LVM THIN partitioning ('Kickstart default thi
 and it requires the installation URL to have squashfs.img image extracted in the
 root folder (or specified via 'liveimg_name' parameter). See oVirt Node documentation
 or RHV Installation Manual, section 5.2. Advanced installation.
+
+By defult, the template expects the squashfs.img to be present inside
```
ghost: Small typo here

ghost (commented):
Small typo -- otherwise looks great

ghost:
@sandrobonazzola

lzap:
Can you merge `liveimg_content_path` and `liveimg_name` so there is just one parameter that works for both? They are both relative paths, the only difference should be base URL (IM or Katello CV) I believe. This way it is little bit confusing.

iNecas:
There is another difference, which is that the 'liveimg_content_path` also enforces using the subscription manager.

If I unified those two, I would need another parameter to specify, that the katello url should be used, instead of installation media.

If we want to eliminate the number of parameters, I would probably go for not exposing `liveimg_name` at all, and instead expecting that if using IM for this purpose, the IM is expecting the squashfs.img to be placed on recommended location.

provisioning_templates/provision/kickstart_ovirt.erb
iNecas: Updated

lzap:
Yes, that's another reason why to merge it, this is a side effect you don't want to always have. Imagine a Katello user doing CentOS CV provisioning, or even SUSE or Debian which is coming.

We still want to expose a plain URL because this will be useful in non-oVirt provisioning, you can create your own image/squashimg and provision normal OS using this technique, this is in Anaconda starting from RHEL 7.3 and it's a valid provisioning option. We want to have the same parameter also in normal kickstart, not only oVirt.

I am searching for consistency between katello and non-katello users and ovirt and non-ovirt users, please don't get me wrong.

iNecas:
Unifying `liveimg_content_path` and `liveimg_name` by itself is not enough. Are you suggesting to add additional parameter to determine if to use katello or not?

lzap:
There is already a condition there: `if @host.operatingsystem.name == 'RHVH'` it is kind of lame, so an explicit parameter might be a good idea for those who do not want to name OS like that.

lzap:
@iNecas this has been merged recently

https://github.com/theforeman/community-templates/pull/302/files#diff-62c51493c4aee87824ebe40d58fe4382R62

I wonder if this template can be designed in the same way, I think so. We can remove the RHVH name check and simply use the snippet, it will render to nothing when there is no rhsm available.

iNecas:
Good idea, updated

provisioning_templates/provision/kickstart_ovirt.erb
```diff
@@ -20,9 +20,33 @@ This kickstart will only work with LVM THIN partitioning ('Kickstart default thi
...
+view the host is assigned to. It's also possible to provide full url,
+in which case it would be used without a change.
 %>
+
 install
-liveimg --url=<%= @host.operatingsystem.medium_uri(@host) %>/<%= @host.params['liveimg_name'] || 'squashfs.img' %>
+<%
+liveimg_name = host_param('liveimg_name') || 'squashfs.img'
+if host_param('kt_activation_keys')
+  liveimg_url = repository_url(liveimg_name, 'isos')
```
lzap, line 43: Where repository_url helper is defined? Can't find it in katello codebase.

lzap (approved):
Anyway, I am fine with that. Thanks.

lzap:
Haven't tried the katello workflow, but I am good with this. TY

provisioning_templates/provision/kickstart_ovirt.erb
iNecas: I mentioned that in the first comment: https://github.com/Katello/katello/pull/7215 - it's close to merge.

mykaul:
@lzap - in which Foreman version will it be available?

lzap:
Hello @mykaul it's 1.19 but you can easily import fresh templates via foreman_templates plugin into any Foreman release.

ekohl:
Note we do have stable-x.y branches and we should be importing those into .z releases. You can create a PR against those branches so they'll be synced on the next release.

PR #467, opened 2018-03-30
Put Ansible Power & Puppet jobs into separate categories

bastilian:
Ansible jobs that share a category with Puppet jobs need their
own category to avoid an issue when invoking a job, where it
shows duplicate inputs for jobs template and parameters.

Before:
screenshot_2018-03-30_11-26-19

After:

screenshot_2018-03-30_12-24-01
screenshot_2018-03-30_12-24-17


ekohl (approved):

ares:
Thanks @bastilian, merging

PR #468, opened 2018-04-03
Ansible packages - Only run pre/post script when present

dLobatog:
@ares I think you discovered this one in https://github.com/theforeman/community-templates/pull/459

### DIFF
--- a/job_templates/package_action_-_ansible_default.erb
+++ b/job_templates/package_action_-_ansible_default.erb
@@ -35,11 +35,15 @@ model: JobTemplate
 # For Windows targets use the win_package module instead.
 ---
 - hosts: all
+  <% if input('pre_script').present? -%>
   pre_tasks:
     - shell: "<%= input('pre_script') %>"
+  <% end -%>
   tasks:
     - package:
         name: <%= input('name') %>
         state: <%= input('state') %>
+  <% if input('post_script').present? -%>
   post_tasks:
     - shell: "<%= input('post_script') %>"
+  <% end -%>


ekohl (approved):

mmoll:
merged, gracias @dLobatog!

job_templates/package_action_-_ansible_default.erb
```diff
@@ -35,11 +35,15 @@ model: JobTemplate
 # For Windows targets use the win_package module instead.
 ---
 - hosts: all
+  <% if input('pre_script').present? -%>
   pre_tasks:
     - shell: "<%= input('pre_script') %>"
+  <% end -%>
```
ares, line 41: actually this introduces a regression, the `-` at the end causes tasks not being well aligned. It's nested under hosts :-/

ares: the proper fix is to add `-` also to the begining so that it removes whitespaces before `<% end %>`, this is why it's better to always start <% at the beginning of the line and do the formatting inside erb, I'll send a fix in a minute

ares: see https://github.com/theforeman/community-templates/pull/469

mmoll: aw, sorry... I forgot how awful YAML is...

ekohl: I did think of that while looking at this PR, but not very well it appears.

PR #471, opened 2018-04-06
Translate ssh package actions for debian

mdellweg:
apt does not have the concept of package groups (there are very dependency-heavy meta packages though). So this translates all Group actions to being plain actions. Also it should not hurt to run `apt-get update` every time.
Also i tried to translate the os_family question to which package manager is being used. This should help as soon as dnf enters the arena.


ekohl (approved):
I like the os and package manager separation.

mdellweg:
@ekohl any update on this one?

mdellweg:
ping @ekohl, @mmoll, @lzap 
This is getting a little bit stale.

ekohl:
Thanks! Apparently I had this still marked as TODO but that list is (too) long. Merged and I'll be happy to accept further PRs if there's a bug we missed :)

mdellweg:
Thanks @ekohl !

PR #472, opened 2018-04-06
Support multiline scripts

ares:


dLobatog (approved):
Thanks, seems to work fine for `ls\nls`

PR #474, opened 2018-04-16
Add params for katello-host-tools

sean797:


ehelms (approved):
A worthy follow up conversation would be whether to alter this to make the agent false by default to allow users to opt-in. This sets the stage for REX being primary (aka the push model) and users to opt-in to the agent (aka pull model).

mmoll:
merged, thanks @sean797!

PR #476, opened 2018-04-18
Add minimal IPv6 support to preseed

arjenz:
Fill /etc/network/interfaces with IPv6 stanza if .ip6 is present

### DIFF
--- a/provisioning_templates/snippet/preseed_networking_setup.erb
+++ b/provisioning_templates/snippet/preseed_networking_setup.erb
@@ -7,10 +7,10 @@ description: this will configure your host networking, it configures your primar
     as other configures NICs. It supports physical, VLAN and Alias interfaces. It's intended to be
     called in your preseed finish template.
 %>
-<% subnet = @host.subnet -%>
-<% dhcp = subnet.dhcp_boot_mode? -%>
-<% subnet6 = @host.subnet6 -%>
-<% dhcp6 = subnet6.dhcp_boot_mode? -%>
+<% host_subnet = @host.subnet -%>
+<% host_dhcp = host_subnet.nil? ? false : host_subnet.dhcp_boot_mode? -%>
+<% host_subnet6 = @host.subnet6 -%>
+<% host_dhcp6 = host_subnet6.nil? ? false : host_subnet6.dhcp_boot_mode? -%>
 
 real=`ip -o link | awk '/<%= @host.mac -%>/ {print $2;}' | sed s/://`
 cat << EOF > /etc/network/interfaces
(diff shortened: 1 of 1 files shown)


provisioning_templates/snippet/preseed_networking_setup.erb
```diff
@@ -27,6 +27,11 @@ iface $real inet <%= dhcp ? 'dhcp' : 'static' %>
     dns-nameservers <%= @host.subnet.dns_primary %> <%= @host.subnet.dns_secondary %>
     dns-search <%= @host.domain %>
 <% end %>
+<% if @host.ip6 %>
```
mmoll: I think there's also a check for `@host.subnet6` needed here.

provisioning_templates/snippet/preseed_networking_setup.erb
```diff
@@ -47,5 +52,10 @@ iface $real inet <%= dhcp ? 'dhcp' : 'static' %>
     address <%= interface.ip %>
     netmask <%= subnet.mask %>
 <% end %>
+<% if interface.ip6 %>
```
mmoll: I think there's also a check for `interface.subnet6` needed here.

mmoll (dismissed):

mmoll:
I'm a bit unsure about this. I do have IPv6 hosts that got their values filled by facts, but don't have a subnet assigned. Even if they have, `gateway` is not a mandatory parameter in the subnet.

arjenz:
I added a couple of if statements

What would happen was that I'd assign IPv6 through foreman, that would be ignored, and then the node would submit a fact with an autoconfig IPv6 IP (if that happened to be available on the local network), overwriting my assignment.
Without IPv6 autoconfig it will just ignore the IPv6 IP during provisioning, I think I've both seen it clear the assignment in foreman (after a puppet run with no IPv6 on any interface), and foreman sticking to the assignment but nothing acting on it. Not 100% sure on this last behaviour because I was  troubleshooting at the same time.

mmoll:
We're getting there... :)
After another look, I'd like to have a check for the "static" boot-mode to get added and use the added sections only when its set. Then I think all is ready.

arjenz:
And I think the `next if` each loop for the managed_interfaces will skip IPv6 if there is no IPv4 subnet present (because it checks for `interface.subnet.nil?`) currently.
I am not using this part managed_interfaces myself now, added it for completeness sake in this file.

I'll try and fix this and look at the static boot-mode logic tomorrow, thank you for reviewing so far.

arjenz:
@mmoll I reworked the managed_interfaces to handle "IPv6 only" interfaces and added checks for subnet6 existence/boot_modes.

mmoll:
@arjenz Sorry, that slipped off my radar...

I did still have some problems in my environment with the latest changes and started to rework the whole snippet, resulting into a quite big change, also pushed here.

Could you please review and check this in your environment?

PR #477, opened 2018-04-18
Set hostname (again) within preseed

arjenz:
netcfg will switch hostname of installer runtime environment to the
PTR for the DHCP IP if present. This causes the first puppet check-in
to be with the "wrong" hostname (which will confuse foreman). Switch
runtime hostname back to whatever foreman supplied before doing other
operations.


provisioning_templates/finish/preseed_default_finish.erb
```diff
@@ -26,6 +26,7 @@ oses:
 /bin/sed -i "s/.*dns-search.*/\tdns-search <%= @host.domain %>/g" /etc/network/interfaces
 /bin/sed -i "s/^<%= @host.ip %>.*/<%= @host.ip %>\t<%= @host.shortname %>.<%= @host.domain %>\t<%= @host.shortname %>/g" /etc/hosts
 /bin/echo <%= @host.shortname %> > /etc/hostname
+hostname  <%= @host.shortname %>.<%= @host.domain %>
```
mmoll: two small things:

- Please wrap the domain part into `<% if @host.domain -%>`
- use `/bin/hostname`

mmoll (changes requested):

provisioning_templates/finish/preseed_default_finish.erb
mmoll: actually, we also don't check for having a domain a bit above, so that part can stay as-is for now...

mmoll (changes requested):

arjenz:
Changed it to /bin/hostname

mmoll (approved):

mmoll:
merged, thanks @arjenz!

PR #478, opened 2018-04-19
Adding snippet to blacklist kernel modules

sbernhard:
With this snippet you can blacklist kernel modules like floppy or pcspkr by setting a global parameter:
key: blacklist_kernel_modules value: "floppy pcspkr"
This solves warning like "blk_update_request: I/O error, dev fd0, sector 0" during system startup.


provisioning_templates/provision/autoyast_sles_default.erb
```diff
@@ -186,6 +186,9 @@ cp /etc/resolv.conf /mnt/etc
 
 <%= snippet('remote_execution_ssh_keys') %>
 
+#blacklist kernel modules
+<%= snippet "blacklist_kernel_modules" %>
```
lzap: No SUSE expert here, but isn't autoyast a XML syntax? This comment looks like kickstart one.

lzap (changes requested):
Frankly, we tend not to add those comments before snippets, it renders to nothing which is fine behavior.

provisioning_templates/provision/autoyast_sles_default.erb
sbernhard: You are right. AutoYAST ist XML syntax but this change is in a script section which are just common shell commands.

sbernhard:
I will remove the comments.

lzap (approved):
Thanks

sbernhard:
How to get this PR "in" / done?

sbernhard:
[ping]

provisioning_templates/snippet/blacklist_kernel_modules.erb
```diff
@@ -0,0 +1,12 @@
+<%#
+kind: snippet
+name: blacklist_kernel_modules
+model: ProvisioningTemplate
+snippet: true
+%>
+
```
lzap: Can you remove the blank line? This will create blank line on every rendered template, we already have couple and we actually want *less* :-)

provisioning_templates/snippet/blacklist_kernel_modules.erb
```diff
@@ -0,0 +1,12 @@
...
+<% if host_param('blacklist_kernel_modules') %>
+  <% host_param('blacklist_kernel_modules').split.each do |mod| %>
```
lzap: Can you use `-%>` here and above to prevent empty lines? This will create blank line on every rendered template as well.

lzap (changes requested):
Sorry I am very late on some notifications, other than that we can merge.

lzap:
@sbernhard just ping me anytime for any PR on IRC if I leave you waiting, this is not intentional at all, sorry about that. Just busy. I am "lzap" on IRC.

lzap (approved):
Thanks.

PR #480, opened 2018-05-30
Refs #21717 - Added support for enable and disable

ik5:


job_templates/service_action_-_ansible_default.erb
```diff
@@ -12,7 +12,7 @@ template_inputs:
     At least one of state and enabled are required. Note that reloaded will start
     the service if it is not already started, even if your chosen init system wouldn't
     normally.
-  options: "started\r\nstopped\r\nrestarted\r\nreloaded"
+  options: "started\r\nstopped\r\nrestarted\r\nreloaded\nenable\ndisable"
```
adamruzicka: Setting `state` to `enable` would lead to passing `enable` into ansible's `service.state` which it wouldn't understand. These options have to directly map to choices for `state`, see [ansible docs](https://docs.ansible.com/ansible/latest/modules/service_module.html)

ik5: i see the `enabled` flag at the documentation.
How can I set it using the `options` to be true or false?
Or should I use something else.

adamruzicka: There's already an input for it here https://github.com/theforeman/community-templates/pull/480/files#diff-28abe89ce1acc27089ab7963b059131cR36

ik5: thank you

ik5:
forgot to push it yesterday, so here it is now :)

job_templates/service_action_-_ansible_default.erb
```diff
@@ -12,7 +12,7 @@ template_inputs:
     At least one of state and enabled are required. Note that reloaded will start
     the service if it is not already started, even if your chosen init system wouldn't
     normally.
-  options: "started\r\nstopped\r\nrestarted\r\nreloaded"
+  options: "started\r\nstopped\r\nrestarted\r\nreloaded
```
adamruzicka: Missing double quotes at the end of the line

job_templates/service_action_-_ssh_default.erb
```diff
@@ -19,5 +19,17 @@ template_inputs:
...
+  <% case input("action")
+  when 'enable'
+%>
+chkconfig --add <%= input("service") %>
+  <%
+    when 'disable'
+  %>
+chkconfig --del <%= input("service") %>
+  <%
+      else
```
adamruzicka: The indentation in this template seems off

adamruzicka (changes requested):
The ansible template needs to be fixed.

job_templates/service_action_-_ansible_default.erb
ik5: good catch, thank you

job_templates/service_action_-_ssh_default.erb
ik5: I hope that it is better now, i did "auto format" before, and now by hand :)

ik5:
i hope that my changes are better, thank you! :)

lzap:
`reload` is a thing too :-)

ik5:
@lzap but this is Sparta :-P 
(done)

ik5:
@adamruzicka wdyt?

adamruzicka (approved):
Looks good to me

iNecas:
Thanks @ik5 and @adamruzicka

PR #481, opened 2018-06-01
Allow snippet remote_execution_ssh_keys to handle strings

dgoetz:
This allows to use the snippet also without using the Remote execution plugin.

refs #16107


ekohl (approved):
I think this is correct but could be good if someone else takes a look as well.

mmoll:
merged, danke @dgoetz!

PR #482, opened 2018-06-08
Refs #23861 pxegrub2_chainload support for OpenSuSE and SLES

mzima:
When a newly provisioned SuSE node boots for the first time, after the initial installation, the chainload Grub2 EFI from ESP fails (because the grubx64.efi file can't be found).

The SuSE UEFI bootloader can be found at this location:

**SLES:**
`($root)/EFI/sles/grubx64.efi`

**OpenSuSE:**
`($root)/EFI/opensuse/grubx64.efi`

This pull request solves this issue.


mmoll:
@lzap @hboetes could you have a look? :point_up:

hboetes:
I changed jobs, so I have no sles and no foreman atm. Although foreman is quite sure to return.

mmoll:
@hboetes allright, thanks for letting us know :+1:

lzap (approved):

lzap (changes requested):
Can you change `provisioning_templates/snippet/pxegrub_chainload.erb` as well? Squash into single commit. Thanks.

mzima:
Recent OpenSuSE/SLES versions only support Grub2 for booting EFI disks. Older SuSE versions (e.g. SLES 11) are using elilo instead, so changing `provisioning_templates/snippet/pxegrub_chainload.erb` would not do much harm, but also not make much sense either ...

mzima:
I've tested grub1 and grub2 UEFI chainloading against current SuSE versions ... and it's working fine. Older SuSE versions (like SLES11) are not covered by this change as they are using the elilo bootloader for UEFI boot.

lzap (approved):
Thanks, if you figure out nicer way of chanbooting in UEFI, let us know :-)

PR #483, opened 2018-06-14
Improved menu titles for PXELinux

lzap:
* Indicate ESC to stop in case of emergency
* Added menu to OS installer PXE with default of 10 seconds delay (configurable)

Previously we booted immediately into OS installer, together with problems we have with wget corrupting images on TFTP, this gives some extra time if users are having issues. They can also create a hostgroup with longer time as a workaround.


mmoll:
looks sensible :+1:

mmoll:
merged, díky @lzap!

PR #461, opened 2018-03-05
Refs #21007 - new failed call for kickstarts

lzap:
Core PR needed: https://github.com/theforeman/foreman/pull/5292

PR #485, opened 2018-06-20
Update pykickstart to the latest version

ekohl:


test/preseed/provision_test.rb
```diff
@@ -1,34 +1,32 @@
...
+        ['9', '4', 'stretch'],
+    ],
+    'Ubuntu' => [
+        ['14', '04', 'trusty'],
+        ['16', '04', 'xenial'],
+        ['18', '04', 'trusty'],
+    ],
+  }.each do |distro, releases|
+    releases.each do |major, minor, release|
+      describe "#{distro} #{major} #{minor} #{release}" do
```
ekohl: Will change this to major.minor on the next update.

test/preseed/provision_test.rb
```diff
@@ -1,34 +1,32 @@
...
+  {
+    'Debian' => [
+        ['7', '11', 'wheezy'],
+        ['8', '6', 'jessie'],
+        ['9', '4', 'stretch'],
+    ],
+    'Ubuntu' => [
+        ['14', '04', 'trusty'],
+        ['16', '04', 'xenial'],
+        ['18', '04', 'trusty'],
```
ekohl: Will change this to bionic on the next update

ekohl (commented):
So the scope increased a bit to easily support multiple versions and I added some kickstart templates since we now test on Fedora 29 as well which threw some deprecation warnings. Debian Stretch and Ubuntu Bionic were added as well.

The benefit of the DSL-style syntax is that we have the proper output matching

@lzap mind having a look?
@mmoll any idea on how to handle the debconf passwords?

ekohl:
I've handled the debconf passwords but now it complains about the keyboard option. This makes me wonder if we ever really dealt with debian provision tests

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -40,7 +40,9 @@ This template accepts the following parameters:
   chef_enabled = @host.respond_to?(:chef_proxy) && @host.chef_proxy
   section_end = (rhel_compatible && os_major <= 5) ? '' : '%end'
 %>
+<% if @host.operatingsystem.name != 'Fedora' || os_major < 29 -%>
 install
+<% end -%>
```
lzap, line 45: Not sure if the statement is correct, what OS/version combination must not have this statement? To me this smells better: `<% if @host.operatingsystem.name == 'Fedora' && os_major >= 28 -%>`

test/kickstart/provision_test.rb
```diff
@@ -1,66 +1,37 @@
...
+    options[:majors].each do |major|
+      describe "#{distro} #{major}" do
+        it 'will compile' do
+          code, stdout, stderr = @renderer.render(distro, major, options[:minor], options[:prefix])
+          stdout.must_equal ''
+          stderr.must_equal ''
+          code.must_equal 0
+        end
+      end
+    end
```
lzap, line 35: Facepalm, doing the same in PR: https://github.com/theforeman/community-templates/pull/461/files

Removing from there then. This is cleaner.

test/test_helper.rb
```diff
@@ -55,9 +56,11 @@ class FakeNamespace
   def initialize(family, name, major, minor, release = nil)
     @mediapath = 'url --url http://localhost/repo/xyz'
     @root_pass = '$1$redhat$9yxjZID8FYVlQzHGhasqW/'
-    @grub_pass = 'blah'
+    @grub_pass = '--password=blah'
```
lzap, line 59: Can you actually make this a valid password so it works even with `--crypted` option? Here is "redhat" one (surprisingly long).

'--password grub.pbkdf2.sha512.10000.A3602AF184F11072D8BB3801806D9B4F3B564F91478CB77D6F7E84E4D323CAE1DE11C8361311B890E4448892C4C64393DC65BFF5F26606A5D173B549931F2F20.A52D4290C63B6DD3EC330617AD71AFFB03C126EC075A6F5FED562898C64B7B4BD70485CD9109A1CD731D8F9F26FD601CA906BC7E4B806FCC2FF2A19DAE2E4123'

.gitignore
```diff
@@ -1,4 +1,5 @@
 .idea
 .DS_Store
+Gemfile.lock
```
lzap, line 3: Facepalm, was doing the same in PR: https://github.com/theforeman/community-templates/pull/486/files

Removed.

lzap (changes requested):
Nice.

lzap:
There are some failures atm.

test/test_helper.rb
ekohl: I did notice that foreman never actually passes in just `--password`, it's always `--iscrypted --password` but EL5 doesn't like that. I guess there we pass in `--md5pass`. This was a bit of a shortcut.

provisioning_templates/provision/kickstart_default.erb
ekohl: Starting Fedora 29 it shouldn't be there, but until then I think it should. The alternative is `<% unless @host.operatingsystem.name == 'Fedora' && os_major >= 29 -%>` which may be a bit easier to read.

test/test_helper.rb
lzap: In my other branch I had to actually do this, I guess you made some change that made it working. I am fine with blah then.

ekohl:
Tests now pass :)

lzap (approved):

lzap:
Thank you very much for your contribution, this is now merged.

PR #486, opened 2018-06-21
Fixes #24026 - Make grub2 working with RHEL6

lzap:
When you try to boot RHEL6 Anaconda with PXEGrub2 UEFI it fails with "Kernel too old" error. It turns out that it's still possible to boot older kernel in UEFI mode but "linuxefi" command cannot be used - "linux" works instead.


mmoll:
merged, díky @lzap!

PR #487, opened 2018-06-27
Fixes #24084 - root pass algo is downcase now

lzap:

PR #488, opened 2018-06-27
Fixes #24084 - root password hash propagation to authconfig

kallies:
If SHA512 is used in host.operatingsystem.password_hash (SHA256 is the default in kickstart_default.erb), authconfig will use SHA256 anyway because it has no value SHA512 (but sha512).

As described in https://github.com/theforeman/foreman/pull/5746


kallies:
Duplicate of #487

lzap:
Sorry @kallies was faster :-) Thanks for the report and the fix tho!

PR #489, opened 2018-06-28
Add more puppet job templates

dgoetz:
This pull request adds more puppet job templates.

* Disabling of the agent including a comment which should help the local admin to see why it was disabled. I also created a feature request to have a macro for the user executing the job to add it to the comment. (https://projects.theforeman.org/issues/23955)
* Enabling of the agent
* Installation of puppet modules to provide something similar to the ansible roles. For the Boolean options I created another feature request, for now I think forcing the user to type true is fine. (https://projects.theforeman.org/issues/23956)

I am not sure if the naming of the files is fine but I tried to align it to the other templates.

### DIFF
--- a/job_templates/puppet_agent_disable_-_ssh_default.erb
+++ b/job_templates/puppet_agent_disable_-_ssh_default.erb
@@ -1,0 +1,16 @@
+<%#
+name: Puppet Agent Disable - SSH Default
+model: JobTemplate
+job_category: Puppet
+description_format: Disable Puppet agent
+snippet: false
+template_inputs:
+- name: comment
+  required: false
+  input_type: user
+  description: Reason for disabling the Puppet agent
+  advanced: false
+provider_type: SSH
+kind: job_template
+-%>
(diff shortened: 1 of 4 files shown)


dgoetz:
Added the current_user macro available from 1.20 to the disable job.

ekohl (approved):
It looks correct, but I'm not that familiar with job templates.

job_templates/puppet_agent_disable_-_ssh_default.erb
```diff
@@ -0,0 +1,17 @@
...
+snippet: false
+template_inputs:
+- name: comment
+  required: false
+  input_type: user
+  description: Reason for disabling the Puppet agent
+  advanced: false
+provider_type: SSH
+kind: job_template
+%>
```
ares: Ideally add - before % to avoid extra line, I suppose you exported this from Foreman, so we may need to update oir exporter

ares (changes requested):
Looks all great, there are extra empty lines, which would be nice to be removed. I haven't tested yet, but from reading it looks great!

dgoetz:
I added the `-` to `%>` and removed the empty lines. 
I am not sure if I wrote them manually and then imported them the first time or did it the other way.

ares (approved):

ares:
Thanks @dgoetz and sorry for the delay. This works great, merging!

PR #491, opened 2018-07-02
Refs #19389 - Use medium_provider framework

ShimShtein:
This commit modifies templates to use `medium_provider` framework introduced in https://github.com/theforeman/foreman/pull/5244.

### DIFF
--- a/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
+++ b/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
@@ -50,7 +50,7 @@ oses:
 
   # optional repository for Atomic
   if @host.operatingsystem.name.match(/Atomic/i)
-    options.push("inst.repo=#{@host.operatingsystem.medium_uri(@host)}")
+    options.push("inst.repo=#{medium_uri}")
   end
 
   if host_param('kickstart_liveimg')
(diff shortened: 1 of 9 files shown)


lzap:
I vote creating helper method for this, so users don't need to figure out what media provider is. Just a single method with clear contract can be documented easily.

mmoll:
@ShimShtein this needs a rebase :)

ShimShtein:
@mmoll, rebased.

ekohl:
The failures appear to be unrelated

mmoll:
merged, thanks @ShimShtein!

lzap:
So we missed 1.19 with this, I tested PXE provisioning with RC1 and it works. Is it safe to use the old code in the templates @ShimShtein ?

ShimShtein:
@lzap  Methods that will work with older code and show deprecation warning:

* medium_uri

Methods that will throw an exception:

* mediumpath
* boot_files_uri

I can make `#boot_files_uri` to work with older templates, but `#mediumpath` is overloaded in each OS. I can make it work with older templates, but if someone adds other OS - I would not be able to help.

If you think it's necessary, I'll add deprecation code.

ShimShtein:
@lzap sorry, I was looking at a wrong commit. In older code there are a lot of methods that don't have enough information to construct a medium provider ([pxe_prefix for example](https://github.com/theforeman/foreman/commit/2455b570ad0d5db8c3306b803460a6d64808a266#diff-ca0b1ebc260d7633163da1e5450eabb9L194)). 

It would be impossible to use older templates with the new code. I have tried to add meaningful exceptions everywhere I could think of to minimize the impact on the users.

lzap:
Thanks for comments, so I assume it should work in 1.19 (I tested this but doublechecking).

PR #492, opened 2018-07-05
Fixes #24151 - bond device fails with vlanid for nil:NilClass

vgunasek:


vgunasek:
Can someone look into this request?

mmoll:
merged, thanks @vgunasek!

PR #493, opened 2018-07-17
Ansible shutdown with 1m delay so callback can return

dLobatog:
Delaying the shutdown 1 minute so the Ansible job can return and the job
invocation can be green. Otherwise 'shutdown' may work but the job
invocation will fail.


ares:
Tested, works as expected, thanks @dLobatog, merging!

PR #495, opened 2018-07-24
Fixes #24360 - add support for provisioning on vlan

ares:


ekohl:
Could you have a look at the test failure? I think we need to mock `vlanid` in our tests.

lzap:
Any chance of updating Debian too?

I think this line needs to be broken apart soon into array so we can add elements into array in an arbitrary Ruby block - would be much cleaner. We do this in PXELinux templates.

timogoebel:
This requires the host to have a trunked network port. I don't think it's a good idea to rely solely on the interfaces vlanid parameter here.

We define the vlanid on the subnet to have a chance to match it with the corresponding vmware network. But the hosts aren't trunked.

lzap:
Can you elaborate your concern Timo? We have multiple customers asking for this, they specifically want to use trunk VLAN on provisioning interface, including PXE and discovery. For this reason I am working on https://github.com/theforeman/foreman/pull/5989

timogoebel:
@lzap: I'd love the feature being added. To hopefully clarify my concern: Not every host that is installed in a subnet that has the vlan id parameter set actually had a trunked interface. I fear we need a parameter "trunked: yes/no" on the nic model.

lzap:
Agreed, this patch is reading vlanid from host tho.

@ares in order to support this in PXE environment, we also need to add this to PXE templates: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/chap-anaconda-boot-options

timogoebel:
> Agreed, this patch is reading vlanid from host tho.

True, but that is basically delegated to the subnet. ;-)

lzap:
> True, but that is basically delegated to the subnet. ;-)

Whoops.

mmoll:
@ares what's the status here?

lzap:
I've implemented this already, closing.

lzap:
@timogoebel about your concern - I don't understand why would you set VLANID on a NIC for a network that is what I call "native VLAN" (VLAN done by switch). If that's only for informative (inventory) purposes, then it could be a use case, then simply do not fill "attached_to" and it will not be used in the template.

ares:
@lzap it would be good to know, which commit implemented it. Please update https://projects.theforeman.org/issues/24360 with the link to the PR so it (and BZ) gets updated.

PR #496, opened 2018-07-25
When booting UEFI, you will receive Kernel Panic

jasondickerson:
Kernel panic - not syncing: VFS:  Unable to mount root fs on unknown-block(0,0) 

Unless you include initrd=initrd.img on the kernel line.  I have tested and this is safe for BIOS and UEFI booting.  Apparently, UEFI needs the location of the initrd.img to be specified or it cannot find it; unlike bios that seems to just look where the kernel is.  This works with bios RHEL 6 and 7, and with UEFI RHEL 7.  UEFI RHEL 6 gets an iPXE exec format error.  I have only tested with RHEL 6 and 7.

### DIFF
--- a/provisioning_templates/iPXE/autoyast_default_ipxe.erb
+++ b/provisioning_templates/iPXE/autoyast_default_ipxe.erb
@@ -12,6 +12,6 @@ oses:
 <% kernel = boot_files_uris[0] -%>
 <% initrd = boot_files_uris[1] -%>
 
-kernel <%= kernel %> splash=silent install=<%=@host.os.medium_uri(@host)%> autoyast=<%= foreman_url('provision') %> text-mode=1 useDHCP=1
+kernel <%= kernel %> initrd=initrd.img splash=silent install=<%=@host.os.medium_uri(@host)%> autoyast=<%= foreman_url('provision') %> text-mode=1 useDHCP=1
 initrd <%= initrd %>
 boot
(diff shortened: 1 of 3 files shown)


lzap (changes requested):
Hey, thank you! I think it needs to be:

    initrd=<%= initrd %>

jasondickerson:
Hey lzap thanks for that comment.  Let me explain my observations/theories on that.  The iPXE template has 3 significant lines.  kernel, initrd, and boot.  The kernel line contains the location of the vmlinuz file to download along with the kernel variables.  The initrd line contains the network location of the initrd file to download, and the final line "boot", initiates the boot sequence of the vmlinuz kernel with the kernel parameters on the kernel line.  The initrd variable on the kernel line must point to the downloaded location of the initrd file.  The reason I use "initrd=initrd.img" is that the file name is consistent between RHEL, CentOS, and Fedora.  If there is an OS that has a different name for the initrd file, then this change would not work.  Instead we would need to use the foreman variable for initrd and parse out the just the file name.  However, I suspect the name of the file is consistent among linux kernels, so I do not see a need for that at this time.  

To come to this conclusion/theory, I viewed the documentation on ipxe.org, and pulled apart a couple of other UEFI compatible boot iso's.  They all had initrd=initrd.img hard coded.  I know this is not 100% conclusive, but I have not found anything to lead me to believe this will not work, as yet.  Also I have tested this with RHEL 6 BIOS and RHEL 7 BIOS and UEFI.  However there seems to be an incompatibility between the RHEL 6 kernel and iPXE on UEFI.  The net effect of this change in my observations, is RHEL 6 and 7 BIOS still works and RHEL 7 UEFI works.  So it is  a net gain, without losing functionality.  

If there is anyone intimately familiar with the kernel parameter initrd, they could further verify my hypothesis.  

I do have this working in a very large enterprise environment.

lzap (approved):
Thanks for explanation, I have missed that.

What puzzles me is why this was working before.

PR #490, opened 2018-06-29
Removed blacklist block for pxe_kernel_options

lzap:
This is now provided by Foreman core directly via pxe_kernel_options

PR #497, opened 2018-08-07
Fixes #24557 - Allow using live images from Katello

adamruzicka:
Before this change the kickstart_liveimg had to be an absolute
URL. With this change it can be either absolute or relative.

If Katello is in play, the relative URL can be in format of
"custom/$PRODUCT/$REPOSITORY/$IMAGE_NAME".


iNecas (approved):
LGTM

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -44,7 +44,19 @@ This template accepts the following parameters:
 install
 <% end -%>
 <% if host_param('kickstart_liveimg') %>
-liveimg --url=<%= host_param('kickstart_liveimg') %> <%= proxy_string %>
+<%
+img_name = host_param('kickstart_liveimg')
```
lzap: This variable could be inside the if - else block.

Is this the final version? I see two commits after @iNecas 's LGTM.

lzap: Lemme try again, I was wondering if this is better:

```ruby
liveimg_url = if host_param('kt_activation_keys')
  repository_url(img_name, 'isos')
else
  img_name = host_param('kickstart_liveimg')
  if ! img_name.nil? && img_name.match(%r|^([\w\-\+]+)://|)
    img_name
  else
    "#{@host.operatingsystem.medium_uri(@host)}/#{img_name}"
  end
end
```

adamruzicka: `img_name` is used on L2 in your snippet as well. Apart from this, I don't plan on making any more changes to this PR

lzap (approved):
Okay.

provisioning_templates/provision/kickstart_default.erb
lzap: Wait, where exactly is used? I don't see it being used, you introduced this variable in this PR.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -44,7 +44,19 @@ This template accepts the following parameters:
...
+liveimg_url = if host_param('kt_activation_keys')
+  repository_url(img_name, 'isos')
+else
+  if img_name.match(%r|^([\w\-\+]+)://|)
```
lzap: This can easily raise `nil` exception.

adamruzicka: It can't, it is inside the true branch of `<% if host_param('kickstart_liveimg') %>` so it is guarranteed to be non-nil

lzap: Oh fine missed that, thanks!

lzap:
I don't have ovirt setup so not merging @iNecas thanks!

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -44,7 +44,19 @@ This template accepts the following parameters:
 install
 <% end -%>
 <% if host_param('kickstart_liveimg') %>
-liveimg --url=<%= host_param('kickstart_liveimg') %> <%= proxy_string %>
+<%
```
iNecas: One nitpick: we currently have:

```ruby
<% if host_param('kickstart_liveimg') %>
<%
img_name = host_param('kickstart_liveimg')
liveimg_url = if host_param('kt_activation_keys')
```

Probably it should be just:

```ruby
<%
if host_param('kickstart_liveimg')
  img_name = host_param('kickstart_liveimg')
  liveimg_url = if host_param('kt_activation_keys')
  # ...
```

iNecas:
One last comment

provisioning_templates/provision/kickstart_default.erb
adamruzicka: Fixed

iNecas:
Thanks @adamruzicka

lzap:
Sorry I have a comment which I realized after this was merged. In Foreman we now have a way of adding things to kernel command line: `pxe_kernel_options` method. This could be part of Foreman core and RedHat operating system class! No need to make our templates that complex, stage2 parameter can be present for all Anaconda installations.

Fixes #24109 - common kernelcmd param for all OSes

There's pending PR which conflicted recently in this repo:

https://github.com/theforeman/community-templates/pull/490

adamruzicka:
If we moved this to core, it would lead to katello-specific code (`repositories_url`) living in core and I'm not sure I like it. Splitting it between core and katello also doesn't seem like a good idea.

iNecas:
I like the current approach in this case better, as otherwise, the user would need to care both about the kickstart_liveimg and the stage2 param, both set in different way. I think it's good to be opinionated here a bit and set things that we know we require.

PR #498, opened 2018-08-07
Fixes #24425 - Templates for modularity actions

parthaa:
This commit contains rex templates for modularity related actions.


job_templates/module_action_-_ansible_default.erb
```diff
@@ -0,0 +1,50 @@
...
+- name: pre_script
+  description: A script to run prior to the module action
+  input_type: user
+  required: false
+  advanced: true
+- name: action
+  description: 'The module action: enable, disable, install, update, or remove, lock, unlock'
+  input_type: user
+  required: true
+  options: "\nlist\ninfo\nenable\ndisable\ninstall\nupdate\nremove\nprovides\nlock\nunlock\nprofile\nstreams"
```
stbenjam: The list of options doesn't match the description above

job_templates/module_action_-_ssh_default.erb
```diff
@@ -0,0 +1,67 @@
...
+- name: pre_script
+  description: A script to run prior to the module action
+  input_type: user
+  required: false
+  advanced: true
+- name: action
+  description: 'The module action: enable, disable, install, update, or remove, lock, unlock'
+  input_type: user
+  required: true
+  options: "\nlist\ninfo\nenable\ndisable\ninstall\nupdate\nremove\nprovides\nlock\nunlock\nprofile\nstreams"
```
stbenjam: The list of options doesn't match the description above

stbenjam (commented):
I can't actually test yet, since I can't get any modules visible on my F28 box - but looks good, just a couple of comments

job_templates/module_action_-_ansible_default.erb
stbenjam: I would probably just leave description as "The module action, e.g. install, enable, etc" or something, instead of providing a long list.

stbenjam:
I managed to test on the f28 box in Forklift, and it works well

job_templates/module_action_-_ssh_default.erb
```diff
@@ -0,0 +1,67 @@
+<%#
+kind: job_template
+name: Module Action - SSH Default
+model: JobTemplate
+job_category: Modules
+description_format: "%{action} %{module_spec}"
```
stbenjam: Can this be prefixed by "Module" like the Ansible one is?

job_templates/module_action_-_ansible_default.erb
```diff
@@ -0,0 +1,50 @@
+<%#
+kind: job_template
+name: Module Action - Ansible Default
+model: JobTemplate
+job_category: Modules
```
stbenjam: Unfortunately the category name has to be "Ansible Modules" to keep conventions with everything else.  Every category only has one job template :-\  There's a BZ open to change that.  Notice the two Template selectors:

screenshot from 2018-08-10 13-00-24

stbenjam:
After changing the ansible category (see above), things work well, however because we are running dnf directly instead of using the dnf module (which doesn't support modularity), Ansible warns:

"Consider using the dnf module rather than running dnf. If you need to use command because dnf is insufficient you can add warn=False to this command task or set command_warnings=False in ansible.cfg to get rid of this message." 


You can add this below register on the shell command:

```yaml
args:
  warn: false
```

stbenjam:
Incidentally, once https://github.com/ansible/ansible/pull/38457/files is merged and in an ansible release, we should move to the native provider.

stbenjam (approved):
LGTM!

PR #500, opened 2018-08-21
Add support for RHEL8

lzap:
Basically an extension of 69923c46b1ca6 - ksvalidator is failing for RHEL8 due to changes made in modern Fedora (29).

Fixes https://github.com/theforeman/community-templates/issues/499


provisioning_templates/provision/kickstart_default.erb
```diff
@@ -29,6 +29,8 @@ This template accepts the following parameters:
 %>
 <%
   rhel_compatible = @host.operatingsystem.family == 'Redhat' && @host.operatingsystem.name != 'Fedora'
+  is_fedora = @host.operatingsystem.family == 'Fedora'
```
ekohl: I think you meant `.name` here instead of `.family`. If you introduce this variable, you can also use it in other places. Like above in `rhel_compatible` or `realm_compatible`.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -80,10 +82,10 @@ firewall --disable
 <% else -%>
 firewall --<%= os_major >= 6 ? 'service=' : '' %>ssh
 <% end -%>
-<% if @host.operatingsystem.name == 'Fedora' && os_major >= 28 -%>
-authselect --useshadow --passalgo=<%= @host.operatingsystem.password_hash.downcase || 'sha256' %> --kickstart
-<% else -%>
+<% if (is_fedora && os_major < 29) || (is_rhel && os_major < 8) -%>
```
ekohl: I liked the previous style better. The true branch should be the current, false should be the legacy way. It's also a smaller diff so easier to review.

You're also introducing a slight change here with `>= 28` vs `< 29`. It means you're changing F28 to use the legacy way while when I tested this with ksvalidator it wanted to use the modern way.

ekohl (changes requested):

lzap:
Good points, rebased.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -40,7 +41,7 @@ This template accepts the following parameters:
   chef_enabled = @host.respond_to?(:chef_proxy) && @host.chef_proxy
   section_end = (rhel_compatible && os_major <= 5) ? '' : '%end'
 %>
-<% if @host.operatingsystem.name != 'Fedora' || os_major < 29 -%>
+<% if (is_fedora && os_major >= 28) || os_major >= 8 -%>
```
ekohl: You reversed the logic here. In other places we also add a check `rhel_compatible` when comparing os majors so that you don't break older Fedora versions.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -80,7 +81,7 @@ firewall --disable
 <% else -%>
 firewall --<%= os_major >= 6 ? 'service=' : '' %>ssh
 <% end -%>
-<% if @host.operatingsystem.name == 'Fedora' && os_major >= 28 -%>
+<% if (is_fedora && os_major >= 28) || os_major >= 8 -%>
```
ekohl: Please use `rhel_compatible` here as well

ekohl (changes requested):

provisioning_templates/provision/kickstart_default.erb
lzap: What did I reverse here? It looks good to me assuming that I will add the missing `rhel_compatible`.

lzap: Oh I get it it's the `install` command which was *removed*.

lzap:
Now it should be correct hopefully.

ekohl (approved):

provisioning_templates/provision/kickstart_default.erb
ekohl: Correct. I'll admit my original code wasn't the clearest.

mmoll:
merged, díky @lzap!

lzap:
Thanks for the patience with me @ekohl this one was tricky to follow my crazy changes :-)

lzap:
THanks @mmoll Danke!

PR #501, opened 2018-08-22
Add custom-snippet parameter

alexjfisher:
Instead of re-syncing my custom kickstart template with the upstream default every time I upgrade Foreman, I thought it might be better to put all my customisations in a 'custom-snippet' and make this a parameter.

### DIFF
--- a/provisioning_templates/finish/preseed_default_finish.erb
+++ b/provisioning_templates/finish/preseed_default_finish.erb
@@ -29,6 +29,7 @@ oses:
 /bin/hostname  <%= @host.shortname %>.<%= @host.domain %>
 <% end -%>
 
+<%= snippet_if_exists(template_name + " custom snippet") %>
 <% if host_enc['parameters']['realm'] && @host.realm && @host.realm.realm_type == 'FreeIPA' -%>
 <%= snippet 'freeipa_register' %>
 <% end -%>
--- a/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
+++ b/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
@@ -67,3 +67,5 @@ title <%= template_name %>
   root (nd)
   kernel (nd)/../<%= @kernel %> ks=<%= foreman_url('provision') %> <%= pxe_kernel_options %> <%= ksoptions %>
   initrd (nd)/../<%= @initrd %>
+
+<%= snippet_if_exists(template_name + " custom menu") %>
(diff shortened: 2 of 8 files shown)


lzap:
Good idea, but let's do this properly for all most important provisioning templates?

* PXELinux kickstart and preseed
* Grub1 kickstart and preseed
* Grub2 kickstart and preseed
* Kickstart
* Preseed and finish script

I suggest using macros `snippet_if_exists` and `template_name` which are present for several Foreman releases now. You can construct this and put it into several different places in kickstart and preseed (pre, post, packages etc):

    %packages
    <%= snippet_if_exists(template_name + " custom packages") -%>

    %post
    <%= snippet_if_exists(template_name + " custom post") -%>

This way you can create customized snippets:

* Kickstart default custom post
* Kickstart default PXEGrub2 custom menu

and similar.

alexjfisher:
@lzap Good suggestion (and sorry for the delay responding).  I'll try to find some cycles to spend on this soon.

alexjfisher:
@lzap I've tried to implement your suggestion (better late than never!!).  I've not used preseed though, so have only tested this with kickstart.

provisioning_templates/finish/preseed_default_finish.erb
```diff
@@ -29,6 +29,7 @@ oses:
 /bin/hostname  <%= @host.shortname %>.<%= @host.domain %>
 <% end -%>
 
+<%= snippet_if_exists(template_name + " custom snippet") %>
```
alexjfisher, line 32: I wasn't sure if I should put anything in `provision/preseed_default.erb` or where it would best go. (I'm only building EL hosts and am not familiar with preseed)

lzap (approved):
Nice idea and execution, well done. Mind creating a tutorial post in our community forums - the new Tutorials category?

lzap:
Oh tests are failing:

    NameError: undefined local variable or method `template_name' for #<FakeNamespace:0x0000000001f1eab8>

This is because we stub helpers/macros in the test, can you add a fake method there?

provisioning_templates/finish/preseed_default_finish.erb
lzap: This is fine, I'd probably name this " custom script".

lzap:
Here, you will need to add both `template_name` and `snippet_if_exists` I guess.

https://github.com/theforeman/community-templates/blob/develop/test/test_helper.rb#L139

Once we extract renderers into separate gem this should be gone.

alexjfisher:
@lzap Functions mocked.  I could probably write something on the forum.  Would this make it into 1.22?

lzap:
Thanks, yep we are on the good track for 1.22.

PR #502, opened 2018-08-22
Remove bashism in function definition

mdellweg:
This is needed, to function on hosts with `/bin/dash` as `/bin/sh`.
The keyword `function` is only known to bash.


lzap:
Thanks!

PR #503, opened 2018-08-28
fix realm-join when using provisioning w/cloud-init

acortina-git:
fix  joining realm when provisioning rhel6 with cloud-init

### DIFF
--- a/provisioning_templates/snippet/fix_hosts.erb
+++ b/provisioning_templates/snippet/fix_hosts.erb
@@ -5,7 +5,7 @@ model: ProvisioningTemplate
 snippet: true
 %>
 echo "<%= @host.shortname %>" > /etc/hostname
-hostname <%= @host.shortname %>
+hostname <%= @host %>
 cat > /etc/hosts << EOF
 <%# simple snippet to generate /etc/hosts when provisioning image based systems -%>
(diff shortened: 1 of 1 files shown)


acortina-git:
tested this change on rhel6 and rhel7, only affects rhel6

mmoll:
related: GH-361

ekohl:
So the tricky thing here is that on RH systems it's convention to store the FQDN in the hostname while Debian prefers the shortname. I also thought that cloud-init sets the hostname and wonder in which cases these fixes are needed. Then again, I'm not that familiar with cloud-init.

acortina-git:
Hi @ekohl 

mm then this could work:

```ruby
<% if @host.operatingsystem.family == 'Redhat' && @host.operatingsystem.major.to_i = 6 %>
hostname <%= @host %> 
<% else %>
hostname <%= @host.shortname %> 
<% end %>
```

ekohl:
That could work yes though it should be `== 6`. Would you mind applying that? Please also rebase on current develop since that includes a fix for the tests.

mmoll:
@alejojo what's the status here?

mmoll:
rebased version: GH-633

PR #505, opened 2018-09-04
Fixes #24753 - Fix bonds and vlans deployment

xprazak2:
This makes sure that vlans come up after
host has been provisioned. If a vlan
is on top of a bond, we have to
make sure it is not conrolled
by NM.

### DIFF
--- a/provisioning_templates/snippet/kickstart_networking_setup.erb
+++ b/provisioning_templates/snippet/kickstart_networking_setup.erb
@@ -34,7 +34,6 @@ kind: snippet
 <%- end -%>
 
 <%- @host.managed_interfaces.each do |interface| %>
-<%-   next if !interface.managed? || (interface.subnet.nil? && interface.subnet6.nil?) -%>
 <%-   next if bonded_interfaces.include?(interface.identifier) -%>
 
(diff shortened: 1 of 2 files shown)


ares:
Why we can't use NM?

xprazak2:
Because when the vlan is on top of the bond and has `NM_CONTROLLED=yes`, it just will not come up. I think it might be because bond is `NM_CONTROLLED=no`.

provisioning_templates/snippet/kickstart_networking_setup.erb
```diff
@@ -34,7 +34,6 @@ kind: snippet
 <%- end -%>
 
 <%- @host.managed_interfaces.each do |interface| %>
-<%-   next if !interface.managed? || (interface.subnet.nil? && interface.subnet6.nil?) -%>
```
ares, line 37: I think this is some leftover from testing, we shouldn't try to configure unmanaged interfaces or interfaces that are not attached to a subnet. All the rest makes sense.

EDIT: oh, perhaps you removed the whole line since vlan on top of bond does not need to have subnet

ares:
I tested the patch and it fixes the issue, the only thing I'd like I think is wrong is the skipping on managed check. I opened separate PR so this can be merged without waiting on @xprazak2 getting back. Let's continue the review there.

PR #506, opened 2018-09-04
Fix Oracle Linux 7.5 provisioning

alexjfisher:
`Mysql` was removed from `Addons` starting with the 7.5 install media.
Furthermore, the workaround that was used before (that removed
the `Server-Mysql` repo by redefining it without a path) no longer
works.


provisioning_templates/provision/kickstart_default.erb
```diff
@@ -31,6 +31,7 @@ This template accepts the following parameters:
   rhel_compatible = @host.operatingsystem.family == 'Redhat' && @host.operatingsystem.name != 'Fedora'
   is_fedora = @host.operatingsystem.name == 'Fedora'
   os_major = @host.operatingsystem.major.to_i
+  os_minor = @host.operatingsystem.minor.to_i
```
alexjfisher, line 34: `minor` isn't mandatory, so could be `nil`? But `nil.to_i == 0`, so that should be fine.

mmoll:
merged, thanks @alexjfisher!

PR #509, opened 2018-09-10
Refs #21204 - use local as the default for grubs

lzap:
See https://github.com/theforeman/foreman/pull/6052 and the ticket for info. Only merge if core is merged please.


ares:
reviewed and merged already to core, merging, thanks @lzap

lzap:
Thanks, snapshot feature will hopefully help us with tracking template changes.

PR #515, opened 2018-09-19
Fixes #24753 - Fix bonds and vlans deployment

ares:
follow up of #505


iNecas:
The changes are reasonable, thanks @ares

PR #520, opened 2018-10-03
Fixes #25117 - support both FreeIPA and Red Hat IDM in kickstart_default

chr15p:
Fixes #25117 and BZ1635680
The kickstart_default template assumes the value of @host.realm.realm_type will either be FreeIPA or Active Directory, but Satellite only allows the creates of AD or "Red Hat Identity Management" type realms.
IDM is essentially the "enterprisey" version of FreeIPA so can be handled in the same way. Therefore this patch adds an extra clause to the if statement so IDM or FreeIPA are both handled identically


provisioning_templates/provision/kickstart_default.erb
```diff
@@ -186,7 +186,7 @@ echo 'proxy = <%= proxy_uri %>' >> /etc/yum.conf
 
 <%= snippet 'redhat_register' %>
 
-<% if host_enc['parameters']['realm'] && @host.realm && @host.realm.realm_type == 'FreeIPA' -%>
+<% if host_enc['parameters']['realm'] && @host.realm  && ( @host.realm.realm_type == 'FreeIPA' ||  @host.realm.realm_type == 'Red Hat Identity Management') -%>
```
ekohl: Nit: could you remove the extra space before `&&` and there's also one before the `@host` of the last `realm_type` comparison.

ekohl (approved):

ekohl:
Thanks!

PR #521, opened 2018-10-05
Refs #25123 - add missing os metadata to provisioning templates

tstrachota:
Adds metadata about associated OSs that used to be hardcoded in the seed script.

https://github.com/theforeman/foreman/pull/6104 changes the seed script to use template imports and after it's merged we'll rely only on metadata.


ares:
Thanks @tstrachota, merging!

PR #522, opened 2018-10-05
Add missing OSes to ptables

ares:


ares:
the same as in #521 but for partition tables, @tstrachota mind to take a look?

tstrachota:
Thanks @ares !

PR #523, opened 2018-10-05
Add report templates

ares:
the sync script in core already works with this structure


ares:
note that kind should only be required for provisioning templates

tstrachota:
:+1: works fine, thanks @ares !

PR #526, opened 2018-10-10
Fixes #25163 - Add Subscriptions report

mbacovsky:
PR Katello/katello#7747 is required for this report.
I'd like to ask reviewer to check if the queries could be further optimized and if the data collected match what was requested.

### DIFF
--- a/report_templates/subscriptions.erb
+++ b/report_templates/subscriptions.erb
@@ -8,17 +8,11 @@ template_inputs:
   description: Search query to filter subscription entitlements. E.g. `quantity >
     0` to exclude subscriptions with unlimited entitlements. Keep unset for all subscriptions.
   advanced: false
-- name: Output format
-  required: true
-  input_type: user
-  description: Format of the generated report (CSV or YAML)
-  options: "csv\r\nyaml"
-  advanced: false
 model: ReportTemplate
 require:
 - plugin: katello
 ￼ version: 3.9.0
-%>
+-%>
 <%- load_pools(search: input('Subscriptions filter'), includes: [:subscription, :products]).each_record do |pool| -%>
 <%-   report_row(
         'ID': pool.id,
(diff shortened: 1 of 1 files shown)


report_templates/subscriptions.erb
```diff
@@ -0,0 +1,32 @@
...
+  required: true
+  input_type: user
+  description: Format of the generated report (CSV or YAML)
+  options: "csv\r\nyaml"
+  advanced: false
+model: ReportTemplate
+require:
+- plugin: katello
+￼ version: 3.9.0
+%>
```
ares: use - in here, otherwise I'm getting empty line as first thing in the report

report_templates/subscriptions.erb
```diff
@@ -0,0 +1,32 @@
...
+name: Subscriptions
+snippet: false
+template_inputs:
+- name: Subscriptions filter
+  required: false
+  input_type: user
+  description: Search query to filter subscription entitlements. E.g. `quantity >
+    0` to exclude subscriptions with unlimited entitlements. Keep unset for all subscriptions.
+  advanced: false
+- name: Output format
```
ares: Is yaml formatter stable enough already? I'm not sure I'd expose this as an input, once we think it's ready, we should have that as separate parameter for generating the report (as long as all reports will always support it). Overridding it per template using user input seems like something we'd need to remove in future.

ares (changes requested):
one nit and one thing to consider, otherwise works well (tested with patched katello)

report_templates/subscriptions.erb
mbacovsky: I consider yaml formatter stable enough as it is basically `to_yaml` called on a list of hashes with filtered content. I don't expect any troubles there. We didn't do a performance testing, but that was not done neither for the csv render. I'll remove the input for the format till we agree we want it there.

mbacovsky:
@ares, thanks for the review, updated.

ares (approved):

ares:
Thanks @mbacovsky, merging!

PR #528, opened 2018-10-12
Fixes #25185 - Dynamically determines which efi file to use for chain…

bennodepenno:
This patch could solve the pxegrub2 chainloading problem with UEFI systems. It now searches any hd*,gpt* three levels deep for efi files. When found, it will attempt a chainload.


lzap (approved):
Nice, gonna test this now with RHEL.

lzap:
So it works, but it's a bit slow. It is searching cd0 and fd0 for some time, randomly this can be as slow as 20 seconds. Do you have any ideas how to filter out CDROM and floppy from this? I don't see why this is actually searched, you explicitly say GPT and MSDOS there.

https://youtu.be/1d8akPd9ui8

provisioning_templates/snippet/pxegrub2_chainload.erb
```diff
@@ -5,22 +5,37 @@ model: ProvisioningTemplate
...
+  insmod fat
+  insmod chain
+
+  <% efi_file_list.each do |search_efi| %>
+  if [ $efi_found != true ]; then
+    echo "Search system for <%= search_efi %>"
+    for efi in (hd*,gpt*)/efi/*/*.efi (hd*,gpt*)/efi/*/*/*.efi (hd*,gpt*)/*.efi (hd*,gpt*)/*/*.efi ; do
+      regexp --set=1:efi_device '^\((.*)\)/' "${efi}"
+      regexp --set=1:efi_path '(.*)/.*$' "${efi}"
+      regexp --set=1:efi_file '.*\/(.*)$' "${efi}"
```
lzap, line 21: Can't we use `search` command for that? This is a bit hacky, I am wondering why `search` is not used.

bennodepenno: the search command does not accept wildcards, one has to specify the exact location of the efi file you're searching. It would require many statements (at least one for each distribution) to determine the root disk. Although  I agree that the for loop is hacky (and I'm not a programmer.) indeed, it was one way I could get it working.

bennodepenno: As for the slowness, the for loop should have eliminated the fd0 and cd0 searches. Unfortunately, it simply ignores this.. Do you have any idea why?

lzap:
Sorry for the delay, here is what I think is cleaner solution to the problem:

https://github.com/theforeman/community-templates/pull/533

bennodepenno:
Great work. I’ll test it on a few different machines and let you know the result.

Grtz. Ben

bennodepenno:
It works like a charm. I've tested in both virtual environments and on different HP machines configured with UEFI boot (HP: Z240, Z230, Z4 , Z2 mini G3, Proliant DL380 Gen9, Proliant DL380 Gen 10) successfully. Great work!

October 25, 2018 3:17 PM, "Lukáš Zapletal" <notifications@github.com (mailto:notifications@github.com?to=%22Luk%C3%A1%C5%A1%20Zapletal%22%20<notifications@github.com>)> wrote:
	Sorry for the delay, here is what I think is cleaner solution to the problem: 

	#533 (https://github.com/theforeman/community-templates/pull/533)

lzap:
Thanks I am closing this one, let's get the other one into upstream.

PR #532, opened 2018-10-25
Fixes #25301 - support hosts without owner

ares:


ares:
this also fixes the other problem, if owner is hostgroup it does not respond to login, it's safer to use to_s which also works on nil

mbacovsky (approved):
Works fine, thanks @ares.

tbrisker (approved):
Thanks @ares !

tbrisker:
:cherries: to 1.20 as e0dc734

tbrisker:
and pulled into core as https://github.com/theforeman/foreman/commit/a25699d4e39778ab13ed681878ee8174975c03f6

ares:
That was quick, thanks @mbacovsky and @tbrisker

PR #542, opened 2018-12-04
Refs #25624 - improve performance for registered hosts

ares:
needs linked fix from core, the major improvement is the facts loading but also we can eager load interfaces and owner to avoid 2n+1 issue

### DIFF
--- a/report_templates/registered_hosts.erb
+++ b/report_templates/registered_hosts.erb
@@ -13,7 +13,7 @@ require:
 - plugin: katello
 ￼ version: 3.9.0
 -%>
-<%- load_hosts(search: input('Hosts filter'), includes: [:operatingsystem, :subscriptions, :fact_values, :fact_names, :applicable_errata, :applicable_rpms]).each_record do |host| -%>
+<%- load_hosts(search: input('Hosts filter'), includes: [:operatingsystem, :subscriptions, :kernel_release, :interfaces, :owner, :applicable_errata, :applicable_rpms]).each_record do |host| -%>
 <%-   report_row(
         'Name': host.name,
         'Ip': host.ip,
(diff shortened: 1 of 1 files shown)


report_templates/registered_hosts.erb
```diff
@@ -13,15 +13,15 @@ require:
 - plugin: katello
 ￼ version: 3.9.0
 -%>
-<%- load_hosts(search: input('Hosts filter'), includes: [:operatingsystem, :subscriptions, :fact_values, :fact_names, :applicable_errata, :applicable_rpms]).each_record do |host| -%>
+<%- load_hosts(search: input('Hosts filter'), includes: [:operatingsystem, :subscriptions, :kernel_release, :interfaces, :owner, :applicable_errata, :applicable_rpms]).each_record do |host| -%>
```
ekohl, line 16: interfaces is needed for `host.ip` to be available?

report_templates/registered_hosts.erb
```diff
@@ -13,15 +13,15 @@ require:
 - plugin: katello
 ￼ version: 3.9.0
```
ekohl, line 14: Should this be bumped to 3.9.2, assuming it'll be cherry picked?

ares: There's no need for katello change, it's core that needs to be updated

report_templates/registered_hosts.erb
ares: Correct, maybe loading primary interface would be enough, I'll check that

ekohl (approved):

report_templates/registered_hosts.erb
tbrisker: did you check? interfaces can be a big table to join sometimes...

ares: Yes, I did, it does not work, the problem is with `Host::Base#get_interface_by_flag`, only when we include all interfaces, we don't trigger n+1 issue. Since load_hosts works in bulks of 1000 records, the number of interfaces shouldn't be much higher than few thousands.

tbrisker: hmm, sounds like maybe we should revisit that method, but that's outside the scope for this PR.

tbrisker (approved):
Thanks @ares

PR #543, opened 2018-12-12
Fix 'medium_uri' in suse-template

m-bucher:
autoyast_default_ipxe did not work (needs to be cherry-picked to 1.20)
I have not tested it on 1.19, yet


ekohl (changes requested):
Could you make sure the file is not executable? You're changing the mode from 0644 to 0755 now.

m-bucher:
@ekohl thanks, just fixed the file-mode

ShimShtein (changes requested):
I think it was a bug in the core that was fixed by https://github.com/theforeman/foreman/pull/6316.

Can you please recheck with newer foreman release?
If it works, we can ask for a backport into 1.20 too.

m-bucher:
@ShimShtein, you are right when I add your change from theforeman/foreman#6316 to my foreman-1.20 it works.

I'd much appreciate the fix to be backported to 1.20

ShimShtein:
@m-bucher, done: https://github.com/theforeman/foreman/pull/6316#issuecomment-452302120

PR #544, opened 2018-12-14
Added rediscover host job template

lzap:
WIP - it does not work for me: error during rendering: wrong number of arguments (given 1, expected 0)

Some job templates guru help me out.

### DIFF
--- a/job_templates/rediscover_host.erb
+++ b/job_templates/rediscover_host.erb
@@ -1,0 +1,52 @@
+<%#
+name: Rediscover node
+model: JobTemplate
+job_category: Discovery
+description_format: Writes grub2 configuration with "Rediscover host" entry.
+snippet: false
+template_inputs:
+- name: boot_interface
+  description: Boot interface MAC address (interactive TUI when blank)
+  required: false
+  input_type: user
+- name: set_default
+  description: Set as the default boot item (set to non-blank value)
+  required: false
+  input_type: user
+  options: "yes"
+- name: reboot
+  description: Reboot immediately (set to non-blank value)
+  required: false
+  input_type: user
+  options: "yes"
+provider_type: SSH
+kind: job_template
+%>
+
+# This template requires two symlinks to be present on Foreman server:
+# ln -s /var/lib/tftpboot/boot/fdi-image/vmlinuz0 /var/www/html/pub/discovery-vmlinuz
+# ln -s /var/lib/tftpboot/boot/fdi-image/initrd0.img /var/www/html/pub/discovery-initrd.img
+
+cat >/etc/grub.d/90_discovery <<EODS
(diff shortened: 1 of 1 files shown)


iNecas:
I don't see the issue when testing it on my setup: the template renders just fine. Do other templates work on your setup?

job_templates/rediscover_host.erb
```diff
@@ -0,0 +1,50 @@
...
+snippet: false
+template_inputs:
+- name: boot_interface
+  description: Boot interface MAC address (interactive TUI when blank)
+  required: false
+  input_type: user
+- name: set_default
+  description: Set as the default boot item (set to non-blank value)
+  required: false
+  input_type: user
```
iNecas, line 15: I wonder if we shouldn't have here options specified, to indicate that we only expect bool value here

```
  options: "yes"
```

job_templates/rediscover_host.erb
```diff
@@ -0,0 +1,50 @@
...
+- name: reboot
+  description: Reboot immediately (set to non-blank value)
+  required: false
+  input_type: user
```
iNecas, line 20: Same suggestion with `options` here

lzap:
Ok thanks how about this.

iNecas:
I guess it wold be worth linking a redmine issue and describing there, how it should work? Perhaps even a note to the docs would be fine?

lzap:
Do we do RM for community templates really? ;-) I am lazy... But here is what you need: https://github.com/theforeman/theforeman.org/pull/1262

ekohl:
@lzap Redmine issues make it very easy for us in changelogs to list this as a feature. If you think users would find this by themselves I'm also fine with it but it can make it easier to spot for upgrading users.

lzap:
Sure, well this is just a prototype, I will let @iNecas to complete the review and then I am creating the ticket.

ares:
@lzap if you provide some information about how to use it, I'm willing to review and test. E.g. I'm not sure how to configure SSH on the discovered machine. Looks useful and easy to finish.

lzap:
The process:

0) Create those symlinks (see the comments) in apache http directory
1) Install a host (preferably via discovery)
2) Run this template via SSH
3) The host will reboot back into discovery so you can reprovision it

When this was created, we did not have HTTP Boot feature yet, now since we have it, if there is a helper that would return Smart Proxy URL which is handling the SSH connection, we could simply download the files from the HTTP Smart Proxy endpoint. This would not require manual step (0).

lzap:
We should have that information if `@host` is available in this context, is it? Then I can use `@host.provision_interface.subnet.httpboot_proxy` to get the URL. We can use HTTPS and ignore the certificate, or even better deploy the certificate before downloading the image.

tbrisker:

PR #549, opened 2019-01-29
Handle zypper info result codes

sbernhard:
Some zypper results codes are just "information" result codes which
don't abort but are shown if the action was still successfull

### DIFF
--- a/job_templates/package_action_-_ssh_default.erb
+++ b/job_templates/package_action_-_ssh_default.erb
@@ -52,6 +52,29 @@ exit_with_message () {
   exit $2
 }
 
+<% if package_manager == 'zypper' -%>
+handle_zypp_res_codes () {
+  RETVAL=$1
+
+  # See https://github.com/openSUSE/zypper/blob/master/src/main.h
+  declare -A ZYPP_RES_CODES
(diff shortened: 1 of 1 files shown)


job_templates/package_action_-_ssh_default.erb
```diff
@@ -50,7 +50,30 @@ template_inputs:
 exit_with_message () {
   echo "${1}, exiting..."
   exit $2
-}
+  }
```
quba42: Is this intentional? (The same indentation also exists in line 75)

job_templates/package_action_-_ssh_default.erb
```diff
@@ -52,6 +52,29 @@ exit_with_message () {
   exit $2
 }
 
+<% if package_manager == 'zypper' -%>
+handle_zypp_res () {
+  RETVAL=$1
+
+  # See https://github.com/openSUSE/zypper/blob/master/src/main.h
+  declare -A ZYPRES
```
mdellweg: For consistency i would call that variable `ZYPP_RES` or even more descriptive `ZYPP_RES_CODES`.

ekohl (approved):

PR #550, opened 2019-02-02
Add template for Ansible Roles, supporting hashes and arrays

fcarrus:
Many Ansible Roles expect variables as structures, not just a plain string value (i.e. system-roles.timesync's **timesync_ntp_servers** variable).

This template helps by importing all the Host parameters as Ansible first-level variables, but trying to decode the strings values as a JSON structure. If it does it successfully, it is carried as a `vars:` child in YAML syntax, to be used by the Roles.

Example Parameters:
```
timesync_ntp_servers = [ { "hostname": "ntp1.inrim.it", "iburst": "yes" }, { "hostname": "ntp2.inrim.it", "iburst": "yes" } ]
other-parameter-hash = { "this": "is", "an": "array", "of": [ "hashes", "and", "hashes" ] }
```
would render like this:
```yaml
---
- name: Ansible Roles - Ansible with Foreman Parameters
  vars:
    kt_env: Library
    kt_cv: RHEL76
    foreman_host_collections: []    
    lifecycle_environment: Library
    timesync_ntp_servers: 
    - hostname: ntp1.inrim.it
      iburst: 'yes'
    - hostname: ntp2.inrim.it
      iburst: 'yes'
    other_parameter_hash: 
      this: is
      an: array
      of:
      - hashes
      - and
      - hashes
 ...
```

Any code style suggestion or optimization is well accepted, since I'm pretty new to ERB.


ares:
The latest version of ansible plugin adds ansible variables, that support all types natively. While this is great work, there's also a PR introducing types to generic parameters. Would either of that solve your usecase?

fcarrus:
Great, I didn't know about that. 
Sure I'll wait for something more *native*.  
Thanks!

PR #554, opened 2019-02-12
Added eject command as opt-in

lzap:


ekohl (changes requested):
Wouldn't it be cleaner to modify the existing reboot section which already has reject for bootdisk: https://github.com/theforeman/community-templates/blob/e541771837b72974e4d755775472eb75aec2f80e/provisioning_templates/provision/kickstart_default.erb#L136-L140

lzap:
Uh I haven't noticed :-) Closing this is useless then.

PR #555, opened 2019-02-14
Support for non-EFI Grub2

lzap:
Grub2 templates are not usable in non-EFI world like IBM POWER or i386 legacy PC. This patch adds that.


lzap:
Anyone? Hello!

ekohl (approved):

PR #556, opened 2019-02-22
Preseed python if ansible provisioning configured

mdellweg:


mdellweg:
I know, i 'made' the tests pass. But i guess, the `@host.all_ansible_roles` method is only available when the ansible plugin is installed. So my question becomes: Is there a possibility to check for active plugin, or can you do something like `@host.try :all_ansible_roles` ?

ares:
I think it would be good to add new macro, something like `plugin_installed?('foreman_ansible') # => true`, if you would like to extend the DSL, it would probably belong to lib/foreman/renderer/scope/macros/base.rb in Foreman.

Also it would be good to follow other [cfgmgmt patterns](https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/provision/kickstart_default.erb#L41-L43) so the variable should be called ansible_enabled. You can use the `respond_to?` trick like we use for chef. But adding explicit plugin_installed? macro would be IMHO good anyway.

I can also imagine we could use host parameter "additional packages" to avoid need of changing the template if users want to install more things by default.

mdellweg:
Thanks, the `respond_to?` should do the trick. Also i changed the logic from 'actual ansible roles selected' to 'ansible plugin installed'.

ares:
I defer to others if it's a good idea to install python by default if we detect ansible, but I'd lean towards yes :-) perhaps we should do something similar n kickstart too?

provisioning_templates/provision/preseed_default.erb
```diff
@@ -122,7 +123,7 @@ d-i apt-setup/local<%= repos %>/key string http://keyserver.ubuntu.com/pks/looku
 tasksel tasksel/first multiselect minimal, ssh-server, openssh-server
 
 # Install some base packages
-d-i pkgsel/include string <%= salt_package %> lsb-release
+d-i pkgsel/include string <%= salt_package %> <%= ansible_enabled ? 'python' : '' %> lsb-release <%= host_param('additional-packages') %>
```
ekohl: Will there be an issue with python vs python3? Would it be an idea to use list?
```ruby
additional_packages = ['lsb-release']
additional_packages << host_param('additional-packages')
additional_packages << 'salt-minion' if salt_enabled
additional_packages << 'python' if ansible_enabled
```

Not sure if our ERB sandbox allows this.

mdellweg:
> I defer to others if it's a good idea to install python by default if we detect ansible, but I'd lean towards yes :-) perhaps we should do something similar n kickstart too?

I decided for yes, because having python installed is not a big deal, and even if you do not have ansible roles defined at provisioning, you might wnat to use ansible for rex-tasks.
I think, for kickstart this is not a problem, because yum depends on python anyway.

provisioning_templates/provision/preseed_default.erb
mdellweg: It looks cleaner. ;)

mdellweg: There is no problem having both python 2 and 3 installed simultaneously.

mdellweg:
Not sure, why the test is failing. Is `host_param` with a default value a new feature?

provisioning_templates/provision/preseed_default.erb
```diff
@@ -8,9 +8,16 @@ oses:
...
+  ansible_enabled = @host.respond_to?(:all_ansible_roles)
   salt_enabled = host_param('salt_master') ? true : false
   os_major = @host.operatingsystem.major.to_i
   squeeze_or_older = (@host.operatingsystem.name == 'Debian' && os_major <= 6)
+
+  additional_packages = ['lsb-release']
+  additional_packages << host_param('additional-packages')
+  additional_packages << 'python' if ansible_enabled
+  additional_packages << 'salt-minion' if salt_enabled
+  additional_packages = additional_packages.join(" ").split(" ").uniq().join(" ")
```
ekohl: Isn't `.split()` safer so you split on any whitespace?

ekohl (approved):

provisioning_templates/provision/preseed_default.erb
mdellweg: I you say so...

ekohl:
LGTM. @ares could you have a final look?

provisioning_templates/provision/preseed_default.erb
```diff
@@ -8,9 +8,16 @@ oses:
 %>
 <%
   proxy_string = host_param('http-proxy') ? " http://#{host_param('http-proxy')}:#{host_param('http-proxy-port')}" : ''
+  ansible_enabled = @host.respond_to?(:all_ansible_roles)
```
ares: should this also check whether some roles are assigned? I think this returns true for each host, after ansible plugin is installed

mdellweg: That helps, if you decide to add roles later.

mdellweg: ... or if you want to use ansible for remote-execution tasks.

ares: Oh, that's a good point. So, how about we add `plugin_present?` macro in core? That's an easy addition and much cleaner approach :-) I can help with that if the need be.

mdellweg: I like the idea of `plugin_present?`and we should definitely do that, but i am not sure we should wait here, as this one is working with today's foreman.
(I begun to dislike interdependent PR's a bit...)

mdellweg: https://github.com/theforeman/foreman/pull/6543

ares: I promise I'll help to get this PR in. Once we have something in, we never refactor...

mdellweg: @ares Thanks for keeping that promise.
Ready for rereview.

ares (approved):
Thank you @mdellweg, works well, merging! Thanks for doing the extra step and implemented the new macro!

PR #557, opened 2019-02-25
Improved iPXE EFI discovery workflow and netboot

lzap:
This aims to improve two areas:

* Global iPXE template for discovery in EFI mode. This was not working due to missing `initrd=` kernel command line option. The patch adds more options to boot discovery from Foreman server, proxy or custom script. Also adds some debugging information - image status and sleep 2 seconds to actually see the results.

* Boot order in EFI mode - until now we ignored this problem. This patch adds opt-in bootorder host param. Users can now set back the previous boot order (PXE usually) or select a specific entry via grep.


provisioning_templates/finish/preseed_default_finish.erb
```diff
@@ -53,4 +53,5 @@ oses:
 <% end -%>
 
 <%= snippet 'preseed_networking_setup' %>
+<%= snippet 'efibootmgr_netboot' %>
```
mmoll, line 57: was this tested with Debian or Ubuntu?

lzap: It was not, however this is opt-in only. No changes are expected if no host param is provided.

lzap:
Rebased.

mmoll:
merged, díky @lzap!

PR #558, opened 2019-02-26
Fixes #26033 - Do not eager load polymorphic owner

xprazak2:


report_templates/registered_hosts.erb
```diff
@@ -13,7 +13,7 @@ require:
 - plugin: katello
 ￼ version: 3.9.0
 -%>
-<%- load_hosts(search: input('Hosts filter'), includes: [:operatingsystem, :subscriptions, :kernel_release, :interfaces, :owner, :applicable_errata, :applicable_rpms]).each_record do |host| -%>
+<%- load_hosts(search: input('Hosts filter'), includes: [:operatingsystem, :subscriptions, :kernel_release, :interfaces, :applicable_errata, :applicable_rpms], joins: "left outer join users on users.id = hosts.owner_id and hosts.owner_type='User' left outer join usergroups on usergroups.id = hosts.owner_id and hosts.owner_type='Usergroup'").each_record do |host| -%>
```
ares: Thanks for the fix, few comments

0) why it does not work through includes? I was not aware of the fact polymorphic relations can't be eager loaded
1) this moves it from includes to joins, meaning data will not be eager loaded. Does this trigger SQL N+1 problem? if we don't include, we don't need the join at all I think
2) could SQL code be upcased?
3) could you break it so that each keyword is on separate line? it became too long with this

ares (changes requested):

report_templates/registered_hosts.erb
timogoebel: In core we simply use `Host.all.preload(:owner)`, see https://github.com/theforeman/foreman/blob/465ef00a13471f45f68ed0220e604722c6c3601f/app/views/dashboard/_hosts_in_build_mode_widget.html.erb#L1.

xprazak2:
I changed it to use preload.

ares:
looks good, we now need to modify the PR in core to allow this in macro right?

ares:
so I was trying to test the fix but I'm unable to reproduce, do you have some reproducing environment? I'm trying to understand what exactly fails

ares (approved):

ares:
Thanks @xprazak2, fixes the issue, merging!

PR #559, opened 2019-02-27
Fixes #26172 - Added ability to override repository with cost

tchellomello:
Added ability to set a cost for all repositories during build time (cloud-init, kickstart) which will force `yum` or `dnf` to prioritize which repository to pick when the two repositories offer the same content. 

Redmine Tracker: https://projects.theforeman.org/issues/26172

This worked effectively and increased the speed on cloud-init builds.


provisioning_templates/snippet/redhat_register.erb
```diff
@@ -179,6 +181,12 @@ snippet: true
     <%= "subscription-manager repos --enable #{host_param('subscription_manager_repos').gsub(',', ' --enable')}" %>
   <% end %>
 
+  <% if host_param('subscription_manager_override_repos_cost') %>
```
lzap, line 184: Please use `host_param_true?` helper which parses "true" and "yes" and stuff.

lzap: Oh it contains the actual cost, scratch that!

lzap (approved):
It works fine.

PR #560, opened 2019-03-01
Add FIPS support

ares:
This takes custom logic from https://github.com/RedHatSatellite/satellite6-fips-client so we have FIPS support available in upstream templates.

@sideangleside, mind to check whether I missed something? Potentially, @pondrejk would you mind testing?

### DIFF
--- a/provisioning_templates/snippet/fips_packages.erb
+++ b/provisioning_templates/snippet/fips_packages.erb
@@ -1,0 +1,8 @@
+<%#
+kind: snippet
+name: fips_packages
+model: ProvisioningTemplate
+snippet: true
+%>
+dracut-fips
+-prelink
--- a/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
+++ b/provisioning_templates/PXEGrub/kickstart_default_pxegrub.erb
@@ -57,6 +57,10 @@ oses:
     options.push('inst.stage2=' + @host.operatingsystem.medium_uri(@host).to_s)
   end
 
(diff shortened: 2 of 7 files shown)


iNecas (changes requested):
Test failures related

iNecas:
Should the templates for other bootloaders (PXEGrub, iPXE) udpated as well?

provisioning_templates/snippet/fips_packages.erb
```diff
@@ -0,0 +1,6 @@
+<%#
+kind: snippet
+name: fips_packages
```
iNecas, line 3: missing `model` and `snippet` attributes

ares:
This should fix tests and add the same to PXEGrub and PXEGrub2. I'm not sure about iPXE, IIUC fips is required only in stage2? @lzap could you advise?

iNecas:
Created an issue for this for better tracking https://projects.theforeman.org/issues/26203

Tested and works well: the provisioned host had fips enabled after provisioning and the installation worked without issues.

Deferring to @lzap regarding the iPXE question.

Also opened discovery counterpart https://github.com/theforeman/foreman_discovery/pull/473 - reviews welcome there.

lzap:
Well iPXE works the same way as PXELinux, except it uses http by default in our case. So sure, add the parameter there as well, host params are available in the iPXE context. There is no FIPS mode in iPXE itself.

ares:
ok, I updated the kickstart_default_ipxe.erb as well, please take another look

iNecas (approved):
All good now. Thanks @ares

iNecas:
Tested and works well. Thanks @ares

PR #561, opened 2019-03-04
Fixes #26220 - Added vmware_image_provisioning snippet

tchellomello:
Added snippet to facilitate the VMware image-based provisioning which will include the snippet to the User Data template during build time (cloud-init, kickstart).

Redmine Tracker: https://projects.theforeman.org/issues/26220

Snippet example:
```
# https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.vm.customization.Specification.html
# encryptionKey expects an array
# globalIPSettings expects a hash, REQUIRED
# identity expects an hash, REQUIRED
# nicSettingMap expects an array
# options expects an hash

#   *   LinuxPrep: <~Hash> - Optional, contains machine-wide settings (note the uppercase P)
#   *     domain: <~String> - REQUIRED, The fully qualified domain name.
#   *     hostName: <~String> - REQUIRED, the network host name
#   *     hwClockUTC: <~Boolean> - Optional, Specifies whether the hardware clock is in UTC or local time
#   *     timeZone: <~String> - Optional, Case sensistive timezone
identity:
  LinuxPrep:
    domain: tatu.home
    hostName: neal-goldner
    hwClockUTC: true
    timeZone: America/New_York

#   * globalIPSettings <~Hash> - REQUIRED
#   *   dnsServerList <~Array> - Optional, list of dns servers - Example: ["10.0.0.2", "10.0.0.3"]
#   *   dnsSuffixList <~Array> - Optional, List of name resolution suffixes - Example: ["dev.example.com", "example.com"]
globalIPSettings:
  dnsSuffixList: [tatu.home]
  dnsServerList: [192.168.169.51,192.168.169.156]
#   * nicSettingMap: <~Array> - Optional, IP settings that are specific to a particular virtual network adapter
#   *   Each item in array:
#   *   adapter: <~Hash> - REQUIRED, IP settings for the associated virtual network adapter
#   *     dnsDomain: <~String> - Optional, DNS domain suffix for adapter
#   *     dnsServerList: <~Array> - Optional, list of dns server ip addresses - Example: ["10.0.0.2", "10.0.0.3"]
#   *     gateway: <~Array> - Optional, list of gateways - Example: ["10.0.0.2", "10.0.0.3"]
#   *     ip: <~String> - Optional, but required if static IP
#   *     ipV6Spec: <~Hash> - Optional, IPv^ settings
#   *       ipAddress: <~String> - Optional, but required if setting static IP
#   *       gateway: <~Array> - Optional, list of ipv6 gateways
#   *     netBIOS: <~String> - Optional, NetBIOS settings, if supplied must be one of: disableNetBIOS','enableNetBIOS','enableNetBIOSViaDhcp'
#   *     primaryWINS: <~String> - Optional, IP address of primary WINS server
#   *     secondaryWINS: <~String> - Optional, IP address of secondary WINS server
#   *     subnetMask: <~String> - Optional, subnet mask for adapter
#   *   macAddress: <~String> - Optional, MAC address of adapter being customized. This cannot be set by the client
nicSettingMap:

  - adapter:
      dnsDomain: tatu.home
      dnsServerList: [192.168.169.51,192.168.169.156]
      gateway: [192.168.169.252]
      ip: 192.168.169.186
      subnetMask: 255.255.255.0

#cloud-config
timezone: America/New_York

hostname: neal-goldner
fqdn: neal-goldner.tatu.home
manage_etc_hosts: true
ssh_pwauth: true
groups:
- admin
users:
- default
- name: admin
  primary-group: admin
  groups: users
  shell: /bin/bash
  sudo: ['ALL=(ALL) ALL']
  lock-passwd: false
```

Tested and working


ares:
Any comments from @timogoebel @ezr-ondrej @Toledo @lzap? Is this somehow related to [foreman_userdata](https://github.com/theforeman/foreman_userdata), looks like the same template type.

lzap:
Yeah I think you want foreman_userdata instead, our plan is to merge this into core and this should be our suggested way of doing image-based provisioning on VMWare.

lzap:
FYI the templates are here:

https://github.com/theforeman/foreman_userdata/tree/master/app/views/foreman/unattended/provisioning_templates

We will move them into community-templates.

timogoebel:
Yeah, thanks for your contribution.
We're currently working on moving the plugin to core. Please see https://github.com/theforeman/foreman/pull/6057 for details.
As you can see there is one UX question we still need to solve: Do users get that there is a new template kind called `cloud-init`? I'd appreciate any help here.

tchellomello:
@timogoebel @lzap Thanks for your comment on this PR.  It's ok, we can close this one and work on the theforeman/foreman#6057. 

One thing that I believe it is important to address it the lines https://github.com/theforeman/foreman/pull/6057/files#diff-ec6da2a9525882ebbdb269373d708317R24 and https://github.com/theforeman/foreman/pull/6057/files#diff-ec6da2a9525882ebbdb269373d708317R32 as in some cases the  
DNS primary or secondary are not specified and since both files are not returning `nil` but an empty string, it will create the result `[ , ]` which will not be a valid `YAML` file.  I'll make this suggestion inline on that PR too. 

Based on my experience, `cloud-init` is a more common term for this kind of template. I think users will understand this without problems.

lzap:
Thanks again.

PR #563, opened 2019-03-08
Refs #26127 - added 'nokaslr' kernel option

lzap:
According to the feedback this vastly improves chances that kexec will work on various bare-metal hardware.


ekohl (approved):

mmoll:
merged, díky @lzap!

PR #566, opened 2019-03-27
Fixes #26467 - support free text host filter

ares:
@lhellebr, it would be nice to confirm this fixes the issue you found


lhellebr:
Based on provided environment (Sat 6.5 snap 21 with the template from this PR), this PR seems to fix the issue.

ekohl (approved):

PR #567, opened 2019-03-28
autoyast_default: add conditional for leap 15 ntp client config

mattiascockburn:
This commit adds a conditional which differentiates between openSUSE Leap 15 and older versions.
This is needed because with Leap 15 the default NTP client is chrony and the configuration flags in AutoYaST are incompatible with older versions.


mattiascockburn:
This was tested successfully on Leap 42.3 and 15.

mmoll:
merged, danke @mattiasgiese!

PR #569, opened 2019-03-29
autoyast_default: enable installation of upstream puppet package

mattiascockburn:
This commit enables the use of the upstream puppet-agent package on openSUSE Leap. Although Puppet Inc. only packages for SLES, these packages work flawlessly on Leap as well.


mmoll:
@mattiasgiese Please check this comment: https://github.com/theforeman/community-templates/pull/447#issuecomment-373821186

Long story short: As long as SUSE doesn't claim that binaries of SLES packages are compatible with specific OpenSUSE versions, I think this change should not get in

PR #570, opened 2019-04-03
Static flag URL corrected

lzap:
The code is generating this:

    http://xxxx:3000/unattended/finish&static=true

while it should be really this:

    http://xxxx:3000/unattended/finish?static=true


ekohl:
Are you sure the finish URL never includes any URL parameters? Could it have a `?token=...` resulting in `?token=...?static=true`?

IMHO the `foreman_url` function should accept a hash and do the URL generation for you.

lzap:
Right, we should do the hash. This probably only hits you when you set token_duration to zero.

mmoll:
@lzap what's the status here?

lzap:
Once https://github.com/theforeman/foreman/pull/7086 is merged I will make use of it here.

ekohl:
That PR is now merged. Setting to waiting on contributor.

lzap:
Done. Can you spot more places where we could take advantage of this new param?

ekohl:
https://github.com/theforeman/community-templates/blob/a4c603a6e87b1cfb1ba5cb9d40f20296ffc2ef94/provisioning_templates/iPXE/kickstart_default_ipxe.erb#L52

There's also one case of calling foreman_url without a parameter which should be deprecated, right?
https://github.com/theforeman/community-templates/blob/a4c603a6e87b1cfb1ba5cb9d40f20296ffc2ef94/provisioning_templates/provision/junos_default_slax.erb#L407

lzap:
I am not sure what you want to deprecate, we still default to "provision" :

    def foreman_url(action = 'provision', params = {})

ekohl:
It's a separate issue, but I'll expand a bit. IMHO `foreman_url` is ambiguous and we should always call it with an explicit action. That's why I'd like to deprecate calling it without one. Not something that needs to be fixed in this PR, but something I noticed while searching for `foreman_url` invocations.

lzap:
Ah ok, well I can add explicit argument in this PR if you want. It's only one place.

ekohl:
It'd be nice to make this the generic "fix foreman_url"-invocation PR. After it's merged, we should deprecate calling foreman_url without arguments.

lzap:
Sorry, I don't understand what you want me to do here. Add a deprecated warning? Or fix the single place where we don't use explicit parameter? Both?

ekohl:
Right now I'm happy with fixing the single place. The deprecation warning should go into core, but doesn't have to happen now.

lzap:
I believe this can be merged now? Prepared the deprecation PR: https://github.com/theforeman/community-templates/pull/570

ekohl (approved):

ekohl:
Thanks!

provisioning_templates/iPXE/ipxe_intermediate_script.erb
```diff
@@ -11,7 +11,7 @@ snippet: false
 :net<%= i %>
 isset ${net<%= i -%>/mac} || goto no_nic
 dhcp net<%= i -%> || goto net<%= i+1 %>
-chain <%= foreman_url('iPXE') %>?mac=${net<%= i -%>/mac} || goto net<%= i+1 %>
+chain <%= foreman_url('iPXE', mac: '${net0/mac}') %> || goto net<%= i+1 %>
```
tbrisker, line 14: I believe this should have been:
```erb
chain <%= foreman_url('iPXE', mac: "${net#{i}/mac}") %> || goto net<%= i+1 %>

lzap: Damn, thanks for that. https://github.com/theforeman/community-templates/pull/733

PR #571, opened 2019-04-05
Fix preseed_default_ipxe with DHCP and no subnet implies DHCP for all Preseed templates

damonmaria:
The preseed_default_ipxe template as it was, was hardcoded to only do static addresses. This allows both static and DHCP.

And across all Preseed templates if the host/interface has no subnet then that implies DHCP addressing.

From this discussion: https://community.theforeman.org/t/foreman-bootdisk-and-external-dhcp/7355/8

Fixes this issue: https://projects.theforeman.org/issues/26448

### DIFF
--- a/provisioning_templates/finish/preseed_default_finish.erb
+++ b/provisioning_templates/finish/preseed_default_finish.erb
@@ -15,10 +15,12 @@ oses:
 %>
 
 <% subnet = @host.subnet -%>
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
-<% dhcp = subnet.dhcp_boot_mode? && !@static -%>
+<% if @static -%>
+  <%- dhcp = false -%>
+<% elsif subnet.nil? -%>
+  <%- dhcp = true -%>
 <% else -%>
-<% dhcp = !@static -%>
+  <%- dhcp = subnet.respond_to?(:dhcp_boot_mode?) && subnet.dhcp_boot_mode? -%>
 <% end -%>
 <% unless dhcp -%>
 # host and domain name need setting as these values may have come from dhcp if pxe booting
--- a/provisioning_templates/finish/preseed_default_finish.erb
+++ b/provisioning_templates/finish/preseed_default_finish.erb
@@ -20,7 +20,7 @@ oses:
 <% elsif subnet.nil? -%>
   <%- dhcp = true -%>
 <% else -%>
-  <%- dhcp = subnet.respond_to?(:dhcp_boot_mode?) && subnet.dhcp_boot_mode? -%>
+  <%- dhcp = subnet.dhcp_boot_mode? -%>
 <% end -%>
 <% unless dhcp -%>
 # host and domain name need setting as these values may have come from dhcp if pxe booting
--- a/provisioning_templates/iPXE/preseed_default_ipxe.erb
+++ b/provisioning_templates/iPXE/preseed_default_ipxe.erb
@@ -8,16 +8,25 @@ oses:
 - Ubuntu
 %>
 <% if @host.operatingsystem.name == 'Debian' -%>
-<% keyboard_params = "auto=true domain=#{@host.domain}" -%>
+  <%- keyboard_params = "auto=true domain=#{@host.domain}" -%>
 <% else -%>
-<% keyboard_params = 'console-setup/ask_detect=false console-setup/layout=USA console-setup/variant=USA keyboard-configuration/layoutcode=us' -%>
+  <%- keyboard_params = 'console-setup/ask_detect=false console-setup/layout=USA console-setup/variant=USA keyboard-configuration/layoutcode=us' -%>
 <% end -%>
+
+<% subnet = @host.subnet -%>
+<% if subnet.nil? || (subnet.respond_to?(:dhcp_boot_mode?) && subnet.dhcp_boot_mode?) -%>
+  <%- provision_url_suffix = '' -%>
+  <%- netcfg_args = '' -%>
+<% else -%>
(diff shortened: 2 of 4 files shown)


provisioning_templates/finish/preseed_default_finish.erb
```diff
@@ -15,10 +15,12 @@ oses:
...
 <% subnet = @host.subnet -%>
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
-<% dhcp = subnet.dhcp_boot_mode? && !@static -%>
+<% if @static -%>
+  <%- dhcp = false -%>
+<% elsif subnet.nil? -%>
+  <%- dhcp = true -%>
 <% else -%>
-<% dhcp = !@static -%>
+  <%- dhcp = subnet.respond_to?(:dhcp_boot_mode?) && subnet.dhcp_boot_mode? -%>
```
lzap: Why would subnet not respond to `dhcp_boot_mode`? It's either subnet class or `nil`, you can use the `subnet&.dhcp_boot_mode?` syntax to cover nil case but that's pretty much all I can see. Did you encounter different object than subnet there?

provisioning_templates/iPXE/preseed_default_ipxe.erb
```diff
@@ -8,16 +8,25 @@ oses:
...
 <% if @host.operatingsystem.name == 'Debian' -%>
-<% keyboard_params = "auto=true domain=#{@host.domain}" -%>
+  <%- keyboard_params = "auto=true domain=#{@host.domain}" -%>
 <% else -%>
-<% keyboard_params = 'console-setup/ask_detect=false console-setup/layout=USA console-setup/variant=USA keyboard-configuration/layoutcode=us' -%>
+  <%- keyboard_params = 'console-setup/ask_detect=false console-setup/layout=USA console-setup/variant=USA keyboard-configuration/layoutcode=us' -%>
 <% end -%>
+
+<% subnet = @host.subnet -%>
+<% if subnet.nil? || (subnet.respond_to?(:dhcp_boot_mode?) && subnet.dhcp_boot_mode?) -%>
```
lzap: Ditto, we should not be storing any other object than Subnet in `@host.subnet`. That sounds like a bug to me.

provisioning_templates/finish/preseed_default_finish.erb
damonmaria: To be honest, this is my first foray into Ruby and Foreman. The `subnet.respond_to?(:dhcp_boot_mode?)` check was there previously so I've just maintained it. If, as you say, it's always a subnet and it's already be checked for `nil` then I presume this check can be removed?

damonmaria:
```suggestion
  <%- dhcp = subnet.dhcp_boot_mode? -%>
```

provisioning_templates/snippet/preseed_networking_setup.erb
```diff
@@ -8,9 +8,9 @@ description: this will configure your host networking, it configures your primar
     called in your preseed finish template.
 %>
 <% host_subnet = @host.subnet -%>
-<% host_dhcp = host_subnet.nil? ? false : host_subnet.dhcp_boot_mode? -%>
+<% host_dhcp = host_subnet.nil? ? true : host_subnet.dhcp_boot_mode? -%>
```
damonmaria, line 11: @lzap I've also just noticed that preseed_default.erb [passes the `static` parameter onto the finish template](https://github.com/theforeman/community-templates/blob/44a23f5d51982e8b3c0e30c87b5a5a02c2a33950/provisioning_templates/provision/preseed_default.erb#L152) but preseed_networking_setup.erb doesn't honor it (preseed_default_finish.erb does). Should it also be checking `static` parameter in several places in here too?

provisioning_templates/finish/preseed_default_finish.erb
damonmaria: Committed all `subnet.respond_to?(:dhcp_boot_mode?)` checks removed.

provisioning_templates/finish/preseed_default_finish.erb
```diff
@@ -15,10 +15,12 @@ oses:
 %>
 
 <% subnet = @host.subnet -%>
-<% if subnet.respond_to?(:dhcp_boot_mode?) -%>
```
lzap, line 18: @ares do you happen to know why subnet would not respond to this? Is this just to prevent `nil` error?

provisioning_templates/snippet/preseed_networking_setup.erb
lzap: Unless @mmoll knows more about why "static" option is added for preseed, I believe that "static" flag was initially added for gPXE/iPXE and it's not longer very tested and might not work at all. https://projects.theforeman.org/issues/470

Honoring it is probably good idea.

provisioning_templates/finish/preseed_default_finish.erb
ekohl: subnet can indeed be nil, but I wonder if we also introduced this a long time ago to be compatible with Foreman versions that didn't have the boot mode method yet.

lzap: I thought so, we are good then.

lzap:
Honoring static is probably for a different PR, let's not make any changes in this one. Any comments @mmoll ? I am gonna test this.

lzap:
Thanks, merged. I am going also fix the kickstart.

lzap:
https://github.com/theforeman/community-templates/pull/600

PR #572, opened 2019-04-05
[Ansible] Job uses GIT module to clone roles

UXabre:
Currently a git command is being executed, which causes the clone to fail if the roles folder is not empty; however, when running this as a recurring job, this fails after the first attempt.

By using the built-in GIT module, this problem seems to be fixed

### DIFF
--- a/job_templates/ansible_roles_-_install_from_git.erb
+++ b/job_templates/ansible_roles_-_install_from_git.erb
@@ -15,6 +15,11 @@ template_inputs:
   description: For example, '/etc/ansible/roles/foobar' . Look at '/etc/ansible/ansible.cfg'
     roles_path option to find what is your roles_path
   advanced: false
+- name: force
+  required: false
+  input_type: user
+  description: If yes, any modified files in the working repository will be discarded.
+  advanced: true
 provider_type: Ansible
 kind: job_template
 model: JobTemplate
@@ -23,6 +28,7 @@ model: JobTemplate
 ---
 - hosts: all
   tasks:
-    - command: git clone <%= input('git_repository') %> <%= input('location') %>
-      register: out
-    - debug: var=out
+    - git:
+        repo: <%= input('git_repository') %> 
+        dest: <%= input('location') %>
+        force: <%= input('force').present? ? input('force') : 'no' %>


ares:
I suppose the force=yes does the trick, though is that safe/expected behavior? doesn't it override existing data?

this is what documentation says 

```
If yes, any modified files in the working repository will be discarded. Prior to 0.7, this was always 'yes' and could not be disabled. Prior to 1.9, the default was `yes`
```

I suppose there's a good reason they switched it to no.

ares:
perhaps we could make this a template input to give user an option?

UXabre:
Hey @ares, actually a good point, safest would be to actually keep the default of no.
I'll add this as template input but default to no. (but after reading you comment, I completely loathe my original decision to set force to "yes")

job_templates/ansible_roles_-_install_from_git.erb
```diff
@@ -15,6 +15,11 @@ template_inputs:
   description: For example, '/etc/ansible/roles/foobar' . Look at '/etc/ansible/ansible.cfg'
     roles_path option to find what is your roles_path
   advanced: false
+- name: force
+  required: false
+  input_type: user
+  description: If yes, any modified files in the working repository will be discarded.
+  advanced: true
```
ares, line 22: Probably would be also good to specify options yes,no. I think if you set no as first, it will be default value

job_templates/ansible_roles_-_install_from_git.erb
```diff
@@ -23,6 +28,7 @@ model: JobTemplate
 ---
 - hosts: all
   tasks:
-    - command: git clone <%= input('git_repository') %> <%= input('location') %>
-      register: out
-    - debug: var=out
+    - git:
+        repo: <%= input('git_repository') %> 
+        dest: <%= input('location') %>
+        force: <%= input('force').present? ? input('force') : 'no' %>
```
ares, line 34: We could just paste input('force') here if you decide to specify options, see above

ares (commented):
One more suggestion, but looks better now :-)

job_templates/ansible_roles_-_install_from_git.erb
UXabre: ah yes, forgot those values. However, from looking in the UI, there doesn't seem to be a way to set a default value (which is a shame :-( ), this is what caused me to add a fallback value to the input('force') below.

But perhaps, there's a way to set a default out of my knowledge?

ares: There's no way to set default and yes, it's a shame :-) the only reasonable option is to specify options, with first being the preselected one. I moved the [tracking issue](https://projects.theforeman.org/issues/11367) from REX to core, it shouldn't be too complicated to add.

UXabre: Hey @ares! Great, will look into that one first and come back to this once merged, if that is okay for you?

ares: That would be awesome, feel free to ping me in that PR directly :-)

ares: @UXabre, the default value was added in Foreman 1.24, I think this was the last thing to finish this PR, could you please add that? since `input('force')` will now always contain the value, the comment below still holds

ares (changes requested):
switching to waiting on contributor, however very little is missing now to get this in, it's no longer blocked on core functionality

job_templates/ansible_roles_-_install_from_git.erb
UXabre: Allright! Let me read up on this default value and finish this template! Thanks for following up!

tbrisker:

PR #573, opened 2019-04-05
Update report template inputs metadata

ares:
This improves input value types definition, on seed, the type will be correctly updated. This should go to 1.22 templates.


ares:
To test: simply sync templates to your Foreman, run rake db:seed, try to generate templates, modified inputs should be either displaying auto complete searches :-)

tbrisker (approved):
code-wise looks reasonable, did not test

ezr-ondrej (approved):
Tested and works, I believe this is the best we can do for now :)

PR #574, opened 2019-04-06
Adding hostname parameter

donnydavis:
CoreOS Cloud Config was missing the hostname parameter, so they all come up as localhost without this parameter


ekohl (approved):

mmoll:
merged, thanks @donnydavis!

PR #575, opened 2019-04-06
Preseed ignore drives marked 'removable'

damonmaria:
Use `dmesg` to find drives marked as removable. This may not be bulletproof or future-proof but if the `dmesg` messages aren't what we are looking for then we are simply in the same situation we were before this PR. 

In my testing of this with Ubuntu 18.04 it correctly ignores both USB sticks and the 'virtual floppy' that Intel AMT creates, both of which the old code of picking the first of `list-devices disk` tried to installed to.

Also makes `grub-installer/bootdev` consistent with `partman-auto/disk`. At the moment (and I have experienced this) GRUB can attempt to be installed on a removable disk using `grub-installer/bootdev string default` and I would think the default should be the boot disk is the same one as the partitioned disk. One niggle I have with this is moving a `grub-installer` command inside the partitioning template which previously was only `partman` commands. But maybe for this setting it makes sense?

### DIFF
--- a/partition_tables_templates/preseed_default.erb
+++ b/partition_tables_templates/preseed_default.erb
@@ -18,11 +18,9 @@ oses:
 d-i partman-auto/disk string <%= host_param('install-disk') %>
 d-i grub-installer/bootdev string <%= host_param('install-disk') %>
 <% else -%>
-# Use the first detected hard disk not detected as removable
+# Use the first detected hard disk
 d-i partman/early_command string \
-  IGNORE_LIST="$(mktemp)"; \
-  dmesg | grep -io "\[[a-z]\+].*removable" | sed -r "s/\[([^]]+).*/\/dev\/\1/" > "$IGNORE_LIST"; \
-  INSTALL_DISK="$(list-devices disk | sort - "$IGNORE_LIST" "$IGNORE_LIST" | uniq -u | head -n1)"; \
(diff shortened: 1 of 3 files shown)


mmoll:
@damonmaria Could you open a separate PR for the `bootdev` change?

I think parsing `dmesg` is way too error-prone... I there maybe another command or some flag in `/sys` to tell, if a device is removable?

damonmaria:
@mmoll I can do the `bootdev` change separately. I'm new to PRs but presume this PR then be based on the `bootdev` one, as they will conflict a lot. 

Yes, I agree, I can totally see how parsing `dmesg` would be unreliable. But if this 'fails' then it will naturally fallback to picking the first drive, just as it does now. So in that way it will improve device selection in some (most?) situations, and degrade it in none. Although I guess it may be possible to have something in `dmesg` of the form `[sda] removable: false` which would trip this up. Hmmm....

Through `/sys` you can accurately figure out if a drive is removable or not (e.g. `/sys/block/sdb/removable` contains `1`) and I originally wrote this up to use that. But it seems that this `sys` info isn't there at the time `d-i partman/early_command` is run. That's as far as I could tell when it failed to work.

damonmaria:
At the request of @mmoll I've extracted the `bootdev` change from this into #584. I can resubmit this PR if needed for a nicer commit history.

PR #577, opened 2019-04-11
Fixes #26587 - improve performance of registered hosts report

ares:
Problem arises when user filters by hosts using google-like syntax e.g. "dell". Performance is good if they use "name ~ dell" or don't use the filter at all. But this is quite common. It also consumer a lot of memory. Removing the inclusion of applied rpms didn't have significant performance impact on other filtering.

### DIFF
--- a/report_templates/registered_hosts.erb
+++ b/report_templates/registered_hosts.erb
@@ -13,7 +13,7 @@ require:
 - plugin: katello
 ￼ version: 3.9.0
 -%>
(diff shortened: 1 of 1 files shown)


lhellebr:
I can confirm this makes the query much faster (20 minutes vs almost immediately). Also, the result is correct. There may be more fields worth omitting but for now, I consider the issue fixed.

lhellebr (approved):

tbrisker (approved):
Thanks @ares !

PR #580, opened 2019-04-15
Discovery template nesting fixed

lzap:
I noticed during core branching review that discovery is incorrectly nested. This fixes it.


tbrisker (approved):
Thanks @lzap !

PR #581, opened 2019-04-15
Removed snippet in snippet

lzap:
It was causing issues for renderer tests, I think I will keep it like this it's no big deal how discovery is added this is clean enough as well.


tbrisker (approved):
Thanks @lzap !

PR #583, opened 2019-04-18
Default roles install into /etc/ansible/roles

xprazak2:
@lpramuk, would this be ok?


ares:
Thanks @xprazak2, merging!

PR #584, opened 2019-04-29
Apply preseed grub-installer/bootdev in partition template

damonmaria:
At the moment `partman-auto/disk` and `grub-installer/bootdev` can end up pointing to different drives. This change moves setting `grub-installer/bootdev` into the preseed partition templates so it is consistent with how `partman-auto/disk` is calculated.

A potential issue with this is that users who have existing custom partition templates might now be missing the `d-i grub-installer/bootdev ...` setting (as it's no longer in the default provisioning template). I'm not sure how best to handle that. One possibility is to leave the setting of `d-i grub-installer/bootdev` in the default provision but I'm not sure how preseed will handle it being set again (and through `d-i partman/early_command`) in the partition template.

This was extracted from #575 at the request of @mmoll

### DIFF
--- a/partition_tables_templates/preseed_default.erb
+++ b/partition_tables_templates/preseed_default.erb
@@ -16,9 +16,13 @@ oses:
 
 <% if host_param('install-disk') -%>
 d-i partman-auto/disk string <%= host_param('install-disk') %>
+d-i grub-installer/bootdev string <%= host_param('install-disk') %>
 <% else -%>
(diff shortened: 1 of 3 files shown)


mmoll:
Existing non-default templates are always a problem, so nothing new here... :boom:

mmoll (approved):

mmoll:
merged, thanks @damonmaria!

PR #585, opened 2019-05-01
Fixes #26715 Append static arg back to provision URL

Rohan21Lobo:
Seems it was fine earlier but got affected in https://github.com/theforeman/community-templates/commit/f4225ecf9f90774180d1e0e0e46c236956ce3333


lzap:
Thanks, merge.

PR #593, opened 2019-06-27
Refs #27007 - Fix passing command to shell module

adamruzicka:
Ansible < 2.8.0 parses the command in a weird way and adds a
single space at the beginning of each line. This breaks scripts
which contain heredocs.

community-templates variant of https://github.com/theforeman/foreman_ansible/pull/275


ares:
Thanks @adamruzicka, merging despite the misaligned project, I don't think it make sense to clone in this case.

PR #594, opened 2019-07-03
Ensure wget is installed by preseed provision (Debian10)

m-bucher:
apparently Debian 10 does no longer install wget by default, however we need it to download the finish-template.


ekohl (approved):
As it stands a good change.

Do you know if curl is present by default and would it make sense to use that instead?

m-bucher:
@ekohl, `curl` was also not installed

ekohl:
This should have been merged a long time ago :(

PR #595, opened 2019-07-08
Add python2 module for sles 15 sp1

m-bucher:
SLES 15SP1 moved Python2 packages into a separate module.


provisioning_templates/provision/autoyast_sles_default.erb
```diff
@@ -245,6 +245,13 @@ rm /etc/resolv.conf
         <product_dir>/Module-Basesystem</product_dir>
         <product>sle-module-basesystem</product>
       </listentry>
+  <% if os_minor >= 1 -%>
```
lzap: I suggest to only put this if the param exists:

```suggestion
  <% if os_minor >= 1 && host_param('sle-module-basesystem-url') -%>
```

Also do you also check for major version somewhere? This will break old versions.

m-bucher: Probably makes sense to only add this if the parameter has been set, but SLES 15 and SLES 15 SP1 will not install if it is not specified, so either way the installation would fail ;-)
The major version is checked above on line 242

lzap: Yeah but this could make existing error "Repository/package not found" a different one like "XML parsing error: empty element XYZ". Just guessing, to be on the safe side. No biggie.

m-bucher: Added the check to both `if` statements

lzap (approved):
LGTM but I haven't tested. @mmoll would you do the honour?

PR #596, opened 2019-07-11
Adds redhat register snippet to cloud init template

san7ket:
Cloud init template was missing the redhat register snippet. 
So, boxes provisioned were not subscribed to the foreman. This helps make the cloud init template closer to the Kickstart default user data template
Preview of the template after snippet is added.
```
runcmd:
- |
  echo ##############################################################
  echo ################# SUBSCRIPTION MANAGER #######################
  echo ##############################################################
  echo
  echo "Starting the subscription-manager registration process"

    if [ -f /usr/bin/dnf ]; then
      dnf -y install subscription-manager
    else
      yum -t -y install subscription-manager
    fi

    rpm -Uvh http://foreman/pub/katello-ca-consumer-latest.noarch.rpm

    subscription-manager register --name="host.local" --org='Default_Organization' --activationkey='ak-rhel-7'

       if [ -f /usr/bin/dnf ]; then
         PACKAGE_MAN="dnf -y"
       else
         PACKAGE_MAN="yum -t -y"
       fi
      $PACKAGE_MAN install katello-agent
```

### DIFF
--- a/provisioning_templates/cloud_init/cloud_init_default.erb
+++ b/provisioning_templates/cloud_init/cloud_init_default.erb
@@ -27,6 +27,7 @@ users: {}
 <% if puppet_enabled %>
 runcmd:
 - |
+<%= snippet 'redhat_register' %>
 <% if host_param_true?('enable-puppetlabs-pc1-repo') || host_param_true?('enable-puppetlabs-repo') || host_param_true?('enable-puppetlabs-puppet5-repo') -%>
 <%= snippet 'puppetlabs_repo' %>
 <% end -%>


lzap:
Looks good, this is probably because @timogoebel haven't tested this with RHEL. My only concern is YAML format - my understanding of this block format is that YAML parses everything until empty newline, therefore we must get sure that the snippet does not (or never) contain empty spaces, otherwise we will get YAML parsing errors.

@san7ket have you tested this change end to end? We can merge this for now and probably create an extra test in core to YAML parse this template to make sure there will be no regressions. What you think Timo?

timogoebel:
> Looks good, this is probably because @timogoebel haven't tested this with RHEL.

I haven't tested this with katello. That's a slight difference. ;-)

> We can merge this for now and probably create an extra test in core to YAML parse this template to make sure there will be no regressions.

Ideally, we'd have proper tests in here. I believe we should work towards adding https://github.com/theforeman/foreman/pull/6700 in here.

san7ket:
@lzap yes I have tested this end-to-end, the box was provisioned and registered successfully

lzap:
> Ideally, we'd have proper tests in here. I believe we should work towards adding theforeman/foreman#6700 in here.

I'd rather merge this PR if you don't mind as Sanket is unlikely be willing to do bigger refactoring (surprise me please).

timogoebel:
> 'd rather merge this PR if you don't mind as Sanket is unlikely be willing to do bigger refactoring (surprise me please).

Sure, feel free to merge at your discretion.

lzap:
Thank you @san7ket for your contribution and extensive testing of this feature! Cloud-init endpoint is now officially part of Foreman core with QA dept assurance and official Red Hat support. If it breaks, @timogoebel, call in! (And I will be the one to fix this most likely, will probably ping ya hahaha).

provisioning_templates/cloud_init/cloud_init_default.erb
```diff
@@ -27,6 +27,7 @@ users: {}
 <% if puppet_enabled %>
 runcmd:
 - |
+<%= snippet 'redhat_register' %>
```
ekohl, line 30: Late review, but note that this is behind `<% if puppet_enabled %>`. I doubt you need Puppet to do RH register. Also note that this template is used on non-RH systems as well but I think the host parameter checks inside that take care of it.

lzap: Damn! Good find, @san7ket can you do another PR and fix this? Just move it out of the puppet_enabled block.

> I think the host parameter checks inside that take care of it.

Yep, I have checked that. Can confirm.

san7ket: ahh yes, will raise a fix

san7ket: Raised fix in #611

PR #601, opened 2019-07-18
All finish templates should call home and exit 0

lzap:
It looks like Foreman actually exits build mode on successful return code. We should take advantage of that. Also few finish templates were missing the callhome via wget/curl, we should explicitly do that when we can. Anyway if that fails, return code should do it anyway.

I left a comment explaining what is going on so users can edit templates and perform `exit 1` when needed.

https://github.com/theforeman/foreman/blob/develop/app/models/concerns/orchestration/ssh_provision.rb#L64


timogoebel:
Is the built call really necessary? If I understand the community thread correctly, it calls built when the script succeeds. We don't need to do that twice.

ekohl:
Wouldn't it actually break on environments where the host can't call back to Foreman where it worked before.

lzap:
@timogoebel I was thinking the same, but do we really want to drop this well understood concept?
@ekohl don't understand ;-) Is your comment on my original proposal or Timo's comment?

All and all, I'd rather keep the current status quo - exiting build mode twice cannot hurt :-)

timogoebel:
> All and all, I'd rather keep the current status quo - exiting build mode twice cannot hurt :-)

I think @ekohl is saying, that it can in fact break in the scenario he's describing.

ekohl:
Exactly. Imagine you have a cloud (like GCE) and a Foreman instance behind a NAT. If you do image based deployments, Foreman SSHs into the server and runs the finish script. If you add this callback, that would fail, the script would exit with an error code and the server isn't marked as built.

Perhaps the built snippet needs a conditional so it's a noop when Foreman is using SSH to finish. Then it can safely be included.

lzap:
> Exactly. Imagine you have a cloud (like GCE) and a Foreman instance behind a NAT. If you do image based deployments, Foreman SSHs into the server and runs the finish script. If you add this callback, that would fail, the script would exit with an error code and the server isn't marked as built.

But we don't execute the script with `-e` flag therefore shell don't exit on failed commands if I am not mistaken:

```ruby
  def command
    # Use the users home to store the provision script since we can't reliably
    # tell if other locations are writeable or executable by the user.
    main_execution = "(chmod 0701 ./#{remote_script} && #{command_prefix} ./#{remote_script} ; echo $? >#{remote_script}.status) 2>&1"
    "#{command_prefix} sh -c '#{main_execution} | tee #{remote_script}.log; exit $(cat #{remote_script}.status)'"
  end
```

We actually make it executable and exec it using shebang, therefore script would need to have `#!/bin/bash -e` and that's not the case:

```
$ grep '^#!' provisioning_templates/finish/*
provisioning_templates/finish/alterator_default_finish.erb:#!/bin/sh
provisioning_templates/finish/jumpstart_default_finish.erb:#!/bin/sh
provisioning_templates/finish/xenserver_default_finish.erb:#!/bin/bash
```

I believe my patch is correct and should not break anything. It's indeed hard to tell but I prefer little bit more consistent codebase here.

timogoebel:
> and that's not the case

Good catch. Can we add `set -e` to all scripts?
I'm not sure if we can add this in core as a finish script could in theory be ruby or python. It does not need to be a bash script, does it?

lzap:
Guys, I have finally figured out why we have built calls in finish scripts. These are used both in image-based provisioning when Foreman scp them, however they are also used for Debian provisioning to "finish off" provisioning since debian installer does not allow arbitrary scripts (only short commands). Hence the name. The same applies for Free BSD. I don't often test Debian or BSD therefore I was confused.

lzap:
> Good catch. Can we add `set -e` to all scripts?

Probably yes, shall I do it? Honestly I don't know what the consensus now is, options:

* keep status quo
* add exit 0 everywhere
* add built call home everywhere
* add both everywhere (current proposal)
* the same plus `set -e`

timogoebel:
I think `exit 0` and `set -e` everywhere would be the best option. If we just ask me, I think we also have a consensus. :-)
Kidding, do we want to discuss this here or bring it to discourse?

ares:
exit 0 and set -e combo sounds good to me, I didn't fully understand the need for second call home, IMHO it's only needed for preseed finish templates

lzap:
> IMHO it's only needed for preseed finish templates

BSD as well.

> I didn't fully understand the need for second call home

Only for consistency when someone takes a finish template as a blueprint for hers own one.

Let me rebase when Timo described.

timogoebel:
> Only for consistency when someone takes a finish template as a blueprint for hers own one.

I think @ekohl concern (https://github.com/theforeman/community-templates/pull/601#issuecomment-513830959) still applies.

ekohl (commented):
I also wonder if users prefer exit-early vs continuing. Personally I prefer knowing when it failed but maybe it should be configurable.

timogoebel:
> but maybe it should be configurable.

We could make this a host parameter

```erb
<% if param_true?('whatever') -%>
set -e
<% end -%>

lzap:
Thanks for the discussion, I am putting this into our documentation and I don't believe it is worth trying to unify this.

PR #603, opened 2019-07-30
Fixes #27470 - Provide template for checking for package updates

adamruzicka:
community-templates counterpart of https://github.com/theforeman/foreman_remote_execution/pull/422


ekohl:
Note the message is correct and you can't use Fixes here. That's because in Redmine it will end up only in one project and this way it won't actually close the issue.

lzap:
I am doing review in the other PR.

adamruzicka:
Moved the changes done in the REX PR here

lzap (approved):
It's a GO.

PR #605, opened 2019-08-02
Fix duplicate DEFAULT entry in PXELinux default local boot

### LINKED ISSUE #604: PXELinux default local boot does not show menu
I believe PR #483 introduced a bug by setting two `DEFAULT` entries in https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/PXELinux/pxelinux_default_local_boot.erb. This means the PXE menu is never displayed and instead the label `local_chain_hd0` is chosen immediately, without respecting the timeout or giving the user the opportunity to press ESC.

I suspect the second `DEFAULT` entry was really supposed to be `ONTIMEOUT`, as it was prior to PR #483, which makes much more sense.

I'll provide a PR to fix this.

antaflos:
Fixes GH-604


lzap (approved):
This sneaked in during my refactoring, thanks!

PR #607, opened 2019-08-09
Fixes #27578 - resolved syntax error in ifcfg_generic snippet

lzap:
Looks like we don't cover this branch of the template in our tests. Bummer.


provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
```diff
@@ -40,7 +40,7 @@ DEFROUTE=<%= primary %>
 <%-   if !@dhcp && @subnet -%>
 <%-     @subnet.dns_servers.each_index { |index| -%>
```
ekohl: Would it make sense to use `do` here since it's multi-line?
```suggestion
<%-     @subnet.dns_servers.each_index do |index| -%>
```

lzap: You actually bring the bug back with this suggestion :-) Anyway, I am fine with this.

lzap:
Rebased, to bring @tbrisker 's attention: we need to pull this into core templates. Thanks!

tbrisker (approved):
Thanks @lzap !

tbrisker:
pulled into 1.23-stable as a55d102 so we'll get it in the next rc (I sync the templates prior to every release)

provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
ekohl: I knew that, but didn't know how else to express it and (correctly) guessed you could figure it out :)

PR #613, opened 2019-09-05
Add Grubby Reprovision SSH job

lzap:
Tested, works great. Needs: https://github.com/theforeman/foreman/pull/7022

### DIFF
--- a/job_templates/run_command_-_grubby_reprovision.erb
+++ b/job_templates/run_command_-_grubby_reprovision.erb
@@ -1,0 +1,28 @@
+<%#
+kind: job_template
+name: Run Command - Grubby Reprovision
+model: JobTemplate
+job_category: Commands
+provider_type: SSH
+%>
+<%
+  if @host.build?
+-%>
+set -e
+KERNEL="/boot/grubby-reporovision-kernel"
+INITRD="/boot/grubby-reporovision-initrd.img"
+wget -O "$KERNEL" "<%= @host.url_for_boot(:kernel) %>"
+wget -O "$INITRD" "<%= @host.url_for_boot(:initrd) %>"
+grubby --add-kernel=$KERNEL --initrd=$INITRD --copy-default \
+  --title "Grubby reprovision <%= @host.operatingsystem %>" \
+  --make-default \
+  --args="ks=<%= foreman_url('provision')%>"
+shutdown -r +1
+<%
+  else
+-%>
+echo Host not in build mode
+false
+<%
+  end
+-%>


tbrisker:
closing for now, thanks

lzap:
Actually safemode change was merged, do we want to merge this? I tested this. @ares or @ezr-ondrej  ?

job_templates/run_command_-_grubby_reprovision.erb
```diff
@@ -0,0 +1,27 @@
+<%#
+kind: job_template
+name: Run Command - Grubby Reprovision
+model: JobTemplate
+job_category: Power
```
ares: this category feels odd, I'd either create a new one - provisioning or use commands

job_templates/run_command_-_grubby_reprovision.erb
```diff
@@ -0,0 +1,27 @@
...
+provider_type: SSH
+%>
+<%
+  if @host.build?
+-%>
+KERNEL="/boot/grubby-reporovision-kernel"
+INITRD="/boot/grubby-reporovision-initrd.img"
+wget -O "$KERNEL" "<%= @host.url_for_boot(:kernel) %>"
```
ares, line 14: if this fails, we still continue, even the REX job looks like success, we need to check for success of this and the next command

job_templates/run_command_-_grubby_reprovision.erb
```diff
@@ -0,0 +1,27 @@
...
+wget -O "$INITRD" "<%= @host.url_for_boot(:initrd) %>"
+grubby --add-kernel=$KERNEL --initrd=$INITRD --copy-default \
```
ares, line 16: is this always present or do we need to ensure we install some package?

job_templates/run_command_-_grubby_reprovision.erb
```diff
@@ -0,0 +1,27 @@
...
+  --title "Grubby reprovision <%= @host.operatingsystem %>" \
+  --make-default \
+  --args="ks=<%= foreman_url('provision')%>"
+shutdown -r +1
```
ares, line 20: we should not reboot unless all above succeeded

ares (changes requested):
when the script succeeds it terminates the SSH so the job fails with error

```
Unexpected disconnect: IOError - closed stream
  63:
Exit status: EXCEPTION
```

AFAIK there's a solution for this, job template can somehow say it expects the reboot or even continue after the reboot, @adamruzicka would know how to do it

besides the comments I left, this is absolutely great addition!

job_templates/run_command_-_grubby_reprovision.erb
lzap: This is rare workflow, I will stick with Commands then.

job_templates/run_command_-_grubby_reprovision.erb
lzap: Adding `set -e` which will cause to stop. I am assuming bash is used.

job_templates/run_command_-_grubby_reprovision.erb
lzap: It can be missing, but do you really want me to put "if os is redhat then yum install" stuff? I mean, it's easier to just error out and user will immediately see this from the result and reschedule this.

lzap:
Rebased.

job_templates/run_command_-_grubby_reprovision.erb
ares: there's a snippet for installing a package ;-) Just use this

```
<%= render_template('Package Action - SSH Default', :action => 'install', :package => 'grubby') %>
```

as a line above this. This will install `grubby` package using yum, apt and zypper

job_templates/run_command_-_grubby_reprovision.erb
ares: I assume this should be resolved by `set -e` (if that works)

job_templates/run_command_-_grubby_reprovision.erb
ares: I'll retest that

tbrisker:

PR #616, opened 2019-09-10
Use custom port for discovery ipxe

lzap:
Labels were not correct and it was probably confusing to provide three options.
This is more consitent - default HTTP port is 8000 and user can customize it if
proxy was deployed with a different one (there is an extra option for that).


canihavethisone:
You have defaulted the port to 8448 - did you mean this to be 8000 ?  I would think that the more common esp for httpboot

lzap:
Default is "local", then both discovery items are explicit. But you are right that key "d" (stands for discovery) was set to both. I am only keeping 8000 now. Rebased, thanks!

canihavethisone:
Yep, I meant that you have 'set port 8448' while I think this should be 8000?
Great work lzap, I've read your guides on Foreman ipxe / efi discovery and found them very valuable.

lzap:
Thanks, nice avatar by the way! Cheers.

@mmoll or @timogoebel how you feel about merging this guy? ;)

timogoebel:
Thanks, @lzap.

PR #618, opened 2019-09-10
Change repo name for puppet 6

### LINKED ISSUE #617: Wrong repo name for puppet v6

jacac:
Fixes GH-617


ekohl:
This looks correct, but this would be a very good one to get cherry picked into 1.23.1. Could you open a Redmine issue so we can track this?

PR #620, opened 2019-09-17
Create bios_uefi_partition_tables_template.erb

pjayKHV:


lzap:
Thank you, I've took your example and extended it with more parameters so other users can take advantage of that. I changed defaults: XFS is the recommended FS by Red Hat and I don't create home by default: https://github.com/theforeman/community-templates/pull/621

PR #621, opened 2019-09-17
Added kickstart custom

lzap:
We don't ship a good kickstart example how to partition EFI and BIOS systems. This supports both. Untested.

### DIFF
--- a/partition_tables_templates/kickstart_custom.erb
+++ b/partition_tables_templates/kickstart_custom.erb
@@ -1,0 +1,56 @@
+<%#
+kind: ptable
+name: Kickstart custom
+model: Ptable
+oses:
+- CentOS
+- Fedora
+- RedHat
+%>
+zerombr
+clearpart --all --initlabel
+<%
+  dev = host_param('part_device') || 'sda'
+  swap = host_param('part_swap_size') || '4096'
+  root = host_param('part_root_percent') || '100'
+  home = host_param('part_home_percent') || '0'
+  fstype = host_param('part_fstype') || 'xfs'
+-%>
+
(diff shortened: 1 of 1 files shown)


pjayKHV:
I think its best to also change 

> part /boot **--fstype="ext4"** --ondisk=<%= dev %> --size=1024

 to 

> part /boot **--fstype="<%= fstype %>"** --ondisk=<%= dev %> --size=1024

 so it would also default to xfs instead of ext4.

pjayKHV (changes requested):
I think its best to also change

part /boot --fstype="ext4" --ondisk=<%= dev %> --size=1024

to

part /boot --fstype="<%= fstype %>" --ondisk=<%= dev %> --size=1024

so it would also default to xfs instead of ext4.

lzap:
Oh yeah, I haven't noticed that on RHEL7 onwards /boot is xfs, but older RHELs still expect this to be ext2-compatible and won't boot. Let me change it and do this only when this is RHEL7+.

lzap:
Updated, if you can test this version that would be *very* appreciated. I haven't tested myself.

mmoll:
@pjayKHV could you have a look? :eyes:

pjayKHV:
> @pjayKHV could you have a look?

I will test it, as soon as i get foreman up and running... Had to kill all of the deployments that i had at the moment... Sorry...

pjayKHV:
Tested the template with Foreman 1.23.1 and CentOS 7.7... Everything seems to be good... Sorry for the delay...

ekohl (approved):

ekohl:
Thanks!

PR #623, opened 2019-09-17
Fix - Added wget as additional package in preseed template

sbschuermann:
As of debian 10 wget it is no longer installed by default but needed e.g. in the preseed last_command.

Although this could be mitigated my adding it to "additional_packages" parameter i consider this to be bug.


ekohl:
I just merged https://github.com/theforeman/community-templates/pull/594 which should have been merged some time ago.

PR #626, opened 2019-09-18
Do not rush opening iPXE menu

lzap:
When there is some error, the iPXE menu quickly clears the screen. This delays it a bit so you can investigate what was wrong before the default menu pops in.


mmoll:
merged, díky @lzap!

PR #628, opened 2019-09-23
modifications for proxy (squid-orcharhino-proxy)

mdellweg:


provisioning_templates/PXELinux/kickstart_default_pxelinux.erb
```diff
@@ -12,6 +12,8 @@ oses:
 <%
   major = @host.operatingsystem.major.to_i
   mac = @host.provision_interface.mac
+  proxy_uri = host_param('http-proxy') ? "http://#{host_param('http-proxy')}:#{host_param('http-proxy-port')}" : nil
```
ekohl, line 15: What if `host_param('http-proxy-port')` is nil/empty?

provisioning_templates/snippet/built.erb
```diff
@@ -15,18 +15,22 @@ snippet: true
...
-    curl_opts << "--insecure"
-    wget_opts << "--no-check-certificate"
+    curl_opts << '--insecure'
+    wget_opts << '--no-check-certificate'
   end
   if @body_file
     curl_opts << "--data @#{@body_file}"
     wget_opts << "--body-file=#{@body_file}"
   end
+  if not host_param('http-proxy')
```
ekohl:
```suggestion
  unless host_param('http-proxy')
```

ekohl (commented):
I have my concerns about the `built` change. AFAIK it's intended to never intended that you proxy it. If you need to there's the `templates` plugin on the smart-proxy that should do that for you. It's supposed to automatically generate a URL that's pointing to the smart-proxy. @lzap probably knows.

provisioning_templates/PXELinux/kickstart_default_pxelinux.erb
mdellweg: I see, what you mean...

mdellweg:
@ekohl we believe, your concern is valid. Using a proxy with the template feature should give the proxy url. We will have another thought about it. Closing this PR meanwhile.

PR #633, opened 2019-10-04
fix realm-join when using provisioning w/cloud-init

mmoll:
closes GH-503


mmoll:
@ekohl could you have a look?

ekohl (approved):
Haven't verified this, but I trust the reporter.

PR #635, opened 2019-10-08
Fixes #28020 - Ansible Inventory report template

ofedoren:


report_templates/ansible_inventory.erb
```diff
@@ -0,0 +1,137 @@
...
+      hostgroup = host.hostgroup.title if host.hostgroup
+      inventory_data.update('Host Group': hostgroup) if input('Host Group') == 'yes'
+      host_collections = host.host_collections.map { |c| c.name } if host.host_collections
+      inventory_data.update('Host Collections': host_collections) if input('Host Collections') == 'yes'
+      inventory_data.update('IPv4': host.ip) if input('IPv4') == 'yes'
+      inventory_data.update('Subnet': host.subnet) if input('Subnet') == 'yes'
+      smart_proxies = host.smart_proxies.map { |p| p.name } if host.smart_proxies
+      inventory_data.update('Smart Proxies': smart_proxies) if input('Smart Proxies') == 'yes'
+      inventory_data.update('Content Attributes': content_attribute_data) if content_attrs
+      inventory_data.update('Facts': host.facts) if input('Facts') == 'yes'
```
sjha4: We would ideally want these keys in lowercase.

report_templates/ansible_inventory.erb
```diff
@@ -0,0 +1,137 @@
...
+-%>
+<%-   end -%>
+<%-   inventory_data = { 'Name': host.name }
+      inventory_data.update('Organization': host.organization) if input('Organization') == 'yes'
+      inventory_data.update('Location': host.location) if input('Location') == 'yes'
+      hostgroup = host.hostgroup.title if host.hostgroup
+      inventory_data.update('Host Group': hostgroup) if input('Host Group') == 'yes'
+      host_collections = host.host_collections.map { |c| c.name } if host.host_collections
+      inventory_data.update('Host Collections': host_collections) if input('Host Collections') == 'yes'
+      inventory_data.update('IPv4': host.ip) if input('IPv4') == 'yes'
```
ekohl: Could you add the IPv6 equivalent as well?

report_templates/ansible_inventory.erb
ofedoren: Done.

report_templates/ansible_inventory.erb
ofedoren: Sure, done.

ofedoren:
I've updated the template to do less db queries, but there is still N+1 problem for `smart_proxies` and `facts` retrieving.

I think is is because of how those methods are implemented for `Host` model:
 - `smart_proxies`: https://github.com/theforeman/foreman/blob/develop/app/models/concerns/hostext/smart_proxy.rb#L5-L7
 - `facts`: https://github.com/theforeman/foreman/blob/develop/app/models/host/base.rb#L254-L261

Maybe there is another way to retrieve such information in case if we want to eliminate N+1 for those queries?

sjha4:
Is there a way to filter this report?

report_templates/ansible_inventory.erb
```diff
@@ -0,0 +1,144 @@
...
+  default: no
+model: ReportTemplate
+require:
+- plugin: foreman_ansible
+  version: '>= 3.1.0'
+- plugin: katello
+￼  version: '>= 3.9.0'
+-%>
+<%- content_attrs = input('Content Attributes') == 'yes' %>
+<%- load_hosts(includes: [:organization, :location, :hostgroup, :host_collections, { interfaces: :subnet }, :content_facet]).each_record do |host| -%>
```
ares, line 94: it would be good to add hosts filter, similar to https://github.com/theforeman/community-templates/blob/develop/report_templates/host_statuses.erb#L15 (see search: `input('hosts')`, that should address @sjha4 comment about filtering

ares:
Thanks @xprazak2, merging!

ares:
I'm sorry I click the button on wrong tab, reverting PR open, please reopen. Alternatively, we can solve n+1 and host filtering separately, the template in general works

report_templates/ansible_inventory.erb
```diff
@@ -0,0 +1,144 @@
...
+  input_type: user
+  description: Whether should be Facts included.
+  advanced: false
+  value_type: plain
+  options: "yes\r\nno"
+  default: no
+model: ReportTemplate
+require:
+- plugin: foreman_ansible
+  version: '>= 3.1.0'
```
ezr-ondrej, line 89: :warning: This is wrong, it can not be seeded to the foreman. Or am I missing something?

ofedoren:
@ezr-ondrej, I've rebased this PR before merge so this is should be fixed, see https://github.com/theforeman/community-templates/blob/develop/report_templates/ansible_inventory.erb#L113

Although I don't quite understand why the changes in this commit are shown as the old ones...

report_templates/ansible_inventory.erb
ares: Fixed already

ofedoren: Oh, right. This is the old PR.

This was fixed in https://github.com/theforeman/community-templates/commit/34f4357d566f313eccdf8af1b46ef5598bb7e265

PR #606, opened 2019-08-08
Simplify grub2 custom file loading

lzap:
This needs https://github.com/theforeman/smart-proxy/pull/670

See the associated RM ticket for more info

PR #636, opened 2019-10-11
Fixes simplified grub2 MAC loading

lzap:
The simplified MAC loading works great:

https://github.com/theforeman/community-templates/pull/606

However, it does not when using grub2 in HTTP UEFI Boot. In that context, grub2 messes URL in memory and then fails to load any file. I have found out that when absolute paths are first, it works fine and only relative paths causes the bug to appear.

Also, instead of "source" we can now use "configfile" statement which prevents from "chainload" items below which can interfere with the included item.


mmoll:
merged, díky @lzap!

PR #637, opened 2019-10-11
Fixes #28023 - fix rancheros PXE template

uvNikita:
Add kernel parameters with APPEND directive, set initrd image with INITRD directive.


mmoll:
merged, thanks @uvNikita!

PR #638, opened 2019-10-11
Added support for grub2 over HTTP Boot

lzap:
I am testing HTTP UEFI Boot via Grub2 and it looks like the path must be absolute not relative including the prefix `/httpboot` which is required. This is the path where proxy httpboot module exports the TFTP directory over HTTP or HTTPS.


mmoll:
merged, díky @lzap!

PR #639, opened 2019-10-11
Revert "Fixes #28020 - Ansible Inventory report template"

ares:
Reverts theforeman/community-templates#635


ares:
I accidentaly merge the PR too early, reverting

ares:
alternatively the two missing improvements can be addressed via separate PRs, the template works

report_templates/ansible_inventory.erb
```diff
@@ -1,144 +0,0 @@
...
-<%-   inventory_data = { 'name': host.name }
-      inventory_data.update('organization': host.organization) if input('Organization') == 'yes'
-      inventory_data.update('location': host.location) if input('Location') == 'yes'
-      hostgroup = host.hostgroup.title if host.hostgroup
-      inventory_data.update('host_group': hostgroup) if input('Host Group') == 'yes'
-      host_collections = host.host_collections.map { |c| c.name } if host.host_collections
-      inventory_data.update('host_collections': host_collections) if input('Host Collections') == 'yes'
-      inventory_data.update('ipv4': host.ip) if input('IPv4') == 'yes'
-      inventory_data.update('ipv6': host.ip6) if input('IPv6') == 'yes'
-      inventory_data.update('subnet': host.subnet) if input('Subnet') == 'yes'
```
ekohl, line 136: The IPv6 subnet is also missing. If this PR changes into fixing it, please have a look at that as well.

report_templates/ansible_inventory.erb
```diff
@@ -1,144 +0,0 @@
...
-  description: Whether should be Facts included.
-  advanced: false
-  value_type: plain
-  options: "yes\r\nno"
-  default: no
-model: ReportTemplate
-require:
-- plugin: foreman_ansible
-  version: '>= 3.1.0'
-- plugin: katello
```
ekohl, line 90: Can it be made so that Katello is optional and it detects whether Katello is installed?

ares: There's no good mechanism for that, we have macro to check if plugin is present but input "include content attributes" would still be available and silently ignored. Given the inventory template is configurable through settings, I'd start with this and people could easily create alternative, dropping some attributes.

ekohl: That's fair. In a way this is becoming an API so it'd need a way to raise an error when an invalid combination was requested.

ares:
ok, closing as fixes the rest https://github.com/theforeman/community-templates/pull/640

PR #641, opened 2019-10-11
Fixes #27180 - VLAN tagged network

lzap:
Adds VLAN support, also slightly improve iPXE template by squashing all ERB individual lines into one block. Also introduced few helper variables to shorten up OS/version detection statements.

I tested Grub2 + kickstart VLAN provisioning end to end. The iPXE should work as well. Going to test PXELinux too.


lzap:
Added tests for PXELinux. Long-term we want to replace these with Foreman renderer after we extract it into a gem.

lzap:
Fixed tests caused by the rebase.

mmoll:
merged, thanks @vgunasek a díky @lzap!

PR #642, opened 2019-10-16
Refs #2802 - fixes the whitespaces

ares:


ekohl (approved):
I'm wondering if we could have some automated script that validates the metadata so CI would capture this.

PR #643, opened 2019-10-16
Fix use-ntp host param not working in el7

nndegz:
Fixes #28077 - Fix use-ntp


ekohl (changes requested):
Can you change the commit message to say `Fixes #28077 - Fix use-ntp ....` so our automation can pick it up?

lzap (changes requested):
This is a regression, I don't believe it is worth confusing users who will upgrade to Foreman next with this. With RHEL8 / CentOS8 chrony will be the default, Foreman-installed boxes of RHEL7 and clones were using ntpd and it should stay like that.

nndegz:
closing per request

PR #644, opened 2019-10-18
Update to Grub2 HTTP UEFI global template

lzap:
Continuing testing UEFI HTTP Boot now with Discovery. It also needs to be
booted with the given prefix, so this change creates four different global
entries and user must pick the correct default value in order to boot into
discovery.


lzap:
Any chance to get this into 1.24? Would love to have UEFI HTTP boot end-to-end support there. Thanks. @mmoll or @ares or @tbrisker

mmoll:
I didn't test this, but trusting @lzap he did.

mmoll:
merged, díky @lzap!

PR #652, opened 2019-10-30
Fixes #28157 - Update Ansible Inventory to have host id/params

ofedoren:
Requires https://github.com/theforeman/foreman/pull/7143.

Also, host IDs are always included as well as host names. Please tell me whether this is desired or not.

### DIFF
--- a/report_templates/ansible_inventory.erb
+++ b/report_templates/ansible_inventory.erb
@@ -35,6 +35,14 @@ template_inputs:
   value_type: plain
   options: "yes\r\nno"
   default: "yes"
+- name: Host Parameters
+  required: true
+  input_type: user
+  description: Whether should be Host Parameters included.
+  advanced: false
+  value_type: plain
+  options: "yes\r\nno"
+  default: "no"
 - name: IPv4
   required: true
   input_type: user
@@ -107,7 +115,11 @@ require:
   version: '3.9.0'
 -%>
 <%- content_attrs = input('Content Attributes') == 'yes' %>
-<%- load_hosts(search: input('Hosts'), includes: [:organization, :location, :hostgroup, :host_collections, { interfaces: :subnet }, :content_facet]).each_record do |host| -%>
+<%- interface_includes = [:subnet] %>
+<%- interface_includes.push(:domain) if input('Host Parameters') == 'yes' %>
+<%- includes = [:organization, :location, :hostgroup, :host_collections, { interfaces: interface_includes }, :content_facet] %>
+<%- includes.push(:operatingsystem) if input('Host Parameters') == 'yes' %>
+<%- load_hosts(search: input('Hosts'), includes: includes).each_record do |host| -%>
 <%-   content_facet = host_content_facet(host) if content_attrs -%>
 <%-   content_attribute_data = nil -%>
 <%-   if content_facet -%>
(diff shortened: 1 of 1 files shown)


sjha4:
Tested and works great. 

One question: I see these 3 additional params, (which I am assuming are global params?) with the report that I don't see in the old API.
remote_execution_connect_by_ip: false
remote_execution_effective_user_method: sudo
remote_execution_ssh_user: root

ares:
these seems to be auto added by REX, there's not much we can do about it

report_templates/ansible_inventory.erb
```diff
@@ -107,7 +115,9 @@ require:
   version: '>= 3.9.0'
 -%>
 <%- content_attrs = input('Content Attributes') == 'yes' %>
-<%- load_hosts(search: input('Hosts'), includes: [:organization, :location, :hostgroup, :host_collections, { interfaces: :subnet }, :content_facet]).each_record do |host| -%>
+<%- includes = [:organization, :location, :hostgroup, :host_collections, { interfaces: :subnet }, :content_facet] %>
+<%- includes = [:organization, :location, :hostgroup, :host_collections, { interfaces: [:subnet, :domain] }, :content_facet, :operatingsystem] if input('Host Parameters') == 'yes' %>
```
ares: I'd prefer to either don't repeat the list

```ruby
interface_includes = [ :subnet ]
interface_includes.push(:domain) if input('Host Parameters') == 'yes'
includes = includes = [:organization, :location, :hostgroup, :host_collections, { interfaces: interface_includes }, :content_facet]
includes.push(:operatingsystem) if input('Host Parameters') == 'yes'
```

or don't define the variable twice
```ruby
if input('Host Parameters') == 'yes'
  includes = [:organization, :location, :hostgroup, :host_collections, { interfaces: [:subnet, :domain] }, :content_facet, :operatingsystem]
else
  includes = [:organization, :location, :hostgroup, :host_collections, { interfaces: :subnet }, :content_facet]
end  
```

the later may be more readable, but the first one more extendable if we want to have more conditions for other inputs in future

ares (changes requested):
one small change requested, otherwise good to be merged, I guess this needs to go to 1.24 to keep tower inventory compatible, I'll open a cherry-pick when it's merged

report_templates/ansible_inventory.erb
```diff
@@ -35,6 +35,14 @@ template_inputs:
...
   options: "yes\r\nno"
   default: yes
+- name: Host Parameters
+  required: true
+  input_type: user
+  description: Whether should be Host Parameters included.
+  advanced: false
+  value_type: plain
+  options: "yes\r\nno"
+  default: no
```
ares: as it turned out, this needs to be wrapped in quotes `"no"`, otherwise YAML parses it as false, which is not one of "yes", "no" options.

ofedoren: Thanks, done.

report_templates/ansible_inventory.erb
ofedoren: I've chosen the first way, thanks :)

ares (approved):
Thanks @ofedoren, merging! I'll also open a CP for 1.24

PR #653, opened 2019-10-30
Update kickstart_default.erb

jkalchik:
Add a "custom repositories" hook into the Kickstart default provisioning template to allow local customization of repositories available during the anaconda initial package install phase.  This is my suggested location, it may be appropriate to move the snippet call up or down.


jkalchik:
@lzap :  new snippet hook we've talked about.

lzap:
Thanks.

PR #656, opened 2019-11-01
Fixes #28156 - properly specify requirements

ares:
Only version number can be specified, they always mean minimum version.
Also default value of yes or no needs to be wrapped in quotes so it's
considered a String by YAML. Otherwise it auto converts it to true/false
which then fails validation of the default being one of defined options
(yes/no).

(cherry picked from commit 8b1c8a4be7807d0e39498c64f6d8f422f7a5edf9)

### DIFF
--- a/report_templates/ansible_inventory.erb
+++ b/report_templates/ansible_inventory.erb
@@ -10,7 +10,7 @@ template_inputs:
   advanced: false
   value_type: plain
   options: "yes\r\nno"
-  default: yes
+  default: "yes"
 - name: Location
   required: true
   input_type: user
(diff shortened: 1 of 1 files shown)


ares:
I pushed second commit that also needs to be cheery-picked to 1.24 so I don't have to rebase it afterwards. @tbrisker the first commit solves the seeding issues, the second is required for this template to be backwards compatible replacement for the old ansible inventory script. This was well tested (through seed, sync from AWX) by me and QE. It needs also https://github.com/theforeman/foreman/pull/7147 in core 1.24, which is extrasupersafe.

PR #659, opened 2019-11-12
Remove ignoreschedules from puppet.conf

damonmaria:
As per [this discussion](https://community.theforeman.org/t/question-why-ignoreschedules-in-puppet-conf/15362/4) `ignoreschedules = true` seems to be a leftover from some forgotten time. Schedules can not be set through Foreman as it is an ENC. However, schedules in included classes most likely don't want to be ignored by default. `ignoreschedules` seems to be a debugging option, not something that should be permanently on.


mmoll:
merged, thanks @damonmaria!

PR #660, opened 2019-11-18
fixes# 28168 - added registered users template

domitea:


report_templates/registered_users.erb
```diff
@@ -0,0 +1,26 @@
+<%#
+name: List of registered users
```
ares: I'd call it just `Registered users` to remain consistent with other reports. Reports are lists usually, so we don't need to repeat that information

ares (changes requested):
Looks good to me, needs https://github.com/theforeman/foreman/pull/7155 to be in first in order to test. I left just one nit pick.

report_templates/registered_users.erb
```diff
@@ -0,0 +1,26 @@
...
+  input_type: user
+  description: Limit the report only on users found by this search query. Keep empty
+    for report on all available users.
+  advanced: false
+  value_type: search
+  resource_type: User
+model: ReportTemplate
+-%>
+<%- load_users(search: input('Users filter'), includes: :auth_source).each_record do |user| -%>
+  <%-  report_row({
```
ares: the indentation here is incorrect, it should start on col 0 and the indentation should be done inside of the ERB, see other reports for an example

report_templates/registered_users.erb
```diff
@@ -0,0 +1,26 @@
...
+  description: Limit the report only on users found by this search query. Keep empty
+    for report on all available users.
+  advanced: false
+  value_type: search
+  resource_type: User
+model: ReportTemplate
+-%>
+<%- load_users(search: input('Users filter'), includes: :auth_source).each_record do |user| -%>
+<%-   report_row({
+         'Login': user.login,
```
ares: sorry for nitpicking, but the indentation should be by 2 spaces, this is 3 spaces comparing to the line above

domitea:
Is it right now? @ares

ares (approved):
Good, tested this with the linked PR and works great! Ready to merge when core PR gets in

mmoll:
merged, díky @domitea!

PR #661, opened 2019-11-19
Make ntp/chrony an exclusive package

lzap:
Looks like chrony is in @Core group now, so we should either install one or another depending on the setting.


tbrisker:
does this need to be cherry-picked to 1.24?

tbrisker (approved):

lzap:
I don't think so, not big deal. Thanks!

PR #662, opened 2019-11-20
Remove fedora-everything repo definition

lzap:
I don't know why this is there but user reports that it is causing troubles when installing Fedora from local mirrors. This does not need to be there, Anaconda should set them up automatically.


mmoll:
merged, díky @lzap!

PR #668, opened 2019-11-29
Add Enable Web Console ansible play

lzap:
SSIA


job_templates/service_action_-_enable_web_console.erb
```diff
@@ -0,0 +1,35 @@
...
+  - name: ensure cockpit is installed
+    package:
+      name: "cockpit"
+      state: present
+
+  - name: enable firewall rule for cockpit
+    block:
+      - name: install firewalld bindings
+        package:
+          name: python-firewall
```
ekohl: Various distros have started to use `python2-` or `python3-` prefixes. Does this work on EL8 for example?

job_templates/service_action_-_enable_web_console.erb
```diff
@@ -0,0 +1,35 @@
...
+          permanent: true
+          immediate: true
+          state: enabled
+
+  - name: ensure cockpit is started
+    systemd:
+      name: "cockpit.socket"
+      state: "started"
+      enabled: true
+      daemon_reload: true
```
ekohl: Why is a daemon reload needed? Usually the package takes care of that.

job_templates/service_action_-_enable_web_console.erb
```diff
@@ -0,0 +1,35 @@
...
+%>
+---
+- hosts: all
+  tasks:
+  - name: ensure cockpit is installed
+    package:
+      name: "cockpit"
+      state: present
+
+  - name: enable firewall rule for cockpit
```
ekohl: Would it make sense to add a parameter to do this? Not everyone is using firewalld.

job_templates/service_action_-_enable_web_console.erb
```diff
@@ -0,0 +1,35 @@
+<%#
+name: Service Action - Enable Web Console
+job_category: Ansible Services
+snippet: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
```
adamruzicka, line 7: Could you add `feature: ansible_enable_web_console` somewhere around here?

lzap:
Let's put this one on hold, I haven been informed in the YT video comment (https://www.youtube.com/watch?v=Ua3b79UG9JM) that Cockpit does NOT need to be enabled on the target host actually. Let's confirm that.

adamruzicka:
It may not need to be enable but as far as I know it needs to be at least installed there. Shouldn't we do at least that?

lzap:
That's a good point, I can confirm that `cockpit` package needs to be enabled for the integration. The service or port is not actually needed.

job_templates/service_action_-_enable_web_console.erb
lzap: Done.

lzap:
With that in mind, the Ansible is now much more simple :-)

ekohl:
Consider my review dismissed/addressed. I'll leave it up to @adamruzicka to ack.

lzap:
I've tested that cockpit-system is the only dependency needed. It also appears to be present on Debian systems. Labeller test do test tho.

ekohl:
@adamruzicka mind having another look?

adamruzicka (approved):
LGTM

ekohl:
Thanks!

ares:
@lzap mind to open a request to document this? It's a small thing but really useful.

PR #670, opened 2019-12-07
comparing packages between content hosts

waldirio:
Hello @ares 

This is the first of many reports that will bump Sat report feature! :-)

I didn't create any redmine to this, please let me know if you believe to be necessary. If not, I'll just push the new reports via github only.

Thank you.

### DIFF
--- a/report_templates/host_compare_content_hosts_packages.erb
+++ b/report_templates/host_compare_content_hosts_packages.erb
@@ -1,0 +1,70 @@
+<%#
+name: Host - compare content hosts packages
+snippet: false
+template_inputs:
+- name: Host 1
+  required: true
+  input_type: user
+  advanced: false
+  value_type: plain
+  resource_type: Katello::ActivationKey
+- name: Host 2
+  required: true
+  input_type: user
+  advanced: false
+  value_type: plain
+  resource_type: Katello::ActivationKey
+model: ReportTemplate
+-%>
+<%-  general_lst = [] -%>
+<%-  stage_lst = [] -%>
+<%-  final_lst = [] -%>
+<%-  arch_lst = [] -%>
+<%-  host1 = Host.find_by_name(input("Host 1")) -%>
+<%-  host2 = Host.find_by_name(input("Host 2")) -%>
+<%-  set1 = host1.installed_packages.group_by { |p| p.name } -%>
+<%-  set2 = host2.installed_packages.group_by { |p| p.name } -%>
+<%-  (set1.keys | set2.keys).uniq.each do |package_name| -%>
+<%-    data = { 'Package' => package_name } -%>
+<%-    data[host1.fqdn] = set1[package_name].try(:first).try(:nvra) || '-' -%>
+<%-    data[host2.fqdn] = set2[package_name].try(:first).try(:nvra) || '-' -%>
+<%-    general_lst << data -%>
+<%-  end -%>
+<%-  general_lst.each do |d| -%>
+<%-    ch1_ver = d[host1.fqdn].gsub(d["Package"]+"-","").gsub(/\.el[0-9][^.]*/,"").gsub(/\.[n|x|i].*/,"") -%>
+<%-    ch2_ver = d[host2.fqdn].gsub(d["Package"]+"-","").gsub(/\.el[0-9][^.]*/,"").gsub(/\.[n|x|i].*/,"") -%>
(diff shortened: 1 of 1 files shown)


ares:
Works nicely without safe mode. I've sent you a patch the reformats the code a bit and modifies the code to work with safe mode. That also needs following PR in core https://github.com/theforeman/foreman/pull/7325

I'll see if I could simplify the code a bit more and extract some common patterns to katello code base, but that wouldn't block merging this.

report_templates/host_compare_content_hosts_packages.erb
```diff
@@ -0,0 +1,70 @@
...
+<%-  stage_lst = [] -%>
+<%-  final_lst = [] -%>
+<%-  arch_lst = [] -%>
+<%-  host1 = Host.find_by_name(input("Host 1")) -%>
+<%-  host2 = Host.find_by_name(input("Host 2")) -%>
+<%-  set1 = host1.installed_packages.group_by { |p| p.name } -%>
+<%-  set2 = host2.installed_packages.group_by { |p| p.name } -%>
+<%-  (set1.keys | set2.keys).uniq.each do |package_name| -%>
+<%-    data = { 'Package' => package_name } -%>
+<%-    data[host1.fqdn] = set1[package_name].try(:first).try(:nvra) || '-' -%>
```
ekohl: Bit of an edge case, but the kernel usually has multiple versions installed. Should it attempt to handle this or ignore that edge case?

ares: good catch, given we can't tell which kernel is running, I'm not sure it's worth trying to list all versions here, we could potentially order them to get the newest at least, but it can be more misleading

I always thought only one package can actually be installed but kernels remain installed until explicitly purged, but I'm probably mixing this with debian world, where packages contain kernel version in their name and "kernel" package just enforce some version to be installed

ares:
Here's link to the PR with more fixes https://github.com/waldirio/community-templates/pull/1, I got quite involved in the code, but also tested quite heavily. If it's merged, someone else should take a look at the code and merge. More things could be deduplicated and extracted to helpers, e.g. parsing if architecture and version from nvra, but that would only put this useful report on hold for another version.

report_templates/host_compare_content_hosts_packages.erb
```diff
@@ -0,0 +1,70 @@
...
+<%-    ch2_ver = d[host2.fqdn].gsub(d["Package"]+"-","").gsub(/\.el[0-9][^.]*/,"").gsub(/\.[n|x|i].*/,"") -%>
+<%-    arch_lst << d[host1.fqdn].split(/\./).last -%>
+<%-    arch_lst << d[host2.fqdn].split(/\./).last -%>
+<%-    arch_lst = arch_lst.uniq -%>
+<%-    arch_lst.delete("-") -%>
+<%-    if ch1_ver == "-" -%>
+<%-      status = "#{host2.fqdn} only"-%>
+<%-    elsif ch2_ver == "-" -%>
+<%-      status = "#{host1.fqdn} only"-%>
+<%-    elsif ch1_ver > ch2_ver -%>
```
ekohl: Does this compare versions properly? For example, `3.10 > 3.2` in versions but in Ruby that won't be true. While within RHEL it may remain the same, you will hit it with releases as well.

ares: Yeah, we may need to wrap it with Gem version comparison, probably deserves a macro

waldirio:
Hey @ares, merged and now updated here. Sorry to be using `develop` branch .., just realized that.

Thank you again and let me know if you need something else.

report_templates/host_compare_content_hosts_packages.erb
ares: this is still an open item, I'll see if I can get to it soon

ares:
This now needs https://github.com/theforeman/foreman/pull/7361/files and therefore it will only be compatible with Foreman 2.0 (if it gets merged on time). @waldirio please feel free to squash all commits from this PR. Note that I added one on top of your last version.

report_templates/host_compare_content_hosts_packages.erb
ares: addressed  in the last commit in this PR

waldirio:
@ares I'm closing this one and will submit a new PR in the correct way.

Thank you!!

PR #674, opened 2019-12-25
Adapt script to work with WIMan

UXabre:
WIMan offers additional optimizations for creating WIM images. To have the best result (and the least amount of configuration) and additonal manifest "images.ini" should be deployed on the server.

However, as not everybody will notice this and chances that somebody has embedded WIMaging in their workflow are real it will fallback to downloading the regular install.wim in case the images.ini file hasn't been found


UXabre:
Friendly reminder for a review :-) Having this in the templates will drastically improve ease of use for windows installation

ekohl:
Does this depend on anything in core or can we merge this now?

UXabre:
Hey! No core dependency; should be all good :-) thanks for checking this out!

ekohl (approved):
I have very little knowledge of Windows provisioning, but it looks correct and I trust you've done your research/testing. Given how little Windows provisioning reviewers we have I'm going to merge this now and if there's comments, they can be sent as new PRs.

PR #675, opened 2020-01-04
Fixes #28637 - set minimal headers for empty reports

ares:
When the report does not render any row, resulting file does not have
any headers. This can be confusing for users, so render_headers macro
was added. It should specify column names that user should always see.
In case some columns are added dynamically, and report_row adds some
additional columns, they are added on demand. Meaning, if we iterate
over some data, new columns can be still added.

### DIFF
--- a/report_templates/ansible_inventory.erb
+++ b/report_templates/ansible_inventory.erb
@@ -114,6 +114,7 @@ require:
 - plugin: katello
   version: '3.9.0'
 -%>
+<%- report_headers 'id', 'name' -%>
 <%- content_attrs = input('Content Attributes') == 'yes' %>
 <%- interface_includes = [:subnet] %>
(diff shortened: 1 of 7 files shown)


ares:
Could someone take a look please? It may easily get out of date. It's pretty much safe change I believe.

ares:
Anyone? This is ultra safe and IMHO very useful.

ekohl (approved):
How does it deal with a mismatch between the headers and row reporting? Does it add a column? Silently drop it?

ares:
good question and probably poor explanation in the first comment :-)
e.g.
```
report_headers :a, :b: 
report_row a: 1, b: 2, c: 3
```

will result in all a,b,c columns, however if report_row is not called e.g. no object is found, the empty report will only contain a,b columns, therefore we should define report_headers only for columns that are always present

ekohl:
That looks like how I'd expect it

PR #677, opened 2020-01-13
Add Windows GPT EFI partition table

canihavethisone:
This partition table creates GPT EFI partition and the Windows Recovery Environment (WinRE) partition, so that a new install places the recovery media there instead of on the primary partition.

Correct IDs are applied to mark partitions thus.


lzap:
Hello, thanks and sorry for the delay. I haven'ŧ tested this but since we have similar template and this is an alternative, I am merging. Syntax-wise looks good AFAIR Windows batch language :-)

PR #678, opened 2020-01-15
Refs #28768 - update templates to start with metadata

ares:


ares:
The core PR has been merged, this one should get in too to keep consitency between what we seed and what is exported, cc @tbrisker

ekohl (approved):

ares:
thank you :-)

PR #679, opened 2020-01-17
Fixes #28782 - complete compare ch report template

waldirio:
The complete code with all changes made before. At this time, using the correct branch also issue on redmine.

### DIFF
--- a/report_templates/host_compare_content_hosts_packages.erb
+++ b/report_templates/host_compare_content_hosts_packages.erb
@@ -1,0 +1,55 @@
+<%#
+name: Host - compare content hosts packages
+snippet: false
+template_inputs:
+- name: Host 1
+  required: true
+  input_type: user
+  advanced: false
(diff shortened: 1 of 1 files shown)


lzap (approved):
Thanks.

ares:
@lzap, this was I believe not ready, see the comment in https://github.com/theforeman/community-templates/pull/670#issuecomment-574812971, I think @waldirio opened this but forgot to link the old PR.

So we need to get https://github.com/theforeman/foreman/pull/7361/files in, the `version` macro does not exist yet in core. Also this template may change based on the comments in this PR.

lzap:
I am going to get the template fixed, sorry about that!

lzap:
I have rewritten this from scratch, but this is flawed - this cannot be done at the moment. We don't have RPM versioning scheme implementation in Ruby yet.

PR #680, opened 2020-01-17
Fixes #28784 - Updating nomenclature of each file

waldirio:
Updating nomenclature of each file for better customer presentation. Once we will be creating a huge number of reports, will be easy when the customer order by `Name`.

### DIFF
--- a/report_templates/ansible_inventory.erb
+++ b/report_templates/ansible_inventory.erb
@@ -1,5 +1,5 @@
 <%#
-name: Ansible Inventory
+name: Ansible - Ansible Inventory
 description: This template has a special meaning and should not be renamed.
 snippet: false
 template_inputs:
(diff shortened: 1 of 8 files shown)


lzap (approved):
Thanks, apologies for late reply. Looks good to me @ares ...

ares:
@lzap, this renamed many templates however didn't rename file names. Also when we'll seed these new names, many templates will be duplicated, once under their old name, second with new name. I assume you'll follow up with a migration in core that will drop old templates. Also I think template files should be renamed in this repo. Renaming Ansible Inventory to Ansible - Ansible Inventory will also require change in foreman_ansible, but I'll take care of it for you.

lzap:
@ares is file naming convention a real requirement? I don't like those `_-_` so if that does not need to be exact, I can make some more resonable file naming convention.

ares:
I think it is. If you import/export templates using foreman_template, you don't want export to save templates under a different name than from which you imported them.

lzap:
Migration pending review, I will rename the filenames in another PR here.

lzap:
And the PR is up.

PR #682, opened 2020-01-18
Fixes #28797 - fix subscription qty in entitlement report

snagoor:
As per the current implementation of entitlement report Quantity shown here https://github.com/theforeman/community-templates/blob/7c2ee7a486a534ca54e4f37173778f5134ff3cc5/report_templates/entitlements.erb#L28 is the total quantity of that specific subscription from the Pool.

It is required to show the quantity **consumed** not the over-all quantity of the Subscription pool.


tbrisker:
closing in favor of #684, Thanks @snagoor !

PR #684, opened 2020-01-27
Fixes #28797 - Add consumed into entitlements report

xprazak2:
Needs https://github.com/Katello/katello/pull/8542


ares (changes requested):
we need to bump katello version to 3.16 I believe

xprazak2:
Updated.

tbrisker (approved):
makes sense, thanks @xprazak2 !

PR #685, opened 2020-01-28
Fixes #28883 - Initial IPv6 support for kickstart

lzap:
This is initial change for PXE and kickstart to support IPv6, however during testing I encountered a bug in Anaconda installer, therefore this cannot be tested end to end. I'd appreciate if we could get this into the next upstream release to get feedback on this.

The patch introduces new snippet which unifies kernel command line across all PXE templates, previously we had all code duplicated in all templates. It also contains changes needed to setup IPv6 networking for provisioning.


provisioning_templates/iPXE/kickstart_default_ipxe.erb
```diff
@@ -9,47 +9,19 @@ oses:
...
-    if (is_fedora && os_major < 17) || (rhel_compatible && os_major < 7)
-      options.push("vlanid=#{iface.tag}")
-    else
-      options.push("vlan=vlan#{iface.tag}:#{iface.attached_to}")
-    end
-  end
-
-  if !is_fedora && os_major >= 7 && host_param_true?('fips_enabled')
-    options.push('fips=1')
+  static = ''
```
ekohl: Alternative:
```ruby
if subnet4 && !subnet4.dhcp_boot_mode?
  ks = foreman_url('provision', static: '1')
elsif subnet6 && !subnet6.dhcp_boot_mode?
  ks = foreman_url('provision', static6: '1')
else
  ks = foreman_url('provision')
```

Maybe the args could be stored in a hash?

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -91,14 +80,87 @@ selinux --<%= host_param('selinux-mode') || host_param('selinux') || 'enforcing'
...
+  # DNS
+  if nameservers.flatten.size > 0
+    network_options.push("--nameserver=#{nameservers.flatten.join(',')}")
+  else
+    network_options.push("--nodns")
+  end
+
+  # minimum MTU takes precedence for dual-stack setup
+  # min/max methods are not allowed: https://github.com/svenfuchs/safemode/pull/30
+  if mtus[1] && mtus[1] > mtus[0]
```
ekohl: I wonder if we should take a min() or max() approach. This implements max() but won't that lead to incompatibilities and MTU issues? Let's say IPv6 allows jumbo frames (MTU 9000) but IPv4 only standard (MTU 1500) it will cause a lot of fragmentation for IPv4. Not an expert in this so carefully consider this.

provisioning_templates/snippet/kickstart_kernel_options.erb
```diff
@@ -0,0 +1,108 @@
...
+  nic_delay ||= subnet6 ? subnet6.nic_delay : nil
+  if nic_delay && rhel_compatible && major >= 7
+    ["dhcp", "iflink", "ifup", "route", "ipv6dad", "ipv6auto", "carrier"].each do |type|
+      options.push("rd.net.timeout.#{type}=#{nic_delay}")
+    end
+  elsif nic_delay
+    options.push("nicdelay=#{nic_delay} linksleep=#{nic_delay}")
+  end
+-%>
+<%= options.join(' ') -%>
```
ekohl: Is the lack of a newline intentional?

ekohl (commented):
How do we handle the exceptions that are raised? I don't think we do it anywhere else in our templates.

provisioning_templates/iPXE/kickstart_default_ipxe.erb
lzap: Much better this one.

provisioning_templates/provision/kickstart_default.erb
lzap: Good point, this could create terrible pains and investigation. I am opting for explicit solution - if MTUs are different, then let's raise an exception.

provisioning_templates/snippet/kickstart_kernel_options.erb
lzap: Yes, this is added to another line actually, it's "line snippet". I've tested this and it renders fine.

ekohl: If you rely on it, I'd add a comment above it. It's too easy to miss in doing a review. Alternatively you can perhaps call `snippet().strip()` when using this if it's important?

provisioning_templates/provision/kickstart_default.erb
ekohl: In IPv6 each route can have its own MTU. I'm not sure what the best approach is.

@jovandeginste @vstone AFAIK you've both done your fair share of networking. Care to weigh in here?

provisioning_templates/snippet/kickstart_kernel_options.erb
lzap: Well as long as this is added as the very last element on that line it actually does not matter, I don't think it is worth the effort, it's pretty obvious when this is broken.

provisioning_templates/provision/kickstart_default.erb
lzap: No, provisioning interface can only have one IPv4 and one IPv6 network (which has MTU assigned in Foreman DB). Unless it does not differ, we don't care how MTU is set for other interfaces. Remember, I dropped the idea of configuring everything through Anaconda.

provisioning_templates/snippet/kickstart_kernel_options.erb
ekohl: Always adding it as the last element so the newline doesn't matter is another valid technical solution.

lzap: Added comment.

lzap:
Added tests which I partially extracted from the other PR, I realized that these are terrible mocks and there is no reason to invest much time in improving them. We need to work on enabling rendering via Foreman core.

provisioning_templates/provision/kickstart_default.erb
jovandeginste: @ekohl Never did different routes with different MTU on the same interface...

lzap:
Re exceptions - an error is thown and it is shown in the interface.

I am changing my mind on the newlines, actually this striked me hard when I was testing this. I haven't noticed but one of my PXE templates rendered with a newline and it's not obvious that you are missing some options unless you boot the broken system and see `/proc/cmdline` :-)

I will implement a way to cleanly render content on one line. The problem is not with a trailing newline, but a newline on the first line. I will push a temporary solution using `gsub` for now and probably will implement a change in `snippet` macro in core.

ekohl:
> I will implement a way to cleanly render content on one line. The problem is not with a trailing newline, but a newline on the first line. I will push a temporary solution using `gsub` for now and probably will implement a change in `snippet` macro in core.

Is `strip()` not allowed in the jail? A `strip` parameter to `snippet()` (and friends) makes sense to me as well since that's probably safer in the `if_exists` case.

lzap:
Good idea, `strip` even doesn't look bad and the code reads fine - no need to introduce another helper. I have tested it, it works in safe mode. I was also able to solve the IPv6 nameserver problem in dracut and I am able to confirm that HTTP UEFI Boot over IPv6 works fine both in libvirt and on bare metal server. This PR is ready for review however there is one missing bit - IPv6 interface will not be initialized yet, the ifcfg changes are planned and I will be working on them the week after FOSDEM/CfgCamp.

I wonder if it makes sense to merge this if there are no big problems @ekohl @ares @tbrisker maybe you could discuss this over a beer and decide. I would like to see this in 2.0 because this is a huge refactoring of PXE templates and little refactoring of kickstart template and although it does not deliver full IPv6 workflow yet (it installs, ifcfg is not configured properly) it would be good to let community users to test this.

ares:
I think this is a good addition and would be good to be merged before branching. @lzap also promised to do the testing during the test week of all variants, so we'll know provisioning works. @ekohl and @tbrisker any concern with merging this week?

tbrisker:
I'm not too happy about merging such a large change at this stage. i'm worried that again we will hit issues when syncing the templates to core (as we had in the 1.24 cycle). This also requires documenting the changes very well in the upgrade warnings section of the manual as there is a good chance that users using a combination of default and custom templates may see breakages.

If this is merged and synced to core before branching (i.e. this week) with no breakages, i'll be ok with that as long as the documentation will be ready for GA (and preferably sooner). otherwise, or if the sync to core leads to issues and risks delays, my finger on the revert button on Monday will be very light.

lzap:
I am back, well it's too late I believe. I will continue in this PR with the remaining ifcfg changes for post-2.0 merge then.

lzap:
I've heard you say we have branched off, can this be merged now, @tbrisker ? :-)

All: The patch is ready for final review. What works:

* IPv6 HTTP UEFI provisioning with an external DHCP
* IPv6 HTTP UEFI provisioning via a static IPv6
* DNS management (AAAA, PTR)

Major workarounds:

* PXE templates must contain `nameserver=` entries due to bug in Dracut/Anaconda (reported).
* Grub2 `grubx64.efi` on TFTP must be replaced with a one from Fedora (EL7/EL8 versions do not work).

I will try VLAN as well tomorrow due to request from a customer but this is out of scope for IPv6 initial support. I'd file a new PR for that one if this gets merged. @ares @ekohl

lzap:
@tbrisker can we merge this PR? I am ready to do followup if there are bugs found. I'd like to get this into nightly sooner than later.

lzap:
Rebased only, conflict. This is pain in a_s it is a major refactor, let's do this!

provisioning_templates/iPXE/kickstart_default_ipxe.erb
```diff
@@ -9,47 +9,19 @@ oses:
...
-  is_fedora = @host.operatingsystem.name == 'Fedora'
-  os_major = @host.operatingsystem.major.to_i
-  os_minor = @host.operatingsystem.minor.to_i
-  major = @host.operatingsystem.major.to_i
   iface = @host.provision_interface
-  mac = @host.provision_interface.mac
-  hostname = @host.name
-  subnet = @host.subnet
-  options = []
+  subnet4 = iface.subnet
```
ekohl, line 13: Nit: these variables are only used once so you might as well use them inline.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -94,14 +84,81 @@ selinux --<%= host_param('selinux-mode') || host_param('selinux') || 'enforcing'
...
+  if (subnet4 && !subnet4.dhcp_boot_mode?) || @static
+    network_options.push("--bootproto static")
+    network_options.push("--ip=#{iface.ip}")
+    network_options.push("--netmask=#{subnet4.mask}")
+    network_options.push("--gateway=#{subnet4.gateway}")
+  elsif subnet4 && subnet4.dhcp_boot_mode?
+    network_options.push("--bootproto dhcp")
+  end
+  if subnet4
+    nameservers.push(subnet4.dns_servers)
```
ekohl: This pushes an array onto an array. Not sure if this is allowed in safe mode, but this is probably what you intended.
```suggestion
    nameservers += subnet4.dns_servers
```

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -94,14 +84,81 @@ selinux --<%= host_param('selinux-mode') || host_param('selinux') || 'enforcing'
...
+      network_options.push("--ipv6gateway=#{subnet6.gateway}")
+    elsif subnet6 && subnet6.dhcp_boot_mode?
+      if host_param_true?('use-slaac')
+        network_options.push("--ipv6 auto")
+      else
+        network_options.push("--ipv6 dhcp")
+      end
+    end
+    if subnet6
+      nameservers.push(subnet6.dns_servers)
```
ekohl: Same as v4 nameservers
```suggestion
      nameservers += subnet6.dns_servers
```

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -94,14 +84,81 @@ selinux --<%= host_param('selinux-mode') || host_param('selinux') || 'enforcing'
...
+  if iface.virtual? && iface.tag.present? && iface.attached_to.present?
+    if rhel_compatible && os_major >= 6
+      network_options.push("--vlanid=#{iface.tag}")
+    elsif rhel_compatible && os_major >= 7
+      network_options.push("--interfacename=vlan#{iface.tag}")
+    end
+  end
+
+  # DNS
+  if nameservers.flatten.size > 0
```
ekohl: Flatten would not be needed if `+=` was used earlier

provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
```diff
@@ -41,6 +43,10 @@ DEFROUTE=<%= primary %>
 <%-     @subnet.dns_servers.each_index do |index| -%>
 <%=       "DNS#{index +1}=\"#{@subnet.dns_servers[index]}\"" %>
 <%-     end -%>
+<%-   elsif !@dhcp && @subnet6 -%>
+<%-     @subnet6.dns_servers.each_index do |index| -%>
+<%=       "DNS#{index +1}=\"#{@subnet6.dns_servers[index]}\"" %>
```
ekohl, line 48: This inded starts at 0 again so it should be:
```suggestion
<%=       "DNS#{index + @subnet.dns_servers.length + 1}=\"#{@subnet6.dns_servers[index]}\"" %>
```

Of course this also needs some check in case there is no IPv4 subnet.

test/kickstart/provision_test.rb
```diff
@@ -61,7 +62,7 @@
     options[:majors].each do |major|
       describe "#{distro} #{major}" do
         it 'will compile' do
-          code, stdout, stderr = @renderer.render(distro, major, options[:minor], options[:prefix], false)
+            code, stdout, stderr = @renderer.render(distro, major, options[:minor], options[:prefix], nil, false)
```
ekohl: Odd indenting
```suggestion
          code, stdout, stderr = @renderer.render(distro, major, options[:minor], options[:prefix], nil, false)
```

ekohl (commented):
Some minor comments, otherwise it looks good.

provisioning_templates/provision/kickstart_default.erb
lzap: It's quite the opposite, `push` is allowed but `+=` is not. That's why I don't use it: https://github.com/svenfuchs/safemode/blob/26d16d223630a0dbee3953aabbb1bfdd12da68ce/lib/safemode/core_jails.rb#L48-L55

provisioning_templates/iPXE/kickstart_default_ipxe.erb
lzap: I am trying here to set some "practice" to always have these variables explicitly set, so when you write `subnet4` it immediately strikes you there's also IPv6. This is intentional, these "headers" often get copied over in this repo.

provisioning_templates/provision/kickstart_default.erb
lzap: Oh I can use `concat` or `push(*array)`. Will fix.

provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
lzap: I am not sure what you mean, DNS entries do start with 1 according to the man page: https://www.systutorials.com/docs/linux/man/5-nm-settings-ifcfg-rh/

I'd rather not touch this, I've only copied this from IPv4 section and I would like to avoid re-testing this again if possible. It works just like this.

test/kickstart/provision_test.rb
lzap: Fixed

lzap:
Fixed indentation and concat problem.

provisioning_templates/snippet/kickstart_ifcfg_generic_interface.erb
ekohl: My bad, I missed the `elsif` and thought you could have both IPv4 and IPv6 DNS servers. Airport reviews are not the best reviews.

provisioning_templates/snippet/kickstart_kernel_options.erb
```diff
@@ -0,0 +1,115 @@
...
+      if @ipxe_net
+        options.push('ip=${netX/ip} netmask=${netX/netmask} gateway=${netX/gateway} dns=${dns}')
+      else
+        options.push("ip=#{iface.ip}", "netmask=#{subnet4.mask}", "gateway=#{subnet4.gateway}", "dns=#{dns_servers}")
+      end
+    else
+      if @ipxe_net
+        options.push("ip=${netX/ip}::${netX/gateway}:${netX/netmask}:#{hostname}:#{iface.identifier}:none nameserver=${dns}")
+      else
+        options.push("ip=#{iface.ip}::#{subnet4.gateway}:#{subnet4.mask}:#{hostname}:#{iface.identifier}:none")
```
ekohl, line 50: Aren't the nameservers missing here?

lzap: iPXE has a special case, we actually want to add `nameserver=${dns}` there. But that could be overriden by DNS from subnet later on, I've ruled that out in the code:

```ruby
  if subnet4 && !@ipxe_net
    subnet4.dns_servers.each { |server|
      options.push("nameserver=#{server}")
    }
  elsif subnet6 && !@ipxe_net
    subnet6.dns_servers.each { |server|
      options.push("nameserver=#{server}")
    }
  end
```

In case of iPXE, these won't be appended.

ekohl (approved):
Let's get this in and iterate on it if needed.

lzap:
Thanks a bunch!

PR #687, opened 2020-01-30
Fixes #28896 Add template name to kickstart provision templates

stejskalleos:
**Description**
Display name of the kickstart template in the `/root/original-ks.cfg` file.

### DIFF
--- a/provisioning_templates/provision/kickstart_default.erb
+++ b/provisioning_templates/provision/kickstart_default.erb
@@ -34,7 +34,7 @@ https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/in
 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/sect-kickstart-syntax
 %>
 
-# This kickstart file was rendered from provisioning template "<%= @template_name %>".
+# This kickstart file was rendered from the Foreman provisioning template "<%= @template_name %>".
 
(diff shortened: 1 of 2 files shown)


ares (changes requested):
Looks sane, however it may be good to provide more context, e.g. `# This kickstart file was rendered from provisioning template "<%= @template_name %>"`

stejskalleos:
Comment changed to `# This kickstart file was rendered from provisioning template "<%= @template_name %>".`

ezr-ondrej (approved):
If the goal is to have it only for kickstart templates, it looks good to me :)
Wouldn't it make sense to have it in more templates though?
Not necessarily in this PR :)

ezr-ondrej:
@ares any more comments?

ares (commented):
Just one tiny comment `from the Foreman provisioning template` would be better :-) I don't recall if debian stores preseed file somewhere and couldn't google ot quickly. @mmoll do you happen to know? If so, I'd say add it to preseed tempmate too, but not a blocker for this PR

stejskalleos:
> Just one tiny comment from the Foreman provisioning template would be better :-)

Updated.

ares (approved):
no more comments, kickstart can be solved in a separate repo

edit: preseed of course

ezr-ondrej (approved):
Thanks @stejskalleos! :+1:

PR #688, opened 2020-02-03
Fix virtual nic configuration in preseed

m-bucher:
The `ifconfig` way of configuring virtual-NICs (multiple IPs on one interface) on Debian distributions no longer works with recent versions:
https://wiki.debian.org/NetworkConfiguration#Multiple_IP_addresses_on_one_Interface

### DIFF
--- a/provisioning_templates/snippet/preseed_networking_setup.erb
+++ b/provisioning_templates/snippet/preseed_networking_setup.erb
@@ -47,7 +47,7 @@ EOF
 real=`ip -o link | awk '/<%= interface.mac -%>/ {print $2;}' | sed s/:$//`
 <% virtual = interface.virtual? -%>
 <% if virtual -%>
-real=`echo <%= interface.identifier -%> | sed s/<%= interface.attached_to -%>/$real/`
+real="<%= interface.attached_to -%>"
 <% end -%>
 
 cat << EOF >> /etc/network/interfaces


sbernhard:
@mmoll  can you confirm?

lzap:
Hello, sorry for the delay. I now see kickstart as well. Can you elaborate what changed? I am not following.

m-bucher:
@lzap apparently we had a similar problem with CentOS, therefore we added the fix for kickstart as well.

lzap:
Yeah, I just don't understand what's this all about. I don't get the description of the problem, why you are doing this. Maybe @ares can help me here?

lzap:
Guys, I am failing to understand this change. Please explain, or we can do a quick meeting if you want to sync up on this one, I am probably just blind.

m-bucher:
OK, so I removed the kickstart stuff for now and looked at all again.

The base of this is that on Debian, specifying virtual-networks using separate interfaces like `eth0:1` (which is the `ifconfig`-way of doing this) is deprecated and should now be added as additional IP address configuration to the base interface (iproute2 method).

That works (at least on my test-system) with the modified networking-snippet.
However, this is probably not the ideal solution, because it requires to set `attached_to` to the interface-name, which one might not know, yet because it will be chosen by the OS.

Probably the best solution would be to create a link between the base-interface and its virtual interfaces in the forman host definition and then use it.

lzap:
I guess I need @ares here. Marek, I am trying to figure out what we are trying to do for RH here:

```erb
<%#
name: kickstart_ifcfg_get_identifier_names
model: ProvisioningTemplate
snippet: true
model: ProvisioningTemplate
kind: snippet
-%>
<%- if @identifier -%>
<%=   "real=\"#{@identifier}\"" %>
<%- else -%>
<%=   "real=`grep -l #{@interface.inheriting_mac} /sys/class/net/*/{bonding_slave/perm_hwaddr,address} 2>/dev/null | awk -F '/' '// {print $5}' | head -1`" -%>
<%-   if @interface.virtual? -%>
<%=     "\nreal=`echo #{@interface.identifier} | sed s/#{@interface.attached_to}/$real/`" -%>
<%-   end -%>
<%- end -%>
<%# ifcfg files are ignored by NM if their name contains colons so we convert colons to underscore %>
sanitized_real=`echo $real | sed s/:/_/`
```

Would you mind commenting on what is the deal with the virtual `sed` there? Then I would like to put this into the context of this patch becuase this PR introduces a departure from what we do for RH systems.

tbrisker:

PR #689, opened 2020-02-07
Add authorized_keys to hosts

mdellweg:


provisioning_templates/snippet/ssh_authorized_keys.erb
```diff
@@ -0,0 +1,38 @@
...
+
+ssh_authorized_keys: public keys to be put in /root/.ssh/authorized_keys
+
+This template sets up SSH keys in any host so that as long as your public
+SSH key is in ssh_authorized_keys, you can SSH into a host.
+%>
+
+<% if !host_param('ssh_authorized_keys').blank? %>
+<% authorized_keys = host_param('ssh_authorized_keys') %>
+<% ssh_user = 'root' %>
```
ezr-ondrej: does ssh user has to be root?

ezr-ondrej (commented):
I believe this would be an awesome addition, but I feel like it would make sense while at it to make ti possible to at least choose the user. The best would be if we could somehow have keys for multiple users. The biggest issue is the potential risk that we add the keys to `root` user, but users may have already have some `ssh_authorized_keys` param for some other automation (Ansible i.e.).
Apart of it the snippet does basically the same thing as `remote_execution_ssh_keys` it would be great if we could somehow unify it into one.

Does it make sense? Or is there even a value to have it for `root`? Any thoughts?

provisioning_templates/snippet/ssh_authorized_keys.erb
mdellweg: The user could be configurable. Sure. The remote_execution_ssh_keys adds sudo to the corresponding user if needed. Also i think there is a way to add the ssh_key of the foreman user into the newly created machine. But, i think, all of those serve different purposes.

I tried to mimic the logic of the `ssh_authorized_keys` parameter as used here:
https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/provision/coreos_provision.erb#L32
And bring it to more platforms.

provisioning_templates/snippet/ssh_authorized_keys.erb
```diff
@@ -0,0 +1,38 @@
...
+<% ssh_path = "~#{ssh_user}/.ssh" %>
```
ekohl: I think `~#{ssh_user}` is a bashism and probably doesn't work on other shells.

provisioning_templates/snippet/ssh_authorized_keys.erb
```diff
@@ -0,0 +1,38 @@
...
+
+Parameters:
+
+ssh_authorized_keys: public keys to be put in /root/.ssh/authorized_keys
+
+This template sets up SSH keys in any host so that as long as your public
+SSH key is in ssh_authorized_keys, you can SSH into a host.
+%>
+
+<% if !host_param('ssh_authorized_keys').blank? %>
```
ekohl:
```suggestion
<% unless host_param('ssh_authorized_keys').blank? -%>
```

provisioning_templates/snippet/ssh_authorized_keys.erb
```diff
@@ -0,0 +1,38 @@
+<%#
+kind: snippet
+name: ssh_authorized_keys
+model: ProvisioningTemplate
+snippet: true
+%>
+
```
ekohl, line 7: This generates a lot of whitespace. Could you have a look at the various blocks?

ekohl (commented):
We also have https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/create_users.erb which is a more native way of handling SSH keys. Any reason to prefer a host param?

mdellweg:
@ekohl With host_params it can be set on the hostgroup level or even specific to domains or just about any other organization structure of foreman.
Also as i tried to express above the parameter `ssh_authorized_keys` is already used on the some cloud provisioning templates, and i think it would be nice to handle it in a consistent way across all platforms.

provisioning_templates/snippet/ssh_authorized_keys.erb
mdellweg: Not invented here... would you prefer calling `getent passwd #{ssh_user} | cut -d":" -f 6`?
https://github.com/theforeman/community-templates/blob/03281e9138e178f336958d1ef83739483c048078/provisioning_templates/snippet/create_users.erb#L12

ekohl: I guess it's safe enough and otherwise patches welcome :)

tbrisker:

PR #690, opened 2020-02-17
Fixes #29031 - resolved boot loop for Grub2

lzap:
All installations using pxegrub2 templates end up in boot loop after the provisioning is done and the target template is switch to the "pxegrub2 default local boot".

This is a small copy-paste error in global template. Requesting backport into 2.0 and 1.24 as this is a blocker for UEFI provisioning, this bug sneaked through my test week testing as it only appears at the very and of the installation phase (and I usually don't let it to finish due lack of time during those tests).


lzap (commented):
Works fine :-)

tbrisker:
2.0 - 6256baf
1.24 - 373501b

PR #691, opened 2020-02-24
Fixes #29141 - CloudInit is compatible w/ Kickstart def user data

ofedoren:


ofedoren:
@lzap, I heard you've got some experience with CloudInit, could you please review those changes?

provisioning_templates/cloud_init/cloud_init_default.erb
```diff
@@ -16,18 +16,49 @@ This template accepts the following parameters:
...
 fqdn: <%= @host %>
 manage_etc_hosts: true
 users: {}
 runcmd:
 - |
+<%= indent(2) { snippet 'fix_hosts' } -%>
+- |
+<% if proxy_uri -%>
+    echo 'proxy = <%= proxy_uri %>' >> /etc/yum.conf
+<% end -%>
```
lzap: Can you extract this into snippet and use from other places as well?

provisioning_templates/cloud_init/cloud_init_default.erb
```diff
@@ -16,18 +16,49 @@ This template accepts the following parameters:
...
+- |
+  echo "updating system time"
+  yum -y install ntpdate
+  /usr/sbin/ntpdate -sub <%= host_param('ntp-server') || '0.fedora.pool.ntp.org' %>
+  /usr/sbin/hwclock --systohc
+- |
```
lzap, line 38: Can you extract this into snippet and use from other places as well?

lzap (changes requested):
Other than that, ok.

ofedoren:
@lzap, done. Extracted into `ntp` and `yum_proxy` snippets.

provisioning_templates/snippet/ntp.erb
```diff
@@ -0,0 +1,10 @@
+<%#
+kind: snippet
+name: ntp
+model: ProvisioningTemplate
+snippet: true
+%>
+echo "updating system time"
+yum -y install ntpdate
+/usr/sbin/ntpdate -sub <%= host_param('ntp-server') || '0.fedora.pool.ntp.org' %>
+/usr/sbin/hwclock --systohc
```
lzap, line 31: One more thing once you on that. In the default kickstart we have this:

```
  use_ntp = host_param_true?('use-ntp') || (is_fedora && os_major < 16) || (rhel_compatible && os_major <= 7)

<% if use_ntp -%>
ntp
-chrony
<% else -%>
chrony
-ntp
<% end -%>
```

Esssentially, on older systems we do install ntp but on modern ones we do default to chrony, unless user opts-into the traditional ntp. Can you respect the very same in the snippet? Copy the check over and install either ntp or chrony. I'd also enable start the service as well: `systemctl enable --now service`.

lzap: Beware that `ntpdate` will not work if chrony is installed, use `chronyc -a makestep` instead.

lzap: Also, I don't like the fedora default pool, I'd probably just provide it an ntp-server if there is any, otherwise i'd keep the default server list that has been installed with the package.

ofedoren:
@lzap, updated.

I've tried to address your comments. Is it better now?

ofedoren:
@lzap, ping :)

lzap (approved):
Sorry for the delay, I haven't tested this end to end, will do during test week tho.

PR #693, opened 2020-02-26
Fixes #28627 - update entitlements report due to added macro

domitea:
I've added a `host_collections` macro in https://github.com/Katello/katello/pull/8586 So I have to update report template also.


xprazak2:
Works for me, maybe the new macro could be used in [Inventory template](https://github.com/theforeman/community-templates/blob/develop/report_templates/ansible_inventory.erb#L161) as well.

ares (commented):
I'm not against, however could this have been solved by `host.host_collection.map { |c| c.to_s }`? The new macro shields us from AR interface, which is great, however it's misleading a bit, because it's not clear it returns string representation of objects. Perhaps if it's call `host_collections_names` it would be better.

xprazak2:
What is the status here? @domitea, do you think this could be solved just by using `map` and therefore no changes in katello would be needed?

domitea:
@ares, @xprazak2 Yes, same behavior can be achieved by `.map(...) { ... }` call but I am fan of hiding core libraries like AR from user as far as I can to avoid possibility of usage some future security issue (like SQL injection attack for non escaped SQL query). So I've updated templates according to new macros created.

xprazak2:
Works as expected with the new macros in katello.

report_templates/ansible_inventory.erb
```diff
@@ -157,7 +157,7 @@ require:
       inventory_data.update('location': host.location) if input('Location') == 'yes'
       hostgroup = host.hostgroup.title if host.hostgroup
       inventory_data.update('host_group': hostgroup) if input('Host Group') == 'yes'
-      host_collections = host.host_collections.map { |c| c.name } if host.host_collections
+      host_collections = host_collections_names(host) if host.host_collections
```
jturel: I'm not sure if this is an equivalent change, but I'm unsure how to test this template. The value is now a comma separated string of names rather than an array of names. If this change isn't required I'd rather omit it.

domitea: I think that my addition is correct because actually we're trying to show multiple values in one cell of table. Also, this template is report template so it's intended for human, not machine (like job template) and that means that multiple value should have a delimiter.

ezr-ondrej: That's not entirely true @domitea. If I select json format as an output format, I might very likely intend to consume it by machine for further processing. That being said, we might want to include formaters for columns that would take care of different formatting of columns. what actually goes over this PR. But I believe the macro should return an array of values and templating take care of the formatting :thinking:

tbrisker: I agree too, macros shouldn't handle formatting. also comma-seperated list might mess up csv reports (not sure if commas in field values are escaped when rendering?)

domitea:
@jturel I've removed change in ansible inventory template. Is this good for you?

jturel (approved):
ACK

Thank you @domitea :+1:

tbrisker:
any further concerns @xprazak2 @ares or @ezr-ondrej ? I'd like to get this one in today if all your concerns have been addressed.

xprazak2:
I have no additional concerns.

ares:
Do we need to bump katello version in metadata/requirements? Iiuc we use new macros. Also after this is merged, we need to sync templates to core. Other than that, all green

tbrisker:
> Do we need to bump katello version in metadata/requirements? Iiuc we use new macros. Also after this is merged, we need to sync templates to core. Other than that, all green

Good point, I think this is worth cherry-picking to older katello versions but we'll need to get ack from katello folks about which version will be the minimal this will land in.

ares:
FTR I'd be ok with merging this as is and adding metadata tomorrow or after katello grooming.

ezr-ondrej:
> any further concerns @xprazak2 @ares or @ezr-ondrej ? I'd like to get this one in today if all your concerns have been addressed.

My (originally @jturel's) formatting comment https://github.com/theforeman/community-templates/pull/693#discussion_r390240101 got ignored out. have we tested it in json? Are we ok with it not being array?
The escaping should happen though, so it's just UX (usability) issue. I'm ok with solving it as follow up later.
Thus I'm ok if it get merged.

tbrisker (approved):
ok, based on all comments i'll go ahead and merge this for now, thanks @domitea and please open followup PRs (using "Refs #28627..." commit message) to address the concerns raised.

ares:
@domitea to be explicit, two concerns I understood:
1) we need to set minimum version of katello if we rely on new macros
2) I don't know how to fix correct structure other than just passing array of strings to report_row macro, either we need to change macros to return array of strings or this template to simply get objects and do `.map { |s| s.name }`, so macros wouldn't be required and we wouldn't have to update metadata

ezr-ondrej:
@domitea any progress with followup PR?

PR #694, opened 2020-02-26
Fixes #29191 - Use medium_uri.to_s in Kickstart default iPXE

ofedoren:
@lzap, not related to this PR, but I've noticed that in this template `stage2` option generated twice with the same value.

https://github.com/ofedoren/community-templates/blob/develop/provisioning_templates/iPXE/kickstart_default_ipxe.erb#L32 will set `inst.stage2` to `medium_uri` if host has `kickstart_liveimg` parameter set.

But also in https://github.com/ofedoren/community-templates/blob/develop/provisioning_templates/iPXE/kickstart_default_ipxe.erb#L52 the `inst.stage2` option will be always set to `medium_uri` (`inst.stage2=<%= medium_uri %>`)

Which makes me wonder if we rather should change the https://github.com/ofedoren/community-templates/blob/develop/provisioning_templates/iPXE/kickstart_default_ipxe.erb#L32 to something like
```ruby
stage2 = host_param('kickstart_liveimg') ? 'inst.stage2=' + host_param('kickstart_liveimg') : ''
```

Please correct me if I'm wrong.

### DIFF
--- a/provisioning_templates/iPXE/kickstart_default_ipxe.erb
+++ b/provisioning_templates/iPXE/kickstart_default_ipxe.erb
@@ -29,7 +29,7 @@ oses:
 
   options.push("linksleep=#{subnet.nic_delay}") if subnet.nic_delay.present?
 
-  stage2 = host_param('kickstart_liveimg') ? 'inst.stage2=' + medium_uri : ''
+  stage2 = host_param('kickstart_liveimg') ? 'inst.stage2=' + medium_uri.to_s : ''
 
   if (is_fedora && os_major < 17) || (rhel_compatible && os_major < 7)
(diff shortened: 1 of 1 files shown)


lzap:
Good find, I think the cleanest way is to simply push into `options` array if `host_param('kickstart_liveimg')` is set.

This will be actually fixed in the big refactoring patch which is almost to be merged, it only needs last blessing. If you can help pushing it then we could close this bug actually:

https://github.com/theforeman/community-templates/pull/685/files#diff-24b922531d54668ea63e8c5aeeb24b65R96-R98

lzap (approved):
You know what, let's merge this as a temporary workaround which can be backported and I will rebase on top of this.

PR #696, opened 2020-03-02
Fixes #29239 - unify discovery grub2 menu items

lzap:
By adding <%= suffix %> to the --id discovery, it is no longer possible to force systems to load the discovery image using the "Default PXE global template entry" in systems mixed with UEFI and non-UEFI machines. Adding the <%= suffix %> to the menu item test makes sense, changing the --id not so much.

This is a rewrite of this section from pregenerated ERB to Grub2 "shell-like" syntax. I have tested this and it works fine.

This also changes /httpboot/ to more expected /EFI/BOOT/: https://github.com/theforeman/smart-proxy/pull/750

@bennodepenno


lzap:
Moved RM issue, this is tested.

lzap:
@tbrisker @ekohl @ares @mmoll anyone? We need this in 2.0, this is tested.

mmoll:
merged, diky @lzap!

2.0: 413c189

tbrisker:
@lzap didn't you say this was dependent on the smart proxy change and not crucial for 2.0?

lzap:
It is, but I was thinking replacing `/EFI/BOOT` with `/httpboot`. The new `/EFI` alias depends on the new PR, however the old endpoint is present.

PR #697, opened 2020-03-03
Add job template for deploying receptor

adamruzicka:


job_templates/deploy_receptor-ansible_default.erb
```diff
@@ -0,0 +1,26 @@
+<%#
+name: Configure receptor
```
ares: suggestion: Connect Red Hat receptor controller

ares (approved):
Thanks @adamruzicka, merging!

PR #699, opened 2020-03-04
Fixes #29261 - Apply roles dynamically

adamruzicka:
We used to render the roles list on the server, which broke things with the move
to batch triggering, because it may have lead to application of roles which
weren't assign to a host.

With this change, we will take the roles from host's variables.


ares (approved):
Thanks @adamruzicka, merging!

PR #700, opened 2020-03-05
Fixes #29260 - Add -y flag for yum command

loitho:
Hi there,

Same as : https://github.com/Katello/katello/pull/8599

As described in : https://projects.theforeman.org/issues/29260
And https://community.theforeman.org/t/using-sudo-su-for-privilege-escalation/17635/4

The Katello Ansible Default templates that are using Ansible shell commands directly for creating Yum action like update are missing a "-y" flag which makes them hang

This fixes the problem by adding the missing -y flag on each template that was missing it.

The files here and on the Katello repository are identical

regards,

Thomas

### DIFF
--- a/job_templates/install_errata_-_katello_ansible_default.erb
+++ b/job_templates/install_errata_-_katello_ansible_default.erb
@@ -15,4 +15,4 @@ kind: job_template
 %>
 
 <% advisories = input(:errata).split(',').map { |e| "--advisory=#{e}" }.join(' ') %>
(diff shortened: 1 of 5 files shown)


ares:
Thanks @loitho, I quickly grepped and you got it all. Merging! I'll take care of syncing to foreman_ansible repo.

ares:
Oh there's nothing to sync to foreman_ansible, they just need to be synced to katello too.

PR #701, opened 2020-03-05
improved formatting

thalunil:


ekohl (approved):
This is a bugfix since the output is getting interpreted as comments without this patch. Looks like this was introduced in 6a7e738039965df8245c84c65a8a15b55b9ee89d so it's been there for a long time now.

lzap (approved):
Thanks.

PR #702, opened 2020-03-09
Adding LICENSE file to project

kshithijiyer:


ekohl (approved):
@tbrisker I don't think we explicitly chose a license, but since it was part of Foreman before (which is GPLv3) it's the only choice.

tbrisker (approved):
correct, this should be the same license as foreman core. Thanks @kshithijiyer !

PR #704, opened 2020-03-10
Use correct role name for deploying receptor

adamruzicka:
We expected the role to be receptor-satellite, however it was published on ansible-galaxy as `project-receptor.satellite_receptor_installer`.


ares (approved):
Thanks @adamruzicka, reflects the packaged version correctly now, merging!

PR #705, opened 2020-03-10
Refs #29037 - Update report template for Entitlements Expiring Soon

jeremylenz:
[SAT-E-459]  Updates the Entitlements report template to allow the user to filter to subscriptions expiring $soon.

* Includes a user input to choose $soon (7, 30, 60, 90, 120, or 180 days)
* Leave it blank to show all entitlements


jeremylenz:
see also: https://github.com/Katello/katello/pull/8606

jeremylenz:
To test, check out Katello PR#8606.  Clone the Entitlements report and replace the ERB code with that in this PR.   You'll also have to manually add the input, as the web UI doesn't import it automatically.  Or, open a rails console and do `SeedHelper.import_raw_template(File.read('path/to/entitlements.erb'))`

report_templates/entitlements.erb
```diff
@@ -2,17 +2,25 @@
 name: Subscription - Entitlement Report
 snippet: false
 model: ReportTemplate
+template_inputs:
+- name: days_from_now
```
ares: This can be human readable string, it's used as an input label in the form.

report_templates/entitlements.erb
```diff
@@ -2,17 +2,25 @@
...
+  input_type: user
+  description: "Show subscriptions expiring in <= this many days.  Leave blank to show all entitlements."
+  advanced: false
+  value_type: plain
+  options: "7\r\n30\r\n60\r\n90\r\n120\r\n180"
 require:
 - plugin: katello
   version: 3.14.0
 -%>
-<%- report_headers 'Name', 'Organization', 'Lifecycle Environment', 'Content View', 'Host Collections', 'Virtual', 'Guest of Host', 'OS',
```
ares, line 9: Why do you remove this? Headers are required, in case of `load_hosts` returns zero objects, the result would be empty report. However even empty dataset should contain the structure. Empty csv file caused confusion in the past.

report_templates/entitlements.erb
```diff
@@ -2,17 +2,25 @@
...
   version: 3.14.0
 -%>
-<%- report_headers 'Name', 'Organization', 'Lifecycle Environment', 'Content View', 'Host Collections', 'Virtual', 'Guest of Host', 'OS', 
-'Arch', 'Sockets', 'RAM', 'Cores', 'SLA', 'Products', 'Subscription Name', 'Subscription Type', 'Subscription Quantity', 
-'Subscription SKU', 'Subscription Contract', 'Subscription Account', 'Subscription Start', 'Subscription End', 'Subscription Guest' -%>
 <%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]]).each_record do |host| -%>
+<%-   days_from_now = input('days_from_now').to_i -%>
+<%-   should_filter = !input('days_from_now').empty? -%>
 <%-   host.pools.each do |pool| -%>
+<%-     next if should_filter && pool.expiring_in_days > days_from_now -%>
```
ares: This is filtering during the filter runtime, is there a way to convert this to SQL through a search condition passed to `load_hosts` macro? Notw that it accepts search syntax. This would have big performance benefit especially for people having many hosts.

ares (changes requested):
I'm happy to see such change, few comments

report_templates/entitlements.erb
jeremylenz: @ares I think I removed it by mistake when copying the ERB from the entitlements report in the web UI editor (it doesn't show there for some reason.) It seemed to work fine so I didn't catch it.  Will add it back.

report_templates/entitlements.erb
jeremylenz: @ares Thanks, updated to use `search` instead of filtering in the erb.
By using scoped_search and a custom external method (see the Katello PR) I am able to get scoped_search to return only hosts _with expiring pools_.  However, when iterating through each host, `host.pools` would still return _all_ of that host's pools, not just the expiring ones.  So, I also added an instance method to `Host`, `pools_expiring_in_days`, so that none of the filtering has to be done inside the report template.

report_templates/entitlements.erb
```diff
@@ -2,17 +2,32 @@
...
+- name: Days from Now
+  required: false
+  input_type: user
+  description: "Show subscriptions expiring in <= this many days.  Leave blank to show all entitlements."
+  advanced: false
+  value_type: plain
+  options: "7\r\n30\r\n60\r\n90\r\n120\r\n180"
 require:
 - plugin: katello
   version: 3.14.0
```
ares: This will likely need to be 3.16 or whather the linked PR will land in

report_templates/entitlements.erb
```diff
@@ -2,17 +2,32 @@
...
-'Subscription SKU', 'Subscription Contract', 'Subscription Account', 'Subscription Start', 'Subscription End', 'Subscription Guest' -%>
-<%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]]).each_record do |host| -%>
-<%-   host.pools.each do |pool| -%>
+<%- days_from_now = input('Days from Now').to_i -%>
+<%- should_filter = !input('Days from Now').empty? -%>
+<%- report_headers 'Name', 'Organization', 'Lifecycle Environment', 'Content View', 'Host Collections', 'Virtual', 'Guest of Host', 'OS', 'Arch', 'Sockets', 'RAM', 'Cores', 'SLA', 'Products', 'Subscription Name', 'Subscription Type', 'Subscription Quantity', 'Subscription SKU', 'Subscription Contract', 'Subscription Account', 'Subscription Start', 'Subscription End', 'Subscription Guest', 'Subscription Ends In' -%>
+<%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]], -%>
+<%- search: should_filter && "pools_expiring_in_days = #{days_from_now}"-%>
```
ares: nitpick, you can indent the search to keep it readable, whitespace are not significant inside erb

 ```erb
<%-        search: ... -%>
```

 also whitespace before the closing tag

File: report_templates/entitlements.erb [hunk 6]
```diff
@@ -34,7 +49,8 @@ require:
           'Subscription Account': pool.account_number,
           'Subscription Start': pool.start_date,
           'Subscription End': pool.end_date,
-          'Subscription Guest': registered_through(host)
+          'Subscription Guest': registered_through(host),
+          'Subscription Ends In': "#{pool.expiring_in_days} days"
```
ares: I'm not a native speaker, but perhaps `'Subscription Expires In` would sound better? please keep it if it sounds right as is :-)

report_templates/entitlements.erb
```diff
@@ -2,17 +2,32 @@
...
-<%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]]).each_record do |host| -%>
-<%-   host.pools.each do |pool| -%>
+<%- days_from_now = input('Days from Now').to_i -%>
+<%- should_filter = !input('Days from Now').empty? -%>
+<%- report_headers 'Name', 'Organization', 'Lifecycle Environment', 'Content View', 'Host Collections', 'Virtual', 'Guest of Host', 'OS', 'Arch', 'Sockets', 'RAM', 'Cores', 'SLA', 'Products', 'Subscription Name', 'Subscription Type', 'Subscription Quantity', 'Subscription SKU', 'Subscription Contract', 'Subscription Account', 'Subscription Start', 'Subscription End', 'Subscription Guest', 'Subscription Ends In' -%>
+<%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]], -%>
+<%- search: should_filter && "pools_expiring_in_days = #{days_from_now}"-%>
+<%- ).each_record do |host| -%>
+<%-   if should_filter -%>
```
ares: I do not insist however I think it's a good practice to keep logic in app code base and not in the report. So I'd suggest one more thing to avoid the "filter or not" logic. Could you add one more method to host in Katello PR, something like `host.pools_expiring(days)` which would call `host.pools_expiring_in_days(days)` if `days` is present, `host.pools` otherwise? That way you wouldn't even have to define `should_filter` variable here and just call it. It would either return all pools or pools expiring in a given number of days. Bonus points for doing `.to_i` in there too, to avoid `days_from_now` definition here too.

ares (changes requested):
Nice! That is way better performance wise. Few more rather small suggestions, but it's really close I think.

report_templates/entitlements.erb
jeremylenz: @ares Thanks, I ended up updating `pools_expiring_in_days` so that it handles both string and integer arguments.  I kept the `should_filter` variable assignment just for readability, but the logic (and the `.to_i`) is now moved into Katello.

report_templates/entitlements.erb
```diff
@@ -12,20 +12,15 @@ template_inputs:
   options: "7\r\n30\r\n60\r\n90\r\n120\r\n180"
 require:
 - plugin: katello
-  version: 3.14.0
+  version: 3.16.0
 -%>
-<%- days_from_now = input('Days from Now').to_i -%>
+<%- days_from_now = input('Days from Now') -%>
 <%- should_filter = !input('Days from Now').empty? -%>
```
ares:
```erb
<%- should_filter = !input('Days from Now').empty? -%>
```
can be now written as
```erb
<%- should_filter = !days_from_now.empty? -%>
```
or even better

```erb
<%- should_filter = days_from_now.present? -%>
```

report_templates/entitlements.erb
```diff
@@ -12,20 +12,15 @@ template_inputs:
...
-<%- search: should_filter && "pools_expiring_in_days = #{days_from_now}"-%>
-<%- ).each_record do |host| -%>
-<%-   if should_filter -%>
-<%-     pools_to_show = host.pools_expiring_in_days(days_from_now) -%>
-<%-   else -%>
-<%-     pools_to_show = host.pools -%>
-<%-   end -%>
-<%-   pools_to_show.each do |pool| -%>
+<%-   search: should_filter && "pools_expiring_in_days = #{days_from_now}" -%>
+<%-   ).each_record do |host| -%>
```
ares: this line indentation seems off, sorry if I was not clear in the last comment, this line now needs minus 2 spaces.

ares (changes requested):
looks great now, last two nits, then I'll retest with katello PR and we're good to go, thanks!

jeremylenz:
Update to address comments and incorporate changes to the Katello PR

report_templates/entitlements.erb
```diff
@@ -2,17 +2,27 @@
...
 snippet: false
 model: ReportTemplate
+template_inputs:
+- name: Days from Now
+  required: false
+  input_type: user
+  description: "Show subscriptions expiring in <= this many days.  Leave blank to show all entitlements."
+  advanced: false
+  value_type: plain
+  options: "7\r\n30\r\n60\r\n90\r\n120\r\n180"
```
ares: this will force this parameter to be present, emptry string is not an option here, but there's probably no good way of add it here

jeremylenz:
Rebased to incorporate #693

ares (changes requested):
Hello, I've tested this with Katello PR and works nicely! There's one limitation found while testing. The parameter is not always required even though `required: false`. This seems as limitation of template metadata not being able to keep empty string as one of the value. I suppose you created this template via export which omtis it. But also even if added manually, the import still ignores it. The only workaround I can think of is having some value like "no limit" and instead of `days_from_now.present?` we'd change this to `days_from_now != 'no limit'`. In such case, that should be also the default value. Is that acceptable or do we need to fix emptry string handling first?

report_templates/entitlements.erb
jeremylenz: @ares Thanks, updated the template.  I'm not sure why I didn't come across this in my testing, maybe because I was manually pasting the ERB into the web UI instead of using `import_raw_template`?

I had to add a line
```
days_from_now = "" if days_from_now == 'no limit'
```
The Katello PR (just merged) was tuned to elegantly handle when `days_from_now` is an empty string, but broke everything when it was `'no limit'`.  Rather than introducing another Katello PR, I added the line here which seems to work fine.  Hopefully this is acceptable.

ares: > I was manually pasting the ERB into the web UI instead of using `import_raw_template`?

I guess you have defined the input manually and that allows empty string as input.

> I had to add a line

good thinking, I like that, let me retest this, if that works, I'm happy to merge

report_templates/entitlements.erb
```diff
@@ -2,17 +2,28 @@
...
-<%-   host.pools.each do |pool| -%>
+<%- days_from_now = input('Days from Now') -%>
+<%- days_from_now = "" if days_from_now == 'no limit' -%>
+<%- should_filter = days_from_now.present? -%>
+<%- report_headers 'Name', 'Organization', 'Lifecycle Environment', 'Content View', 'Host Collections', 'Virtual', 'Guest of Host', 'OS', 'Arch', 'Sockets', 'RAM', 'Cores', 'SLA', 'Products', 'Subscription Name', 'Subscription Type', 'Subscription Quantity', 'Subscriptions Consumed', 'Subscription SKU', 'Subscription Contract', 'Subscription Account', 'Subscription Start', 'Subscription End', 'Subscription Guest', 'Subscription Expires In' -%>
+<%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]], -%>
+<%-   search: should_filter && "pools_expiring_in_days = #{days_from_now}" -%>
+<%- ).each_record do |host| -%>
+<%-   host.pools.expiring_in_days(days_from_now).sort_by { |p| p.end_date }.each do |pool| -%>
```
ares, line 24: sort_by is actually not allowed in safe mode, getting an error now (not sure how I missed that before), we'll either need to allow that or use 

```
.sort { |a,b| a.end_date <=> b.end_date }
```

ares (changes requested):
one thing that needs change which I realized during testing

report_templates/entitlements.erb
ekohl: IMHO sort_by should be allowed

jeremylenz: @ares @ekohl I added `sort_by` in the Katello PR [here](https://github.com/jeremylenz/katello/blob/bda88d6bf38f5e087e1e186fd5885f2643b61279/app/models/katello/pool.rb#L120) (merged now)

report_templates/entitlements.erb
jeremylenz: @ares Did you get a chance to retest?

jeremylenz:
changed 'Subscription Expires In' to 'Days Remaining' as per [jturel comment](https://github.com/Katello/katello/pull/8638#discussion_r401773279)

jturel:
Can this be merged?

ares:
I'm sorry, I'll try to get back to this soon, I just didn't have time to retest, hopefully this week

report_templates/entitlements.erb
```diff
@@ -2,17 +2,28 @@
...
+<%- days_from_now = "" if days_from_now == 'no limit' -%>
+<%- should_filter = days_from_now.present? -%>
+<%- report_headers 'Name', 'Organization', 'Lifecycle Environment', 'Content View', 'Host Collections', 'Virtual', 'Guest of Host', 'OS', 'Arch', 'Sockets', 'RAM', 'Cores', 'SLA', 'Products', 'Subscription Name', 'Subscription Type', 'Subscription Quantity', 'Subscriptions Consumed', 'Subscription SKU', 'Subscription Contract', 'Subscription Account', 'Subscription Start', 'Subscription End', 'Subscription Guest', 'Days Remaining' -%>
+<%- load_hosts(includes: [:lifecycle_environment, :operatingsystem, :architecture, :content_view, :organization, :reported_data, :subscription_facet, :pools => [:subscription]], -%>
+<%-   search: should_filter && "pools_expiring_in_days = #{days_from_now}" -%>
+<%- ).each_record do |host| -%>
+<%-   host.pools.expiring_in_days(days_from_now).sort_by { |p| p.end_date }.each do |pool| -%>
 <%-     report_row(
-          'Name': host.name,
+          'Host Name': host.name,
```
ares, line 26: this needs to be reflected also on line 20 in report_headers

ares (changes requested):
I've tested with latest nightlies and works fine. Just one last thing, then can be merged (will conflict with https://github.com/theforeman/community-templates/pull/712 though). You only need to rename the file.

report_templates/entitlements.erb
jeremylenz: @ares Updated, thanks

jeremylenz:
Updated file name to match #712

ares (approved):
Thanks for your patience @jeremylenz, merging now!

ares:
oh, the filename was different, I'll submit a different PR to fix that

jeremylenz:
@ares Oops, sorry I missed that! :fearful:

PR #706, opened 2020-03-11
Use pxe loader helper

lzap:
To prevent nil pointer exception a helper from core could be used instead making this if statement too complex. A core patch needed before this can be merged!


ares (approved):
Thanks @lzap, merging!

ares (approved):

PR #708, opened 2020-03-12
Match receptor installer role's expectations

adamruzicka:
It expects satellite_url to end with a slash


adamruzicka:
Fixed it in the role, closing

PR #711, opened 2020-03-18
Refs #28782 - updates compare host report

lzap:
This is a rewrite of the original report, unfortunately this is flawed by design - gem version comparison is not compatible with RPM NVRE versioning scheme. It will only work for versions to some degree, RPM versioning is much more complicated. I'd rather remove this template, for now I am putting a warning comment in that report at least. For more info:

https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/

To get this implemented correctly, it would need to implement full versioning parser as it's implemented in Pulp.

This fixes https://github.com/theforeman/community-templates/pull/679


report_templates/host_compare_content_hosts_packages.erb
```diff
@@ -16,40 +16,38 @@ template_inputs:
...
-        'Package': data['Package'],
-        "#{host1.name}": data[host1.name],
-        "#{host2.name}": data[host2.name],
-        'Architecture': arch_list[0],
-        'Status': status
-      )
--%>
-<%- end -%>
-<%= report_render -%>
+<%
```
ares: why do you switch to single erb tag? if you want to start a new best practice, please update all other reports, however I don't think it's actually better

also I believe this does not work, report_render needs to be in `<%=` so that it actually prints the report

ares (changes requested):
Could you please list few examples where we'd get wrong results? to see how off we are with gem version comparison. Katello seems to have NVREA parsing helpers, could we use those? see ./app/lib/katello/util/package.rb`

lzap:
Both release and epoch when in use are giving wrong results. Example: `1.0-1` == `1.0-2` according to this code, while its not true for the actual RPM package.

report_templates/host_compare_content_hosts_packages.erb
lzap: The missing `<%=` was a typo thanks for that.

I won't be fixing all templates in this patch, sorry. A single tag is million times more friendly to edits than what is the "best practice", but it is solely my opinion and I accept that other people might like different things.

lzap:
Rebased, based on the gem compare.

I think we should either remove this template (my bad I merged it no doubt) or to keep it in this simple form which often will give incorrect results. I don't think we should be solving this NVRE problem in a template, there should be a feature built-in katello if there is demand for this. Currently, Katello codebase does not offer any method that compares two versions, looks like quite complex problem: https://docs.pulpproject.org/en/2.16/plugins/pulp_rpm/tech-reference/sort-index.html

I don't feel competent to tell or implement this, I believe that version resolution is done in Pulp via RPM libraries it uses.

lzap:
Rebased, @ezr-ondrej what you think? It's not perfect but very much improved from what we had initially. Maybe if we can squeeze this one into 2.1?

report_templates/host_-_compare_content_hosts_packages.erb
```diff
@@ -16,40 +16,38 @@ template_inputs:
...
+    1
+  else
+    gem_version_compare(version1, version2)
+  end
+
+  report_row(
+    "Package": package_name,
+    "Status": status,
+    "Version #{host1.name}": version1,
+    "Version #{host2.name}": version2,
```
ezr-ondrej:
```suggestion
    "Version on #{host1.name}": version1,
    "Version on #{host2.name}": version2,
```

report_templates/host_-_compare_content_hosts_packages.erb
```diff
@@ -16,40 +16,38 @@ template_inputs:
...
+    "Architecture #{host1.name}": arch1,
+    "Architecture #{host2.name}": arch2,
```
ezr-ondrej:
```suggestion
    "Architecture on #{host1.name}": arch1,
    "Architecture on #{host2.name}": arch2,
```

report_templates/host_-_compare_content_hosts_packages.erb
```diff
@@ -16,40 +16,38 @@ template_inputs:
...
+  )
+end
 -%>
-<%- end -%>
-<%= report_render -%>
+<%= report_render -%>
```
ezr-ondrej: missing last '\n'

report_templates/host_-_compare_content_hosts_packages.erb
```diff
@@ -1,5 +1,5 @@
 <%#
-name: Host - compare content hosts packages
+name: Host - Compare content hosts packages
```
ezr-ondrej: Isn't this cause troubles? Will we need migration just because of this?

ezr-ondrej (commented):
Generally I'm all in! Appart of the naming issue just nitpicks, will test this and let get this in.

report_templates/host_-_compare_content_hosts_packages.erb
ezr-ondrej: Oh it's merged recently?

lzap: Good catch, I did not mean to change the name.

lzap:
Rebased, thanks.

tbrisker:

PR #713, opened 2020-03-20
Hide inputs from job when configuring receptor

adamruzicka:


ares:
Thanks @adamruzicka, merging!

PR #714, opened 2020-03-30
Refs #29239 - unify discovery grub2 menu items

lzap:
I meant to backport it with slight change. The `/EFI` path is not yet available, but `/httpboot` will do the job.


tbrisker (approved):
Thanks @lzap !

PR #715, opened 2020-04-02
Refs #29464 - hide the password value in CCC playbook

ares:
requires https://github.com/theforeman/foreman/pull/7559

### DIFF
--- a/job_templates/configure_cloud_connector_-_ansible_default.erb
+++ b/job_templates/configure_cloud_connector_-_ansible_default.erb
@@ -7,11 +7,13 @@ template_inputs:
   input_type: user
   advanced: false
   value_type: plain
+  hidden_value: false
 - name: satellite_password
   required: true
   input_type: user
   advanced: false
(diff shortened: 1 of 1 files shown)


adamruzicka (approved):
Works nicely, ACK

job_templates/configure_cloud_connector_-_ansible_default.erb
```diff
@@ -7,11 +7,13 @@ template_inputs:
   input_type: user
   advanced: false
   value_type: plain
+  hidden_value: false
```
ezr-ondrej, line 10: this is a default, so it should not be necessary, right?

ares: it is not, however exporting the template would add it (same as e.g. advanced: false above)

ezr-ondrej:
Thanks @ares and thanks @adamruzicka :+1:

PR #718, opened 2020-04-07
Eject bootdisk before host built notification

sbernhard:
See https://github.com/theforeman/foreman_bootdisk/pull/92, too


lzap (approved):
Makes sense alone, thanks.

PR #720, opened 2020-04-22
Fixes #29590 - All Installed Packages template

waldirio:
Hello, here the new report template to generate the complete list of installed packages.

### DIFF
--- a/report_templates/host_-_all_installed_packages.erb
+++ b/report_templates/host_-_all_installed_packages.erb
@@ -10,15 +10,15 @@ template_inputs:
   resource_type: Katello::ActivationKey
   hidden_value: false
 model: ReportTemplate
-%>
-<%  load_hosts(search: input('hosts')).each_record do |host| -%>
-<%    host.installed_packages.each do |pkg| -%>
+-%>
+<%-  load_hosts(search: input('hosts')).each_record do |host| -%>
+<%-    host.installed_packages.each do |pkg| -%>
 <%-   report_row(
         'Host': host.name,
         'Package Name': pkg.name,
         'Package NVRA': pkg.nvra,
         'Package NVREA': pkg.nvrea
       )-%>
-<%    end -%>
-<%  end -%>
+<%-    end -%>
+<%-  end -%>
 <%= report_render -%>
--- a/report_templates/host_-_all_installed_packages.erb
+++ b/report_templates/host_-_all_installed_packages.erb
@@ -6,7 +6,7 @@ template_inputs:
   required: false
   input_type: user
   advanced: false
-  value_type: plain
+  value_type: search
   resource_type: Katello::ActivationKey
   hidden_value: false
 model: ReportTemplate
--- a/report_templates/host_-_all_installed_packages.erb
+++ b/report_templates/host_-_all_installed_packages.erb
@@ -11,7 +11,7 @@ template_inputs:
   hidden_value: false
 model: ReportTemplate
 -%>
-<%-  load_hosts(search: input('hosts')).each_record do |host| -%>
+<%-  load_hosts(search: input('hosts'), preload: [:installed_packages]).each_record do |host| -%>
 <%-    host.installed_packages.each do |pkg| -%>
 <%-   report_row(
         'Host': host.name,
--- a/report_templates/host_-_all_installed_packages.erb
+++ b/report_templates/host_-_all_installed_packages.erb
@@ -1,0 +1,24 @@
+<%#
+name: Host - All Installed Packages
+snippet: false
+template_inputs:
+- name: hosts
+  required: false
+  input_type: user
+  advanced: false
+  value_type: plain
+  resource_type: Katello::ActivationKey
+  hidden_value: false
+model: ReportTemplate
+%>
+<%  load_hosts(search: input('hosts')).each_record do |host| -%>
+<%    host.installed_packages.each do |pkg| -%>
+<%-   report_row(
+        'Host': host.name,
+        'Package Name': pkg.name,
+        'Package NVRA': pkg.nvra,
+        'Package NVREA': pkg.nvrea
+      )-%>
+<%    end -%>
+<%  end -%>
+<%= report_render -%>


waldirio:
Hey @ares good morning, do you believe to be interesting add a new input and filter by the package name? If yes, could you tell me how to do that? I can implement and push the change. I was trying to do that but no success.

Thank you
Waldirio

report_templates/host_-_all_installed_packages.erb
```diff
@@ -0,0 +1,24 @@
+<%#
+name: Host - All Installed Packages
+snippet: false
+template_inputs:
+- name: hosts
+  required: false
+  input_type: user
+  advanced: false
+  value_type: plain
+  resource_type: Katello::ActivationKey
```
adamruzicka, line 10: I'm pretty sure this shouldn't be here

waldirio: > I'm pretty sure this shouldn't be here

I just create a new report template and export it, that said, this is part of the export process. I'm not sure if is totally mandatory during the installation/creation. If not, let me know and I'll update the commit.

Thank you

ares: the resource_type is listed even though only useful for value_type: search

I actually think having that instead of plain would be beneficial, this makes the field smarted and allows autocompletion in the generate form, @waldirio could you please update that?

report_templates/host_-_all_installed_packages.erb
```diff
@@ -0,0 +1,24 @@
...
+  hidden_value: false
+model: ReportTemplate
+%>
+<%  load_hosts(search: input('hosts')).each_record do |host| -%>
```
ares: if you could s/<% /<%-/ here and on the line 16, 22 and 23, that would be great

report_templates/host_-_all_installed_packages.erb
```diff
@@ -0,0 +1,24 @@
...
+template_inputs:
+- name: hosts
+  required: false
+  input_type: user
+  advanced: false
+  value_type: plain
+  resource_type: Katello::ActivationKey
+  hidden_value: false
+model: ReportTemplate
+%>
```
ares: this needs to be `-%>` otherwise the report would always start with \n if I'm not mistaken, I thought we fixed that in the export, perhaps you used older version of Foreman for the export?

ares:
so good news is, the searching on package name should be possible, your host search should allow that by "installed_package_name = mc", this is the standard search query

to avoid n+1 issue, you'll also need to add `preload: :installed_packages` to load_hosts macro

all the notes from above though are small things, this is in a pretty good shape already!

report_templates/host_-_all_installed_packages.erb
waldirio: Hello @ares I'm doing this report on Sat6.7. In this version, still as above. Also, the line `<%  load_hosts(search: input('hosts')).each_record do |host| -%>` still in the beginning `first line`.

Hope this help.
Waldirio

report_templates/host_-_all_installed_packages.erb
waldirio: @ares this is an interesting point, when adding like below we can see new lines
```
<% ... %>
```
Then moving forward and keeping only like this, working as expected
```
<% ... -%>
```
However, when doing as below, looks like indifferent
```
<%- ... -%>
```

Insights about it?

Thank you

waldirio: @ares code updated according to your request.

Thank you!

waldirio:
> so good news is, the searching on package name should be possible, your host search should allow that by "installed_package_name = mc", this is the standard search query
> 
> to avoid n+1 issue, you'll also need to add `preload: :installed_packages` to load_hosts macro
> 
> all the notes from above though are small things, this is in a pretty good shape already!

Hey @ares all the changes were made, however, I'm a bit lost of how to proceed to add the package search. Could you help me with that?

Thank you

report_templates/host_-_all_installed_packages.erb
ares: Just for the record <%- trims ghe whitespace left of the opening tag. Usually that is not needed, the opening tag starts on the beginning of the line. However it has the same length as <%=, so the inner code is nicely indented, while it does no harm.

report_templates/host_-_all_installed_packages.erb
ares: Seems some fixes didn't make it to 6.7, I was able to reproduce the first line issue too

ares:
The thing is, you don't need to add any new parameter for searching based on installed packages. You can use the hosts input you already have. For example, use the search syntax I mentioned above to filter only hosts that have mc installed.

There are still 2 changes pending.

1) add preload to load_hosts as I mentioned in previous comment
2) make the hosts input of search type on resource Host instead of a plain.

ares (changes requested):
Waiting on contributor now, @waldirio ping me on IRC if you need a hand with finishing these two details.

report_templates/host_-_all_installed_packages.erb
waldirio: hey @ares change made. Thank you!

report_templates/host_-_all_installed_packages.erb
waldirio: @ares perfect. Thanks for the explanation.

waldirio:
> The thing is, you don't need to add any new parameter for searching based on installed packages. You can use the hosts input you already have. For example, use the search syntax I mentioned above to filter only hosts that have mc installed.
> 
> There are still 2 changes pending.
> 
>     1. add preload to load_hosts as I mentioned in previous comment
> 
>     2. make the hosts input of search type on resource Host instead of a plain.

@ares I got your point, when using for example `installed_package_name = mc` I can see the machine that has already the package `mc` installed, however, will be listed all the packages from that Content host.

What I was thinking is, let's suppose I'm looking for the make package, then I'm going to proceed with the search and the response could be something like below
```
dhcp-8-29-169.lab.eng.rdu2.redhat.com,make,make-3.82-23.el7.x86_64,make-3.82-23.el7.x86_64
server2
server3
server4,make,make-3.82-23.el7.x86_64,make-3.82-23.el7.x86_64
server5,make,make-3.82-23.el7.x86_64,make-3.82-23.el7.x86_64
```

Then the customer will be able to get the complete list of Content Hosts with that single package information. For sure, generating a complete report, the customer should be able to filter and confirm which CH has or not the package but this is a great feature.

thank you as always for your help.

waldirio:
@ares let me know if anything else is pending.

Best
Waldirio

ares:
Ah, I think I now better understand what you want to achieve. You still want to list host but if a specific package is mentioned, you want to give the information that the package is **not present** on the host. The `installed_packaged` is not allowed in safe mode, we'll need to add that. I'll also think about how we could limit loaded packages in an efficient way.

tbrisker:

PR #721, opened 2020-04-22
Remove wrongly named report

ares:
As part of recent renaming, one template got wrong name. Deleting that now, the most recent version is called 

`subscription_-_entitlement_report.erb`
not
`subscription-_-entitlement_report.erb`

The later can be safely removed


jturel (approved):

jturel:
Thanks Marek!

ezr-ondrej (approved):
Thanks @ares and @jturel !

PR #723, opened 2020-04-25
new report template: host - system currency

giovannisciortino:
A new report template to extract "System Currency" report.
It generate a score for each host based of a weighted sum of errata applicable to the host.
It was present in previous version of Satellite 5 and some users have interest to have it also in Satellite 6 (ie. https://bugzilla.redhat.com/show_bug.cgi?id=1477117 )

### DIFF
--- a/report_templates/host_-_system_currency.erb
+++ b/report_templates/host_-_system_currency.erb
@@ -1,5 +1,5 @@
 <%#
-name: Host - System currency
+name: Host - System errata currency
 snippet: false
 template_inputs:
 - name: Hosts filter
--- a/report_templates/host_-_system_currency.erb
+++ b/report_templates/host_-_system_currency.erb
@@ -1,0 +1,73 @@
+<%#
+name: Host - System currency
+snippet: false
+template_inputs:
+- name: Hosts filter
+  required: false
+  input_type: user
+  description: Limit the report only on hosts found by this search query. Keep empty
+    for report on all available hosts.
+  advanced: false
+  value_type: search
+  resource_type: Host
+- name: Errata filter
+  required: false
+  input_type: user
+  description: Limit the report only on errata found by this search query. Keep empty
+    for report on all available errata.
+  advanced: false
+model: ReportTemplate
+require:
+- plugin: katello
+￼ version: 3.9.0
+-%>
+<%  # multiplier for critical security errata -%>
+<%- sc_crit = 32 -%>
+<%  # multiplier for important security errata -%>
+<%- sc_imp = 16 -%>
+<%  # multiplier for moderate important security errata -%>
+<%- sc_mod = 8 -%>
+<%  # multiplier for low important security errata -%>
+<%- sc_low = 4 -%>
+<%  # multiplier for bugfix errrata -%>
+<%- sc_bug = 2 -%>
+<%  # multiplier for ehnancement errata -%>
+<%- sc_enh = 1 -%>
+<%- report_entries = [] -%>
+<%- load_hosts(search: input('Hosts filter'), includes: [:applicable_errata]).each_record do |host| -%>
+<%-   critical = 0 -%>
+<%-   important = 0 -%>
+<%-   moderate = 0 -%>
+<%-   low = 0 -%>
+<%-   bugfix = 0 -%>
+<%-   enhancement = 0 -%>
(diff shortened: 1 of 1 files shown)


report_templates/host_-_system_currency.erb
```diff
@@ -0,0 +1,73 @@
...
+<%-   low = 0 -%>
+<%-   bugfix = 0 -%>
+<%-   enhancement = 0 -%>
+<%-   host_applicable_errata_filtered(host, input('Errata filter')).each do |erratum| -%>
+<%-   critical+=1 if erratum.errata_type == 'security' and erratum.severity == 'Critical' -%>
+<%-   important+=1 if erratum.errata_type == 'security' and erratum.severity == 'Important' -%>
+<%-   moderate+=1 if erratum.errata_type == 'security' and erratum.severity == 'Moderate' -%>
+<%-   low+=1 if erratum.errata_type == 'security' and erratum.severity == 'Low' -%>
+<%-   bugfix+=1 if erratum.errata_type == 'bugfix' -%>
+<%-   enhancement+=1 if erratum.errata_type == 'enhancement' -%>
```
ezr-ondrej, line 50: wouldn't this deserver to live in `katello` itself? like `security_impact_score` with most of the impact scoring logic?
Or this is not interesting elsewhere than this report?

giovannisciortino: Currently I don't see any other report in this repository or in katello repository using the same logic. I tried to search some reference about the `security_impact_score` that you mentioned.

ezr-ondrej: I don't believe there is some, but I think it could be beneficial somewhere else as well and I belive that the logic should be in Katello itself. It would look much cleaner. Thus I was suggesting to implement that method there :)

ares (changes requested):
I agree with @ezr-ondrej, this logic should ideally live in Katello so that we can e.g. show the currency on host detail page too. @giovannisciortino would you be interested in moving the code there? That would simplify this template too.

report_templates/host_-_system_currency.erb
```diff
@@ -0,0 +1,73 @@
+<%#
+name: Host - System currency
```
ekohl: When I read this I expected a report on the values of `LC_CURRENCY` but turns out that's `LC_MONETARY`. Perhaps:
```suggestion
name: Host - System errata currency
```

giovannisciortino:
@ares I would be interested but I haven't enough knowledge of Katello repository to identify where this part of report logic should be moved and how justify the existence of this logic in Katello repository

ares:
Alrighty, that's good, let me help you. The method to calculate the currency for the host should be added to Content Facet definition which [lives here](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/content_facet_host_extensions.rb), just put it as a new method before [this end](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/content_facet_host_extensions.rb#L60).

It should look like this
```ruby
def errata_currency
  critical = 0
  important = 0
  ...
  score = critical*sc_crit + important*sc_imp + moderate*sc_mod + low*sc_low + bugfix*sc_bug + enhancement*sc_enh
  { score: score, critical: critical, ... }
end
```

Multipliers like `sc_crit = 32` or `sc_imp = 16` should be defined as constants in the same file, similarly like [other existing constansts](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/content_facet_host_extensions.rb#L7) in that file.

Once you have that method, you'll need to allow access to it in templates, by adding its name to the [list at the bottom](https://github.com/Katello/katello/blob/master/app/models/katello/concerns/content_facet_host_extensions.rb#L119).

After that, you can easily reduce this template to

```
load_hosts(search: input('Hosts filter'), includes: [:applicable_errata]).each_record do |host|
  score = host.errata_currency
  report_row "Score": score[:score], "Host": host.name, ...
end
```

Don't worry about sorting for now, we'll address that later, report templates now supports sorting natively. If you open the PR against Katello, feel free to ping me there, I can do the review. I don't have merge permissions there but I can ACK. Then we can continue here or by that time we may need to reopen this PR against Foreman core. This repository is about to be merged there.

Please let me know if you need any more hints.

tbrisker:

PR #724, opened 2020-04-27
Refresh subscription-manager before package installation

sbernhard:
Refresh the state of the subscription-manager on debian based
distributions before updating the repositories and installting a
package. This will only be done if the subscription manager exists.
Refreshing the subscription manager is necessary to get the latest state
of activated (or disabled) repositories.


lzap (approved):
That'd work.

PR #729, opened 2020-05-06
Add /opt/puppetlabs/bin to PATH on debian

sbernhard:
The PATH to /opt/puppetlabs/bin is not set on debian based systems as it
would be set by puppt-agent.sh in /etc/profile.d/.  On Debian based
systems, /etc/profile.d/puppet-agent.sh is not sourced on no-login
shell. No-login shell is used if you run a remote execution task.
As puppet from puppetlabs is installed in /opt/puppetlabs the REX
job fails

### DIFF
--- a/job_templates/puppet_agent_disable_-_ssh_default.erb
+++ b/job_templates/puppet_agent_disable_-_ssh_default.erb
@@ -13,4 +13,7 @@ template_inputs:
 provider_type: SSH
 kind: job_template
 -%>
+<% if @host.operatingsystem.family == 'Debian' -%>
+export PATH=/opt/puppetlabs/bin:$PATH
+<% end -%>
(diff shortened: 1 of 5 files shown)


job_templates/puppet_agent_disable_-_ssh_default.erb
```diff
@@ -13,4 +13,7 @@ template_inputs:
 provider_type: SSH
 kind: job_template
 -%>
+<% if @host.operatingsystem.family == 'Debian' -%>
```
ekohl, line 16: Why is this only a problem on Debian?

sbernhard: What we have found out is, that on CentOS/RH the $PATH is set even for no-login shells.

ekohl (approved):
:+1: from a shell perspective. Not sure about the ansible part. Perhaps @evgeni has thoughts about that.

evgeni (approved):
almost feels like the shell part should be a snippet? ;-)

Ansible seems fine to me

sbernhard:
ping :)

evgeni:
Well, given I have merge permissions here...

PR #732, opened 2020-05-16
correcting post logpath

tomabg:
here not the chroot path must be used
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/installation_guide/s1-kickstart2-postinstallconfig


lzap:
I think I have a PR somewhere with this change but let's merge this and I can reabase. Thanks!

PR #734, opened 2020-05-21
Refs #26925 - Add registration template

ezr-ondrej:
After theforeman/foreman#6813 is merged, we should add register templates folder and got them in our ecosystem.


provisioning_templates/registration/linux_registration_default.erb
```diff
@@ -0,0 +1,67 @@
...
+
+rpm -Uvh https://yum.puppetlabs.com/puppet6/puppet6-release-el-7.noarch.rpm
+
+# Install puppet
+
+if [ -f /usr/bin/dnf ]; then
+  dnf -y install puppet
+else
+  yum -t -y install puppet-agent
+fi
```
ekohl: This should use https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/puppetlabs_repo.erb

provisioning_templates/registration/linux_registration_default.erb
```diff
@@ -0,0 +1,67 @@
...
+[agent]
+pluginsync      = true
+report          = true
+ignoreschedules = true
+<%- if @host.puppet_ca_server.present? -%>
+ca_server       = <%= @host.puppet_ca_server %>
+<%- end -%>
+certname        = <%= @host.certname %>
+environment     = <%= @host.environment %>
+server          = <%= @host.puppetmaster %>
```
ekohl:
```suggestion
<%= snippet 'puppet.conf' %>
```

ekohl (commented):
There's a lot of overlap with https://github.com/theforeman/community-templates/blob/develop/provisioning_templates/snippet/puppet_setup.erb as well.

ezr-ondrej:
Oh yeah, the template is much simplier now, loving it :heart: xD
Thanks @ekohl :+1:

ezr-ondrej:
I believe this is ready. @ekohl or @timogoebel ?

provisioning_templates/registration/linux_registration_default.erb
```diff
@@ -0,0 +1,27 @@
+<%#
+kind: registration
+name: Linux registration default
+model: ProvisioningTemplate
+oses:
+- CentOS
+- Fedora
```
ekohl, line 7: I think it now also works on Debian (and any platform that has curl and where puppetlabs_repo / puppet_setup works).

timogoebel: Otherwise, looks good!

ezr-ondrej: I don't really understand the matching policy, so I can't say what OSes are available. I've rolled with what we have elsewhere :man_shrugging:
I guess we are missing some better oses targeting, but it's just for default selections, so I guess we're good with the most common ones.

ekohl (approved):
Thanks! This is indeed good enough.

ekohl:
@ezr-ondrej / @tbrisker should this be picked into 2.1?

tbrisker:
no, host registration feature was only added after 2.1 branched

PR #736, opened 2020-05-26
Fixes #29921 - Corrected elsif vlanid order

lzap:
This was not working correctly at all.


lzap:
@tbrisker any chance of squeezing this into 2.1? IHAC who spotted this problem and my testing shows that this is probably the last thing missing from VLAN provisioning workflow. Let me know what we'd need (ticket, backport PR here and in core).

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -149,10 +149,10 @@ skipx
 
   # VLAN (only on physical is recognized)
   if iface.virtual? && iface.tag.present? && iface.attached_to.present?
-    if rhel_compatible && os_major >= 6
-      network_options.push("--vlanid=#{iface.tag}")
-    elsif rhel_compatible && os_major >= 7
+    if rhel_compatible && os_major >= 7
       network_options.push("--interfacename=vlan#{iface.tag}")
+    elsif rhel_compatible && os_major >= 6
```
tbrisker: shouldn't this be 
```suggestion
    elsif rhel_compatible && os_major == 6
```
since >= 7 is already handled earlier?
Or even simpler, change the first condition to
```
if rhel_compatible && os_major == 6
```
and then you don't need to modify anything else

tbrisker (commented):
Please open an issue with target version 2.1, and make sure that this is merged and cherry-picked to 2.1-stable before rc2 is built.

provisioning_templates/provision/kickstart_default.erb
lzap: Yeah it should be `<=`. Honestly I have only checked RHEL6 docs, not sure if RHEL5 supports VLAN. Probably?

tbrisker: Let's keep it as == then to preserve the original behaviour, just in case someone still provisions EL5 for some odd reason.

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -149,10 +149,10 @@ skipx
 
   # VLAN (only on physical is recognized)
   if iface.virtual? && iface.tag.present? && iface.attached_to.present?
-    if rhel_compatible && os_major >= 6
-      network_options.push("--vlanid=#{iface.tag}")
-    elsif rhel_compatible && os_major >= 7
+    if rhel_compatible && os_major >= 7
       network_options.push("--interfacename=vlan#{iface.tag}")
+    elsif rhel_compatible && os_major <= 6
+      network_options.push("--vlanid=#{iface.tag}")
```
tbrisker, line 153: Previously we added this option in all cases, not just rhel_compatible. should we still do that?

provisioning_templates/provision/kickstart_default.erb
```diff
@@ -149,10 +149,10 @@ skipx
 
   # VLAN (only on physical is recognized)
   if iface.virtual? && iface.tag.present? && iface.attached_to.present?
-    if rhel_compatible && os_major >= 6
-      network_options.push("--vlanid=#{iface.tag}")
-    elsif rhel_compatible && os_major >= 7
+    if rhel_compatible && os_major >= 7
```
tbrisker: previously this was also added for fedora >=17, should it still be done?

provisioning_templates/provision/kickstart_default.erb
lzap: Okay.

provisioning_templates/provision/kickstart_default.erb
lzap: True, let me rethink this.

provisioning_templates/provision/kickstart_default.erb
lzap: True, let me rethink this.

lzap:
Ok I am starting over :-)

The most readable seems to be:

```ruby
    if rhel_compatible && os_major == 6
      network_options.push("--vlanid=#{iface.tag}")
    else
      network_options.push("--interfacename=vlan#{iface.tag}")
    end
```

* We keep only RHEL6 as the old supported version.
* All other systems including Fedora must use the new syntax.
* Some very old Fedora versions (e.g. 16 and older) would not work but this is unsupported for years.

tbrisker (approved):
Thanks @lzap !

tbrisker:
2.1 - 986c8bb

PR #737, opened 2020-05-28
Update oVirt templates

lzap:
Update some changes which we made in default kicsktart template in oVirt and Atomic templates. Namely instead `@host.ip` or mac which point to primary_interface, it now uses the correct provision_interface.

Also include IPA and ReX snippets for oVirt.

### DIFF
--- a/provisioning_templates/PXELinux/kickstart_ovirt_pxelinux.erb
+++ b/provisioning_templates/PXELinux/kickstart_ovirt_pxelinux.erb
@@ -10,5 +10,5 @@ DEFAULT rhvh
 
 LABEL rhvh
 KERNEL <%= @kernel %>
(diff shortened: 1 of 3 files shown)


lzap:
One more comment, a customer found that ssh keys aren't deployed when installing from liveimg, because it seems that all changes in `/mnt/sysroot` are overwritten after restart. One workaround would be something like:

```
ssh_path=`df -h |grep /dev/mapper/rhvh-rhvh |awk '{print $NF}'`
mkdir -p $ssh_path/root/.ssh
cat << EOF >> $ssh_path/root/.ssh/authorized_keys
ssh-rsa AAAAB3NzaxxxxxxxxxxpQ7 foreman-proxy@xxxxxx.corp
EOF
```

But I will talk to Anacondy engineers if there is a better way so we could use our standard snippet instead which writes to `~user/.ssh`.

lzap:
Do not merge yet until I sort this out please.

tbrisker:

PR #741, opened 2020-06-04
Fixes #30013 - Allow passing http proxy to cloud connector

adamruzicka:
companion PR to https://github.com/project-receptor/satellite-receptor-installer/pull/18

### DIFF
--- a/job_templates/configure_cloud_connector_-_ansible_default.erb
+++ b/job_templates/configure_cloud_connector_-_ansible_default.erb
@@ -14,6 +14,13 @@ template_inputs:
   advanced: false
   value_type: plain
   hidden_value: true
+- name: http_proxy
+  required: false
+  input_type: user
+  advanced: true
+  value_type: plain
+  hidden_value: false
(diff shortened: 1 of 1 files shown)


job_templates/configure_cloud_connector_-_ansible_default.erb
```diff
@@ -14,6 +14,12 @@ template_inputs:
   advanced: false
   value_type: plain
   hidden_value: true
+- name: http_proxy
+  required: false
+  input_type: user
+  advanced: true
+  value_type: plain
+  hidden_value: false
```
ares, line 22: could you please add a description, something like "You can specify a HTTP proxy address that should be used for Cloud Connector connection to the cloud.redhat.com. Note that it must be HTTP proxy, not HTTPS. The tunelling of SSL (secured web socket connection) in SSL (HTTPS proxy) is currently unsupported."

adamruzicka: Added verbatim

ares (approved):
Thanks @adamruzicka, merging!

PR #742, opened 2020-06-05
Add networking to cloud init

lzap:
This is just an untested prototype - someone needs to test this. I wrote this for a customer, I don't know if networkmanager needs to be restarted or network reloaded after these changes.

We should do cloutinit networking properly some day but I have no time for investigating how this works - there are multiple formats and networking in cloud-init is simply a mess at the moment. Something should work on particular providers, something is partially generic. This workaround at least should work on redhat and debian families, there is a flag to disable this when this is irrelevant.

### DIFF
--- a/provisioning_templates/user_data/userdata_default.erb
+++ b/provisioning_templates/user_data/userdata_default.erb
@@ -76,6 +76,14 @@ runcmd:
 - |
 <%= indent(2) { snippet('saltstack_setup') } %>
 <% end -%>
+- |
+<%=
+  if rhel_compatible && !host_param_false?('cloudinit-networking')
+    indent(2) { snippet 'kickstart_networking_setup' }
+  else
+    indent(2) { snippet 'preseed_networking_setup' }
+  end
+%>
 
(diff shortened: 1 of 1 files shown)


provisioning_templates/user_data/userdata_default.erb
```diff
@@ -76,6 +76,14 @@ runcmd:
 - |
 <%= indent(2) { snippet('saltstack_setup') } %>
 <% end -%>
+- |
+<%=
+  if rhel_compatible && !host_param_false?('cloudinit-networking')
+    indent(2) { snippet 'kickstart_networking_setup' }
+  else
```
sbernhard, line 83: what about sles and other OS?

lzap: On my TODO, I am still trying to get this working on RHEL as you can see it's marked as WIP.

lzap:
Ok this does not work reliably, MAC address is not present when Foreman is passing in user data. We need to create RFE to support oVirt built in "networking" API flags which go into OpenStack metadata for cloud-init.

PR #747, opened 2020-06-22
Fix installation of salt minion

sbernhard:
1. The salt-minion package is installed by the saltstack_setup snippet
2. salt (and puppet, too) should be executed after redhat_register (see kickstart_default provisioning) - autoyast_sles_default
3. get rid of salt_package which is no longer used in preseed

### DIFF
--- a/provisioning_templates/snippet/saltstack_setup.erb
+++ b/provisioning_templates/snippet/saltstack_setup.erb
@@ -21,6 +21,8 @@ if [ -f /usr/bin/dnf ]; then
 else
   yum -t -y install salt-minion
 fi
+<% elsif @host.operatingsystem.family == 'Suse' -%>
+  /usr/bin/zypper -n install salt-minion
 <% end -%>
 
 cat > <%= etc_path %>/minion << EOF
--- a/provisioning_templates/provision/autoyast_default.erb
+++ b/provisioning_templates/provision/autoyast_default.erb
@@ -139,9 +139,6 @@ oses:
 <% if puppet_enabled -%>
       <package>rubygem-puppet</package>
 <% end -%>
-<% if salt_enabled -%>
-      <package>salt-minion</package>
-<% end -%>
     </packages>
   </software>
   <users config:type="list">
--- a/provisioning_templates/provision/autoyast_sles_default.erb
+++ b/provisioning_templates/provision/autoyast_sles_default.erb
@@ -153,9 +153,6 @@ oses:
       <package>rubygem-puppet</package>
 <% end -%>
 <% end -%>
-<% if salt_enabled -%>
-      <package>salt-minion</package>
-<% end -%>
 <% if spacewalk_enabled -%>
       <package>rhn-setup</package>
 <% end -%>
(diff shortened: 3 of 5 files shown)


sbernhard:
@bochi, maybe something you can review!?

bochi (commented):
removing the package installation from the autoinstallation and handle it within the saltstack-setup looks good to me. 
I'd change something though, which I always diskliked but never got around to file a PR yet: the snippet writes the configuration directly to /etc/salt/minion but IMHO it should use /etc/salt/minion.d/master.conf (or whatever filename, but something in minion.d)

sbernhard:
> I'd change something though, which I always diskliked but never got around to file a PR yet: the snippet writes the configuration directly to /etc/salt/minion but IMHO it should use /etc/salt/minion.d/master.conf (or whatever filename, but something in minion.d)

I guess, you mean https://github.com/theforeman/community-templates/blob/5048c9fa45679d189896a0fa47f14418e78e7ca7/provisioning_templates/snippet/saltstack_setup.erb#L27 ?

bochi:
https://github.com/theforeman/community-templates/blob/5048c9fa45679d189896a0fa47f14418e78e7ca7/provisioning_templates/snippet/saltstack_setup.erb#L26

IMHO should read: 

cat > <%= etc_path %>/minion.d/master.conf << EOF

so that the main config file is not touched/changed, that's why we have a dropin directory for salt minion config :)

bochi:
Good Morning, I filed another PR for what I wrote above since I had to fix something in the same file anyway :)
https://github.com/theforeman/community-templates/pull/749

provisioning_templates/snippet/saltstack_setup.erb
```diff
@@ -21,6 +21,8 @@ if [ -f /usr/bin/dnf ]; then
 else
   yum -t -y install salt-minion
 fi
+<% elsif @host.operatingsystem.family == 'Suse' -%>
+  /usr/bin/zypper -n install salt-minion
```
adamruzicka, line 25: I don't have much experience with zypper so I'll ask, doesn't this need something like `-y` or `--no-confirm`?

adamruzicka (commented):
One question, otherwise looks good

provisioning_templates/snippet/saltstack_setup.erb
sbernhard: --non-interactive, -n	Do not ask anything, use default answers
				automatically.

adamruzicka:
```
           -n, --name
               Select packages by their name, don’t try to select by capabilities.
```

From https://en.opensuse.org/SDB:Zypper_manual_%28plain%29

bochi: Global Options vs. Command Options
Syntax is ``zypper <global-options> <command> <command-options>``

The global option -n is --non-interactive, so @sbernhard is right
BTW: -y would work too, its an alias ;)

sbernhard: just because I copied it from the puppet snippet :)

adamruzicka (approved):
ACK

adamruzicka:
Thank you @sbernhard & @bochi !

PR #748, opened 2020-06-25
Update kickstart_default_user_data.erb

calvingsmith:
Added support for ansible tower provisioning callback. This is already in the finish template, adding here. Storing provisioning callback script  in /tmp so that the tmp directory cleaner will delete it in time, but it's there in the near term for inspection

### DIFF
--- a/provisioning_templates/user_data/kickstart_default_user_data.erb
+++ b/provisioning_templates/user_data/kickstart_default_user_data.erb
@@ -68,10 +68,8 @@ fi
 
 <% if host_param_true?('ansible_tower_provisioning') -%>
     <%= save_to_file('/tmp/ansible_provisioning_call.sh', snippet('ansible_tower_callback_script')) %>
-    bash /tmp/ansible_provisioning_call.sh
+    /bin/sh /tmp/ansible_provisioning_call.sh
 <% end -%>
-
-
 
 # UserData still needs to mark the build as finished
 <%= snippet 'built' %>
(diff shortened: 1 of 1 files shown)


provisioning_templates/user_data/kickstart_default_user_data.erb
```diff
@@ -66,5 +66,12 @@ fi
 <%= snippet 'saltstack_setup' %>
 <% end -%>
 
+<% if host_param_true?('ansible_tower_provisioning') -%>
+    <%= save_to_file('/tmp/ansible_provisioning_call.sh', snippet('ansible_tower_callback_script')) %>
+    bash /tmp/ansible_provisioning_call.sh
+<% end -%>
```
lzap, line 72: Can you standardize both here and in the kickstart finish template on calling `/bin/sh`? Bash might not be present on some non-RH images.

In the finish template the script is called directly however I don't understand how this can work since it's being created using `cat` without executable permissions. Weird, that's probably a bug.

provisioning_templates/user_data/kickstart_default_user_data.erb
```diff
@@ -66,5 +66,12 @@ fi
...
+
+
+
```
lzap: Just one empty line should be enough, we have too many of them actually.

lzap (approved):
Thanks, two requests.

lzap (approved):
Thanks.

PR #749, opened 2020-06-26
Fix minion config

bochi:
This PR includes 2 commits:
 
* use /etc/salt/minion.d/minion.conf for configuring the salt-minion, that's just because it is the recommended way rather than directly modify /etc/salt/minion

* write /etc/salt/minion_id during deployment
This is important because in some cases, ie. when using DHCP, the salt-minion may compute a minion_id that is different to what the system name in foreman is. This will make accepting the salt key fail and the system will never be accepted/working with foreman. So when using the foreman salt plugin, the minion_id should always be equal to the system's name in foreman.


provisioning_templates/snippet/saltstack_setup.erb
```diff
@@ -23,10 +23,12 @@ else
 fi
 <% end -%>
 
-cat > <%= etc_path %>/minion << EOF
+cat > <%= etc_path %>/minion.d/minion.conf << EOF
```
sbernhard, line 26: Is this path automatically used by salt or do we need to adapt something else?

provisioning_templates/snippet/saltstack_setup.erb
```diff
@@ -23,10 +23,12 @@ else
...
 <%= snippet 'saltstack_minion' %>
 EOF
 
+echo <%= @host.name %> > <%= etc_path %>/minion_id
```
sbernhard, line 30: Is this the preferred path or should it be in etc_path/minion.d/minion_id, too?

provisioning_templates/snippet/saltstack_setup.erb
sbernhard: https://docs.saltstack.com/en/latest/ref/configuration/examples.html#example-minion-configuration-file ->

    # Per default the minion will automatically include all config files 
    # from minion.d/*.conf (minion.d is a directory in the same directory
    # as the main minion config file).

provisioning_templates/snippet/saltstack_setup.erb
bochi: The minion_id will always reside in /etc/salt directly

provisioning_templates/snippet/saltstack_setup.erb
bochi: Correct, all files within the .d directory are automatically included. The advantage is that in case the OS package for salt-minion provides a new config file it can update it without any conflicts.

sbernhard (approved):

sbernhard:
LGTM. Ready to be merged!

adamruzicka (approved):
Makes sense, ACK

adamruzicka:
Thank you @bochi & @sbernhard !

